PR #12, opened 2016-07-20
added a couple more missing files to the makefile

ejohnstown:
These were missing in v0.2.0. They are included in the archives for the website.


ejohnstown:
Chris, could you verify this since you discovered it?

cconlon:
Verified, thanks!

PR #13, opened 2016-07-22
More Cleanup

ejohnstown:
More proper return code and length checking.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -411,16 +411,16 @@ int GrowBuffer(Buffer* buf, uint32_t sz, uint32_t usedSz)
 
 void ShrinkBuffer(Buffer* buf, int forcedFree)
 {
-    WLOG(WS_LOG_DEBUG, "Entering %s", __func__);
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -658,67 +659,130 @@ static int GetInputData(WOLFSSH* ssh, uint32_t size)
...
+                     uint8_t* buf, uint32_t len, uint32_t *idx)
+{
+    int result;
+
+    result = GetUint32(sSz, buf, len, idx);
+
+    if (result == WS_SUCCESS) {
+        result = WS_BUFFER_E;
+        if (*idx < len && *idx + *sSz <= len) {
+            XMEMCPY(s, buf + *idx, *sSz);
```
JacobBarthelmeh: In other places in code WMEMCPY was used. Is this a typo?

ejohnstown:
Typo. And I found another one. I just checked a bunch of the STDLIB aliases we use in wolfSSL and didn't find them. Thanks for catching this.

JacobBarthelmeh:
Thanks, looks good.

PR #14, opened 2016-08-10
Update misc inline code

ejohnstown:
Change misc.c error to warning and exclude the misc.c code from being compiled. Most people include all .c files and by default inlining is allowed, which in turn causes an #error in misc.c and it must be excluded. Since we know its already been properly included there is no reason to throw error here. Instead, show warning and exclude code in .c file.

Copied from wolfSSL pull request #521.


ejohnstown:
David, could you take a look at this? I copied your similar patch to wolfCrypt's misc.c.

PR #15, opened 2016-08-25
err_sys() static analysis issue

ejohnstown:
Copied change to err_sys() from wolfSSL. Cleaned up a static analysis issue. Also added the attribute noreturn to the function since it doesn't actually return.


dgarske:
Looks great John. Thanks!

PR #17, opened 2016-10-04
Transfer Highwater Mark, Session Key Refactoring

ejohnstown:
1. Add in tracking of the number of bytes encrypted and decrypted. Trigger a callback function if the highwater mark is reached in either direction.
2. Refactor the session key storage and use to make re-keying easier to implement.
3. Moved the context and session init and resource free functions from ssh.c to internal.c.


JacobBarthelmeh:
Thanks for the test case, looks good.

PR #18, opened 2016-10-05
wolfCrypt

ejohnstown:
1. Add call to wolfCrypt_Init() to wolfSSH_Init().
2. Modify have_wolfssl.m4 to check for wolfCrypt, not wolfSSL.
3. Fix bug in echoserver using Sha256 with wolfSSL using AES-NI.


JacobBarthelmeh:
Looks good!

PR #20, opened 2016-10-13
RSA Private Key Generation Wrapper Function

ejohnstown:
1. Implemented the RSA key generation wrapper function.
2. Rearrange the unit test code.
3. Added RSA key generation wrapper function to unit test.
4. Removed certificate load from echoserver.
5. Not using certificates, removed the functions and storage.
6. Removed unused certificate files.
7. Renamed certs directory as keys.
8. Removed the example server from the build for now.
9. Created new server key with the RSA key generation function.
10. Add IDs for public keys for keygen use.
11. Whitespace cleanup.


JacobBarthelmeh:
Thanks for discussion. The code looks good!

PR #21, opened 2016-10-16
Data windows

ejohnstown:
Updates to the channel receive window handling. Needed to send updates to the client when enough free space in the receive window is made available.


ejohnstown:
I'm going to consolidate my three pending branches into one pull request.

PR #22, opened 2016-10-23
Rekeying

ejohnstown:
1. Added rekeying during runtime, with a trigger function and highwater marks for number of bytes received and transmit.
2. Added a key generation function for producing a new RSA key for server identity.
3. Suggest updates from beta test.


ejohnstown:
This is a release candidate for v1.0.0.

src/internal.c
```diff
@@ -2303,11 +2501,47 @@ static int DoPacket(WOLFSSH* ssh)
...
+                        ret = WS_MEMORY_E;
+                    }
+                }
+
+                if (ssh->keyingState == KEYING_KEYED ||
+                    ssh->keyingState == KEYING_KEXINIT_SENT) {
+
+                    uint32_t idSz;
+
+                    wc_ShaUpdate(&ssh->handshake->hash,
```
toddouska: Return code check?

src/internal.c
```diff
@@ -2303,11 +2501,47 @@ static int DoPacket(WOLFSSH* ssh)
...
+                                 ssh->clientId, ssh->clientIdSz);
+
+                    idSz = (uint32_t)WSTRLEN(sshIdStr) - SSH_PROTO_EOL_SZ;
```
toddouska: What is strlen() result is less than SSH_PROTO_EOL_SZ?

ejohnstown: `sshIdStr` is a contant string.

src/internal.c
ejohnstown: I'll run through and fix the rest of those.

PR #37, opened 2017-09-12
IDE Build

ejohnstown:
Added an IDE build directory with a sample Visual Studio solution inside. Builds for Windows 64- and 32-bit Debug and Release builds. The solution builds the library, echoserver, API test, and unit test.


JacobBarthelmeh:
- Suggestion of Readme for VS build pointing to wolfcrypt directory Readme (wolfssh/ide/winvs/Readme)

- wolfssl/options.h being included on windows?

- is user_settings.h assumed?

- Echoserver no wolfssh/test.h (test.h needs added to include.am)

- _ProcessBuffer already defined in wolfssh.lib (project api-test)

- After copying test.h over there was some errors on build. It appeared to be related to threading but I may be missing a header needed? (was using "make dist-zip" then copying zip to Windows VM)

JacobBarthelmeh:
Tested building with VS 2012 on Window 7, thanks John.

ejohnstown:
options.h does properly exist is VS, it is just empty boilerplate. Part of the release process is to copy options.h.in over-top options.h.

ejohnstown:
This should build (mostly) cleanly. (The Debug Win32 warns that edit-and-continue will be disabled due to some other option. Edit-and-continue is set on the library build for the Win32 wolfssl. I don't know why it is set, as none of the others use it and it just allows one to swap out the library at debug runtime.)

Getting the DLLs to work got tricky. I had to add the file wolfssh.props to define some user macro properties in the project so that the wolfSSL libraries and includes can be stored anywhere. For each of the test tools, I added a vcxproj.user file setting the executable path for the DLL debug runs of the test tools from within Visual Studio. (You have to deal with your own PATH to the wolfssl.dll on the command line.)

I left in options.h as we tell people at support to make sure to include it before the other headers. The gcc/clang builds won't work without it. A step of the wolfSSL release process is to make a clean copy of it without any options, so under Windows it becomes a don't-care.

ejohnstown:
David: I made the change to the echoserver, wrt options.h, that you had in the wolfMQTT examples.

PR #41, opened 2017-09-28
Logging Update

ejohnstown:
1. Added a logging function that always works whether or not debugging is enabled.
2. Modified the WLOG macro to always be built in, but disables cleanly.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -1,4 +1,4 @@
-/* log.c 
+/* log.c
  *
  * Copyright (C) 2014-2016 wolfSSL Inc.
  *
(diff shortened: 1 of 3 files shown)


src/log.c
```diff
@@ -34,27 +34,31 @@
 #include <wolfssh/log.h>
 #include <wolfssh/error.h>
 
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdarg.h>
+#include <time.h>
```
JacobBarthelmeh: This now always tries to include time.h and other headders. Should it be conditional for portability/embedded development?

wolfssh/log.h
```diff
@@ -47,20 +47,24 @@ enum wolfSSH_LogLevel {
...
-    #define WLOG(a, b, ...)
-#endif /* DEBUG_WOLFSSH */
+    #define FMTCHECK
+#endif /* __GNUC__ */
+
+
+WOLFSSH_API void wolfSSH_Log(enum wolfSSH_LogLevel,
+                             const char *const, ...) FMTCHECK;
+
+#define WLOG(...) do { \
```
JacobBarthelmeh: Previously WLOG could be compiled out. Why the transition to always calling the function wolfSSH_LogEnabled in the case where DEBUG_WOLFSSH is not defined?

src/log.c
ejohnstown: The time stuff is optional, so I wrapped it in a guard check.

wolfssh/log.h
ejohnstown: It'll get optimized out in this case. This ensures any changes won't break the logging. The code is still checked by the compiler, despite the inline function indicating that logging is disabled always returning that. SSH likes to have some things logged whether or not our debug logging is enabled.

PR #46, opened 2018-03-27
update SHA256 name

JacobBarthelmeh:


JacobBarthelmeh:
For linking against wolfSSL when the wolfSSL library is compiled with --enable-opensslextra

PR #50, opened 2018-04-04
check AES block size and return verify mac error

JacobBarthelmeh:


src/internal.c
```diff
@@ -4070,6 +4071,7 @@ static INLINE int Decrypt(WOLFSSH* ssh, byte* plain, const byte* input,
             break;
 
         case ID_AES128_CBC:
+            sz = sz - (sz % AES_BLOCK_SIZE); /* decrypt AES block size */
```
JacobBarthelmeh: Am reviewing this code section. It could be redundant or may be better to be an error case if it gets to this point and the buffer is not a multiple of AES block size.

ejohnstown:
Can you fix this so it is on its own branch in your repo rather than on the master branch?

JacobBarthelmeh:
Ok, will close this PR out and open a new one shortly.

PR #52, opened 2018-04-06
add example code for MPLABX use

JacobBarthelmeh:


cconlon:
I think we need to expand the README with more exact instructions for the user.  It would be nice to walk them through converting a specific Harmony example app.  Although I wasn't able to get an example building yet, a few comments.  I was trying to convert the "berkely_tcp_server.X" example.

- README is unclear if APP_SSH_Initialize() and APP_SSH_Tasks() should replace existing APP_Initialize() and APP_Tasks() or be added alongside.
- At first, I was using the crypto layer inside Harmony and needed to enable ECC then re-generate the harmony code.  Note that Harmony Configurator rewrites system_init.c and system_tasks.c.
- On first try, build couldn't find <wolfssl/wolfcrypt/settings.h>.  ssh.h includes this when WOLFSSL_USER_SETTINGS has been defined.  Is this correct or should we be including "user_settings.h" instead?
- README doesn't mention the need to include and build wolfSSL/wolfCrypt source files.
- I needed to add an "extern" definition in wc_port.h for pic32_time().  Not sure if this is needed in final build.  I may have been missing a step.
- README also doesn't mention needing to include/build the wolfSSH source files.

ejohnstown:
@JacobBarthelmeh, you have this as a branch in your fork of the repo. It has been stale for 5ish years now. I'm going to close it. The history exists. It can be reopened when updated as is ready, or you can make a new PR.

PR #55, opened 2018-04-21
Add missing error string "invalid algorithm id"

slcasner:


ejohnstown:
Superceded by PR #59. I added wolfSSL's error string unit test and found some other missing strings. Thank you for submitting this.

PR #56, opened 2018-05-03
SCP Server Support

cconlon:


cconlon:
Added notes to README for scp, thanks.

PR #58, opened 2018-05-03
fixes for scan-build warnings

cconlon:
Fixes for scan-build warnings that showed up with:

```
$ clang --version
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```


ejohnstown:
Looks good. Thanks!

PR #59, opened 2018-05-10
Error Code Unit Test

ejohnstown:
1. Add a unit test to check all error codes have strings.
2. Add public function to return the error string for an error code.
3. Fill in the missing error strings.


JacobBarthelmeh:
can you add a fix for WS_VERIFY_MAC_E and WS_CREATE_MAC_E strings to this please. When printing out the test case I noticed that both have the same string of "verify mac error".

ejohnstown:
Fixed, rebased, and force pushed.

PR #67, opened 2018-05-25
Fix Fuzz

ejohnstown:
Fixed a few fuzz tester found issues.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -2200,8 +2200,12 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     if (ret == WS_SUCCESS) {
         f = buf + begin;
         ret = GetUint32(&fSz, buf, len, &begin);
-        if (ret == WS_SUCCESS && (fSz > len - begin - LENGTH_SZ)) {
-            ret = WS_BUFFER_E;
+    }
+
+    if (ret == WS_SUCCESS) {
+        if (fSz > len - begin) {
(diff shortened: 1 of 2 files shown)


src/internal.c
```diff
@@ -2200,7 +2200,7 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     if (ret == WS_SUCCESS) {
         f = buf + begin;
         ret = GetUint32(&fSz, buf, len, &begin);
-        if (ret == WS_SUCCESS && (begin + fSz + LENGTH_SZ > len)) {
+        if (ret == WS_SUCCESS && (fSz > len - begin - LENGTH_SZ)) {
```
JacobBarthelmeh: Could this lead to a possible underflow if begin + LENGTH_SZ is greater than len?

JacobBarthelmeh: An example case is if len is 34 and begin is 29 initially. A call to GetUint32 would update begin to be 33. In that case taking len (34) - begin (33) - LENGTH_SZ (4) would result in -3. 
fsz, len, and begin are unsigned values so it would make it a very larger unsigned value.

ejohnstown: And I think one of the other spots I fixed this overflow there was a check for the underflow condition. Will add.

ejohnstown: GetUint32() should return an error in that case.

ejohnstown: Duh. I'm having a can't do math issue.

PR #73, opened 2018-06-07
SFTP/SCP on Nucleus

JacobBarthelmeh:


src/wolfscp.c
```diff
@@ -2025,7 +2161,7 @@ int wsScpSendCallback(WOLFSSH* ssh, int state, const char* peerRequest,
             }
 
             ret = (word32)WFREAD(buf, 1, bufSz, sendCtx->fp);
-            if ((ret < 0) || (fileOffset + ret == *totalFileSz)) {
+            if ((ret <= 0) || (fileOffset + ret == *totalFileSz)) {
```
JacobBarthelmeh: changes behavior for test after read. Now also when file read returns 0 it is considered an error.

wolfssh/wolfscp.h
```diff
@@ -24,6 +24,7 @@
 #define WOLFSSH_WOLFSCP_H
 
 #include <wolfssh/settings.h>
+#include <wolfssh/ssh.h>
```
JacobBarthelmeh: Added for WOLFSSH structure arguments in function declartions

src/wolfscp.c
```diff
@@ -1104,6 +1106,7 @@ int ParseScpCommand(WOLFSSH* ssh)
                             /* skip space */
                             idx += 2;
                             ssh->scpBasePath = cmd + idx;
+                            clean_path((char*)ssh->scpBasePath);
```
JacobBarthelmeh: clean_path will make adjustments to the buffer (always smaller than starting string if changed). Since this is a pointer to cmd + idx it will alter the buffer cmd. I did not see any issue with it but is worth mentioning.

PR #81, opened 2018-07-02
additional sftp and scp tests

JacobBarthelmeh:


JacobBarthelmeh:
Note: This adds external connection test to test.rebex.net

PR #83, opened 2018-07-27
rename from different drives and list drives

JacobBarthelmeh:


JacobBarthelmeh:
Two changes for Nucleus port

- Change behavior to print out all drives when opening the directory "/" instead of reporting an error

- Enhancement to handle renaming a file from drive A to drive B. NU_Rename is similar to Windows rename in that it reports an error for this case. The workaround is to use fopen, fread, fwrite in this special case.

PR #98, opened 2018-09-05
TCP Port Forwarding

ejohnstown:
Implementation of client side initiated port forwarding. Use the example wolffwd tool to set up a tunnel to a server.


src/internal.c
```diff
@@ -3544,6 +3605,19 @@ static int DoGlobalRequest(WOLFSSH* ssh,
         ret = GetBoolean(&wantReply, buf, len, &begin);
     }
 
+#ifdef WOLFSSH_FWD
+    if (ret == WS_SUCCESS) {
+        if (WSTRNCMP(name, "tcpip-forward", nameSz) == 0) {
+            fprintf(stderr, "tcpip-forward requested\n");
```
JacobBarthelmeh: frpintf and fflush should be mapped from a macro for portability

src/ssh.c
```diff
@@ -744,6 +759,18 @@ int wolfSSH_stream_exit(WOLFSSH* ssh, int status)
 }
 
 
+int wolfSSH_SendIgnore(WOLFSSH* ssh, const byte* buf, word32 bufSz)
```
JacobBarthelmeh: is the unused buf and bufSz inputs for future use cases?

src/ssh.c
```diff
@@ -935,3 +982,279 @@ const char* wolfSSH_GetSessionCommand(const WOLFSSH* ssh)
 
     return NULL;
 }
+
+
+int wolfSSH_worker(WOLFSSH* ssh, word32* channelId)
+{
+    int ret = WS_SUCCESS;
+
+    printf("Entering wolfSSH_worker()\n");
```
JacobBarthelmeh: this printf, and the rest in this function, should be mapped with macro for portability

tests/include.am
```diff
@@ -8,20 +8,20 @@ noinst_PROGRAMS += tests/unit.test tests/api.test \
                    tests/testsuite.test
 
 tests_unit_test_SOURCES      = tests/unit.c
-tests_unit_test_CFLAGS       = -DNO_MAIN_DRIVER $(AM_CFLAGS)
-tests_unit_test_LDADD        = src/libwolfssh.la $(LIB_STATIC_ADD)
+tests_unit_test_CPPFLAGS     = -DNO_MAIN_DRIVER
+tests_unit_test_LDADD        = src/libwolfssh.la
```
JacobBarthelmeh: why remove the additional static librarys here? will this hinder cases where testing with memory tools that require linking against a third party library?

src/ssh.c
ejohnstown: In the spec, the ignore message can be about any size, any data. It is ignored by the peer. The idea is to use random sizes to disguise actual traffic patterns. I haven't seen OpenSSH use it ever. I wanted to update the function at some point to use the parameters.

src/ssh.c
ejohnstown: They need to be deleted. They were added for debugging when I didn't want the massive spray of all the other debugging messages.

tests/include.am
ejohnstown: I don't believe LIB_STATIC_ADD is common or recommended anywhere except in wolfSSL's Makefile.am. I understand the common practice is to add "LDFLAGS=-llibraryname" to the configure command to add any required libraries.

src/internal.c
ejohnstown: They shouldn't be used here at all. That should be a WLOG().

ejohnstown: I deleted that block because while I was checking for the string I wasn't doing anything with it.

ejohnstown:
The valgrind issue you saw I had fixed in commit 8a13a82.

JacobBarthelmeh:
Thanks!

PR #105, opened 2018-09-20
free example SFTP client working directory

JacobBarthelmeh:


wolfsftp/client/sftpclient.c
```diff
@@ -888,6 +888,7 @@ THREAD_RETURN WOLFSSH_THREAD sftpclient_test(void* args)
 
     doCmds();
 
+    XFREE(workingDir, NULL, DYNAMIC_TYPE_TMP_BUFFER);
```
aidangarske, line 891: **🔵 [Low] Static workingDir left dangling after XFREE in sftpclient_test()** · _Logic_

The PR adds `XFREE(workingDir, NULL, DYNAMIC_TYPE_TMP_BUFFER);` at line 891, correctly fixing a real leak of the buffer allocated at line 877. Verified independently against the source: `workingDir` is a file-scope `static char*` (line 34), and the new free does not reset it to NULL, leaving the static dangling after the function returns. Not currently exploitable — `sftpclient_test()` has exactly one caller (`main()` at line 924, guarded by `!NO_MAIN_DRIVER`, confirmed by full-tree grep: only the prototype in sftpclient.h:26, the definition at sftpclient.c:778, and that one call site), and on any hypothetical re-entry the pointer is unconditionally reassigned by the `XMALLOC` at line 877 before any read. Also not a double-free risk: the other `XFREE(workingDir, ...)` at line 524 (the `cd` handler) always reassigns via `XMALLOC` on the next line, with `err_sys()` terminating on allocation failure, so `workingDir` is always a live allocation when `doCmds()` returns. The concern is defense-in-depth only: `sftpclient_test` is declared with `THREAD_RETURN WOLFSSH_THREAD` linkage in sftpclient.h, signalling intent for it to be driven repeatedly by a `NO_MAIN_DRIVER` test harness. In that shape, a dangling static that survives between invocations is a latent hazard if a future edit adds a read of `workingDir` before the line-877 reassignment (for example an early-exit or error path between entry and the RealPath block). Severity held at Low: no reachable use-after-free path exists in the current tree, so the CWE-416 mapping describes the latent class, not a present defect.

**Fix:** Set the pointer to NULL after the free so the static does not outlive its allocation:

    XFREE(workingDir, NULL, DYNAMIC_TYPE_TMP_BUFFER);
    workingDir = NULL;

XFREE(NULL, ...) is a no-op, so this remains safe on any re-entry.

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Static workingDir left dangling after XFREE in sftpclient_test()** — `wolfsftp/client/sftpclient.c:891`

PR #114, opened 2018-11-15
Static Analysis Fixes

ejohnstown:
Ran the clang static analysis and infer and fixed most of the reported items. There were many that infer found that looked like false positives.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -136,10 +136,10 @@ static int SFTP_ServerRecvInit(WOLFSSH* ssh) {
     if (sz > 0) {
         byte* data = (byte*)WMALLOC(sz, NULL, DYNTYPE_BUFFER);
         if (data ==  NULL) return WS_MEMORY_E;
-        if ((len = wolfSSH_stream_read(ssh, data, sz)) != (int)sz) {
-            return len;
-        }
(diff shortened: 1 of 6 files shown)


src/wolfsftp.c
```diff
@@ -903,6 +903,7 @@ int wolfSSH_SFTP_RecvOpenDir(WOLFSSH* ssh, int reqId, word32 maxSz)
                 DYNTYPE_SFTP);
         if (dirList == NULL) {
             WFREE(dir, ssh->ctx->heap, DYNTYPE_BUFFER);
+			WCLOSEDIR(&ctx);
```
JacobBarthelmeh: it looks like spacing or tabs are off here?

wolfsftp/client/sftpclient.c
```diff
@@ -308,6 +308,8 @@ int doCmds()
             if (pt[0] != '/') {
                 int maxSz = (int)WSTRLEN(workingDir) + sz + 2;
                 f = XMALLOC(maxSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+				if (f == NULL)
```
JacobBarthelmeh: it's a good check, just the indent looks farther than should be

PR #129, opened 2018-12-13
check on mpint shared secret

JacobBarthelmeh:
Fix for an intermittent connection fail case when leading 0's are in the shared secret.


JacobBarthelmeh:
> Are there other spots where `CreateMpint()` can be used?

I think there is a couple additional spots where it could be used. Will go through and update them.

wolfssh/misc.h
```diff
@@ -45,6 +45,7 @@ WOLFSSH_LOCAL void ato32(const byte*, word32*);
 WOLFSSH_LOCAL void c32toa(word32, byte*);
 WOLFSSH_LOCAL void ForceZero(const void*, word32);
 WOLFSSH_LOCAL int ConstantCompare(const byte*, const byte*, word32);
+WOLFSSH_LOCAL void CreateMpint(byte*, wrod32*, byte*);
```
ejohnstown: s/wrod/word/

JacobBarthelmeh: thanks!

PR #130, opened 2018-12-14
Inline Fix

ejohnstown:
Add check if building echoserver to test.h so the static functions aren't included when building not-inline.


JacobBarthelmeh:
Thanks, tested and it resolve the --disable-inline warning

PR #134, opened 2019-02-19
Non-blocking wolfSFTP

ejohnstown:
Overhaul of wolfSFTP to allow for non-blocking sockets. Also updates to work with Windows desktop and CE with the Win32 API rather than the posix wrappers.


ejohnstown:
@dgarske
> These changes are great! Just a few minor items.

I fixed the two issues you identified with commit 2350ffe.

PR #135, opened 2019-02-21
Configure Update

ejohnstown:
1. Add a disable for the example tools to configure.
2. Add a disable for the client side code to configure.
3. Add a disable for the server side code to configure.


JacobBarthelmeh:
This could use revisited for SFTP builds (removing SFTP client and not having tests when no server is available). Along with tie into NO_WOLFSSH_SERVER with --disable-server in the future.

PR #145, opened 2019-03-26
Public Key Check Callback

ejohnstown:
The public key check callback hook is given a pointer to the public key, the size of the key, and the application-specific context data.
1. Added a callback function hook for checking the public key sent to the client by the server. It defaults to accepting the key.
2. Added accessors for the public key check callback function and context data.
3. Added a dummy callback to all the example tools.

### DIFF
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -152,6 +152,16 @@ static int wsUserAuth(byte authType,
 }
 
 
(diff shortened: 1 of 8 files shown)


JacobBarthelmeh:
should WOLFSSH structure inherit the callback in SshInit?

ejohnstown:
It is the pattern we use for the other callbacks. The function is stored in the WOLFSSH_CTX and the callback's application specific context data is stored in the WOLFSSH session.

PR #146, opened 2019-04-01
Add AES-CTR, make Algo configurable

kojo1:


ejohnstown:
Two issues:

1. It looks like you are submitting the PR based on your master branch, instead of working in a topic branch.

2. You don't have the shell variable WOLFSSH_EXTERNAL_TEST set to 1. Try running

    $ WOLFSSH_EXTERNAL_TEST=1 make check

The external test is failing. I will look into updating the Jenkins test and/or wolfSSL's wolfSSH configure option.

src/internal.c
```diff
@@ -2706,6 +2709,7 @@ static int DoNewKeys(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
                 break;
 
             case ID_AES128_CBC:
+            case ID_AES128_CTR:
```
ejohnstown: When using AesSetKey for an AES-CTR mode key, you need to use AES_ENCRYPTION.

src/internal.c
```diff
@@ -1707,7 +1708,7 @@ static int GetNameList(byte* idList, word32* idListSz,
 }
 
 
-static const byte  cannedEncAlgo[] = {ID_AES128_GCM, ID_AES128_CBC};
+static const byte  cannedEncAlgo[] = {ID_AES128_GCM, ID_AES128_CTR, ID_AES128_CBC};
```
ejohnstown: When you add items to the cannedXxxAlgo lists, you need to add the corresponding names to the cannedXxxAlgoNames lists. (In the future, this is something that needs to be configurable.)

kojo1: added. Still need to be configurable.

src/internal.c
kojo1: Fixed

ejohnstown:
Closed in favor of PR #147.

PR #147, opened 2019-04-02
Add AES-CTR, make Algo configurable

kojo1:


ejohnstown:
I found it the problem. wolfCrypt's AES-CTR mode doesn't work when in and out are the same. It encrypts the counter into out, then it XORs in into out. If in and out are the same, you overwrite in with the encrypted counter, then XOR out and in together, which are the same, you get zeroes. I'm going to submit a PR over on wolfSSL to change this to use an intermediate temp.

kojo1:
Hi John,

Thank you for looking into and taking care.

Kojo



2019年4月4日(木) 6:56 John Safranek <notifications@github.com>:

kojo1:
Thank you for taking care.
Meanwhile, made cannedXXXAlgoNames configurable.

ejohnstown:
Jenkins please retest

ejohnstown:
retest this please

ejohnstown:
1. Add `#define WOLFSSL_AESCTR` to the file ide/winvs/user_settings.h.
2. Add `#ifdef WOLFSSL_AESCTR` guards around the calls to wc_AesCtrEncrypt().
3. Also add the guard around the AES-CTR cases in SendNewKeys() and DoNewKeys().

kojo1:
Would they be ok with WOLFSSH_NO_xxx? Would the names be aligned with WOLFSSL_xxx?
VS build is not working.

ide/winvs/user_settings.h
```diff
@@ -11,7 +11,7 @@
 #define HAVE_ECC
 #define HAVE_AESGCM
 #define HAVE_HASHDRBG
-#define WOLFSSL_AESCTR
+#define WOLFSSL_AES_COUNTER
```
ejohnstown: Oh yeah, the flag is `WOLFSSL_AES_COUNTER`. I'm sorry for the wrong request.

ejohnstown:
> Would they be ok with WOLFSSH_NO_xxx? Would the names be aligned with WOLFSSL_xxx?
> VS build is not working.

They should be referring to the wolfCrypt configure options, which start with WOLFSSL. For example, for the key generation functions we check for WOLFSSL_KEY_GEN. The wolfSSL/wolfCrypt configuration does guide the wolfSSH configuration.

ide/winvs/user_settings.h
kojo1: But still not seem to work. Don't see why.

Issue #159, opened 2019-04-25
scp and sftp build errors with gcc 8.3.0


bagder:
I build wolfssh straight from git, currently on commit 57a869ecf908cb041 (dated April 22) on Debian Linux 64 bit (unstable) with gcc 8.3.0 and then I get this set of build errors:

~~~shell
$ CPPFLAGS=-I/home/daniel/build-wolfssl/include LDFLAGS=-L/home/daniel/build-wolfssl/lib ./configure --prefix=/home/daniel/build-wolfssh --enable-scp --enable-sftp --disable-examples
[cut]
$ make
make -j9  all-am
make[1]: Entering directory '/home/daniel/src/bagder-wolfssh'
  CC       src/libwolfssh_la-wolfsftp.lo
  CC       src/libwolfssh_la-wolfscp.lo
In file included from ./wolfssh/ssh.h:37,
                 from ./wolfssh/wolfscp.h:27,
                 from src/wolfscp.c:34:
src/wolfscp.c: In function ‘ParseBasePathHelper’:
./wolfssh/port.h:228:35: error: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
         #define WSTRNCPY(s1,s2,n) strncpy((s1),(s2),(n))
                                   ^~~~~~~~~~~~~~~~~~~~~~
src/wolfscp.c:1074:9: note: in expansion of macro ‘WSTRNCPY’
         WSTRNCPY(buf, ssh->scpBasePath, sz);
         ^~~~~~~~
./wolfssh/port.h:207:31: note: length computed here
     #define WSTRLEN(s1)       strlen((s1))
                               ^~~~~~~~~~~~
src/wolfscp.c:1067:24: note: in expansion of macro ‘WSTRLEN’
         int  sz = (int)WSTRLEN(ssh->scpBasePath);
                        ^~~~~~~
src/wolfscp.c: In function ‘wolfSSH_SetScpErrorMsg’:
./wolfssh/port.h:228:35: error: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
         #define WSTRNCPY(s1,s2,n) strncpy((s1),(s2),(n))
                                   ^~~~~~~~~~~~~~~~~~~~~~
src/wolfscp.c:710:9: note: in expansion of macro ‘WSTRNCPY’
         WSTRNCPY(value + 1, message, valueSz);
[...]

PR #160, opened 2019-04-25
fix for gcc 8 build warnings

JacobBarthelmeh:
For https://github.com/wolfSSL/wolfssh/issues/159


JacobBarthelmeh:
testing a bit farther with SCP change on valueSz

PR #163, opened 2019-04-26
Add example ssh server code for renesas CS+

miyazakh:
Added example ssh server under renesas CS+ and alpha project
Fixed CC-RX compiler warnings

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -692,7 +692,7 @@ int GenerateKey(byte hashId, byte keyId,
         return WS_BAD_ARGUMENT;
     }
 
-    digestSz = wc_HashGetDigestSize(hashId);
+    digestSz = wc_HashGetDigestSize((enum wc_HashType)hashId);
     if (digestSz == 0) {
         WLOG(WS_LOG_DEBUG, "GK: bad hash ID");
         return WS_BAD_ARGUMENT;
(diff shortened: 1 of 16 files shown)


src/internal.c
```diff
@@ -687,7 +687,7 @@ int GenerateKey(byte hashId, byte keyId,
         return WS_BAD_ARGUMENT;
     }
 
-    digestSz = wc_HashGetDigestSize(hashId);
+    digestSz = wc_HashGetDigestSize((enum wc_HashType)hashId);
```
cconlon: Casting hashId to (enum wc_HashType) so many times seems a little messy.  Could we make a new variable and do the cast once?  Or, alternatively change the prototype for GenerateKey() to accept a "enum wc_HashType" for hashId?  GenerateKey() is an internal/local function, so changing an argument type should be ok.

src/internal.c
```diff
@@ -699,25 +699,25 @@ int GenerateKey(byte hashId, byte keyId,
...
+        ret = wc_HashUpdate(&hash, (enum wc_HashType)hashId, &pad, 1);
     if (ret == WS_SUCCESS)
-        ret = wc_HashUpdate(&hash, hashId, k, kSz);
+        ret = wc_HashUpdate(&hash, (enum wc_HashType)hashId, k, kSz);
     if (ret == WS_SUCCESS)
-        ret = wc_HashUpdate(&hash, hashId, h, hSz);
+        ret = wc_HashUpdate(&hash, (enum wc_HashType)hashId, h, hSz);
     if (ret == WS_SUCCESS)
-        ret = wc_HashUpdate(&hash, hashId, &keyId, sizeof(keyId));
+        ret = wc_HashUpdate(&hash, (enum wc_HashType)hashId, &keyId, sizeof(keyId));
```
cconlon: Some of these lines extend past our coding standard 80-character line limit.  Please adjust, thanks.

src/internal.c
```diff
@@ -2186,59 +2186,59 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         }
 
         if (ret == WS_SUCCESS)
-            ret = wc_HashInit(&ssh->handshake->hash, ssh->handshake->hashId);
+            ret = wc_HashInit(&ssh->handshake->hash, (enum wc_HashType)ssh->handshake->hashId);
```
cconlon: Similar to above, we could make a new variable and make this cast once.  It may look cleaner and be easier to read.

src/internal.c
miyazakh: Thank you for the review. Introduced new variable instead of casting each call.

src/internal.c
miyazakh: By introducing shorter length of variable, they should be within 80-character line limit now.

src/internal.c
miyazakh: Introduced new variable.

PR #168, opened 2019-06-05
Refactor Packet Assembly

ejohnstown:
1. PreparePacket() is to be given an estimated payloadSz, not the actual
payloadSz. The payloadSz should be larger or equal to the actual.
2. BuildPacket() calculates the actual payloadSz based on the position
of idx and value of idx stored before PreparePacket() returns. The size
of the padding is also calculated at this point.

Currently, everything going into a packet needs to be calculated ahead
of time and saved locally until the output buffer is prepared. This
requires saving RSA and ECDSA signatures in large buffers to be copied
later. Now such things can be calculated directly into the output buffer
without the temporary storage and copy.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -5160,26 +5160,20 @@ int SendProtoId(WOLFSSH* ssh)
 }
 
 
+/* payloadSz is an estimate. It should be a worst case. The actual value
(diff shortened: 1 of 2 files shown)


kaleb-himes:
retest this please.

PR #169, opened 2019-06-06
Add Client Side Public Key Authentication

ejohnstown:


src/internal.c
```diff
@@ -6697,11 +6702,441 @@ int SendServiceAccept(WOLFSSH* ssh, byte serviceId)
...
+        word32 idx = 0;
+        ret = wc_EccPrivateKeyDecode(authData->sf.publicKey.privateKey,
+                &idx, &keySig->ks.ecc.key,
+                authData->sf.publicKey.privateKeySz);
+    }
+
+    if (ret == WS_SUCCESS) {
+        if (authData->sf.publicKey.hasSignature) {
+            int sigSz = wc_ecc_sig_size(&keySig->ks.ecc.key);
+            sigSz = 64;
```
cconlon: Should sigSz be forced to 64 here, when the line above it sets it based on wc_ecc_sig_size()?

src/internal.c
```diff
@@ -6697,11 +6702,441 @@ int SendServiceAccept(WOLFSSH* ssh, byte serviceId)
...
+{
+    wc_HashAlg hash;
+    byte digest[WC_MAX_DIGEST_SIZE];
+    word32 digestSz;
+    word32 begin;
+    enum wc_HashType hashId = WC_HASH_TYPE_SHA;
+    int ret = WS_SUCCESS;
+    byte* r;
+    byte* s;
+    byte sig[512];
```
cconlon: Should MAX_ECC_BITS be used here instead of 512?

src/internal.c
ejohnstown: Oh yeah, that was for something I was testing and forgot to remove. Thanks for catching that.

src/internal.c
ejohnstown: It should be big enough to fit a signature for a prime 521 key, so 144 bytes. I think.

ejohnstown: Reduced the sig buffer to 72 bytes, which is the size of a signature for a prime256 key.

PR #170, opened 2019-06-06
Global request

kojo1:
Add Global Request.

### DIFF
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -366,6 +366,28 @@ static THREAD_RET readPeer(void* in)
 }
 #endif /* !SINGLE_THREADED && !WOLFSSL_NUCLEUS */
 
+#if defined(WOLFSSL_PTHREADS) && defined(WOLFSSL_TEST_GLOBAL_REQ)
+
+static int callbackGlobalReq(WOLFSSH *ssh, void *buf, word32 sz, int reply, void *ctx)
+{
+    char reqStr[] = "SampleRequest";
+
+    if ((WOLFSSH *)ssh != *(WOLFSSH **)ctx)
+    {
+        printf("ssh(%x) != ctx(%x)\n", (unsigned int)ssh, (unsigned int)*(WOLFSSH **)ctx);
+        return WS_FATAL_ERROR;
+    }
+
+    if (strlen(reqStr) == sz && (strncmp((char *)buf, reqStr, sz) == 0)
+        && reply == 1){
+        printf("Global Request\n");
+        return WS_SUCCESS;
+    } else {
+        return WS_FATAL_ERROR;
+    }
+
+}
+#endif
 
 THREAD_RETURN WOLFSSH_THREAD client_test(void* args)
 {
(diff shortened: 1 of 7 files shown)


examples/client/client.c
```diff
@@ -366,6 +366,28 @@ static THREAD_RET readPeer(void* in)
 }
 #endif /* !SINGLE_THREADED && !WOLFSSL_NUCLEUS */
 
+#if defined(WOLFSSL_PTHREADS) && defined(WOLFSSL_TEST_GLOBAL_REQ)
+
+static int callbackGlobalReq(WOLFSSH *ssh, void *buf, word32 sz, int reply, void *ctx)
+{
+    char reqStr[] = "SampleRequest";
+
+    if ((WOLFSSH *)ssh != *(WOLFSSH **)ctx)
```
ejohnstown: Is it required that `ctx` is a pointer to a pointer to the WOLFSSH object, or is this just a test that the assigned `ctx` is really being passed to the callback?

ejohnstown: Nevermind. Of course it is. It is in a code block named WOLFSSL_TEST_GLOBAL_REQ.

examples/echoserver/echoserver.c
```diff
@@ -121,12 +121,14 @@ static int dump_stats(thread_ctx_t* ctx)
 
 #define SSH_TIMEOUT 10
 
-static void callbackReqSuccess(WOLFSSH *ssh, void *buf, word32 sz, void *ctx)
+static int callbackReqSuccess(WOLFSSH *ssh, void *buf, word32 sz, void *ctx)
```
ejohnstown: I don't see that the client is responding to the server's Global Request. It isn't set up to anywhere, is it?

src/io.c
```diff
@@ -102,6 +102,11 @@ void* wolfSSH_GetIOWriteCtx(WOLFSSH* ssh)
     return NULL;
 }
 
+void wolfSSH_SetReqSuccess(WOLFSSH_CTX *ctx, WS_CallbackReqSuccess cb)
```
ejohnstown: Can you move this over to ssh.c, please? All the other callbacks are there, except for the I/O callbacks in io.c. The I/O callbacks are here because there will be several default callbacks for various OS.

examples/echoserver/echoserver.c
```diff
@@ -110,6 +117,49 @@ static int dump_stats(thread_ctx_t* ctx)
     return wolfSSH_stream_send(ctx->ssh, (byte*)stats, statsSz);
 }
 
+#if defined(WOLFSSL_PTHREADS) && defined(WOLFSSL_TEST_GLOBAL_REQ)
+
+#define SSH_TIMEOUT 20
+
+static void callbackReqSuccess(WOLFSSH *ssh, void *buf, word32 sz, void *ctx)
```
ejohnstown: I didn't see where this callback is getting tested by the client responding. Is the client responding with its own Global Request?

examples/echoserver/echoserver.c
kojo1: client.c has callbackGlobalReq:371.

src/io.c
kojo1: Moved them to ssh.c

Issue #172, opened 2019-06-24
Possible init bug in BuildUserAuthRequestEcc


LinuxJedi:
rPad and sPad are set to 1 or 0 and then immediately set to 0 afterwards. Not sure which is the correct behaviour yet but I suspect this could be a bug.

JacobBarthelmeh:
Hi @LinuxJedi,

Is this in reference to the rPad and sPad in the wolfSSH file src/internal.c around line 6070? What version of wolfSSH is being used? 

Regards,
Jacob

LinuxJedi:
Hi,
Sorry, I should have narrowed it down more in my first message.

The ones around line 7080 of that file in master branch.

JacobBarthelmeh:
Got it, thanks LinuxJedi. I see the line of code now immediately after the ternary operation that sets both rPad and sPad to 0. Will investigate into it.

LinuxJedi:
That's the one. I would have submitted a PR but I've only had a glance at the code and am not sure yet what the correct behaviour is.

ejohnstown:
I did that one. I'll take a look.

JacobBarthelmeh:
@ejohnstown thanks for taking a look at it, there is a PR here with a possible change for review (https://github.com/wolfSSL/wolfssh/pull/174). Out of curiosity @LinuxJedi what tool was used to find these? Was it found with using scan-build?

LinuxJedi:
Yes, I used scan-build to find these. I hope it was helpful.

JacobBarthelmeh:
We appreciate the reports, it was helpful! And feel free to open any additional issues that you come across. Thank you for confirming the tool used, it will help us with knowing where we can improve continuous integration testing with wolfSSH.

PR #174, opened 2019-06-24
fix for setting rPad/sPad value

JacobBarthelmeh:


JacobBarthelmeh:
Fix for report from https://github.com/wolfSSL/wolfssh/issues/172

PR #178, opened 2019-07-03
More NULL ptr dereference fixes

LinuxJedi:
Found using cppcheck. Also fixed debug logging getting missed from my
previous PR.


LinuxJedi:
I should note that I don't recommend cppcheck for your CI yet. There are several variable scope reductions that would be needed and a few bits of code it currently gets confused over.

kaleb-himes:
**@LinuxJedi is an approved contributor, OK to test.

PR #181, opened 2019-07-05
Remove void cast on non-existent variable in single threaded mode

kaleb-himes:


kaleb-himes:
https://github.com/wolfSSL/meta-wolfssl/issues/9

PR #182, opened 2019-07-24
add back in password auth

JacobBarthelmeh:


ejohnstown:
Remove the public key. This needs to be configurable and able to support both at the same time. Right now you can do publickey or password.

ejohnstown:
Thanks!

PR #193, opened 2019-08-30
Fixes for Cygwin build

cconlon:
This PR fixes two issues with building wolfSSH on Cygwin:

1.  FIONREAD was undefined, Cygwin defines it in <sys/socket.h> instead of <sys/ioctl.h>

2.  Visibility warnings.  Added a check for \_\_CYGWIN\_\_ in visibility.h and rearranged slightly.  These were fixed in wolfSSL proper with https://github.com/wolfSSL/wolfssl/pull/2253

### DIFF
--- a/wolfssh/port.h
+++ b/wolfssh/port.h
@@ -607,10 +607,14 @@ extern "C" {
 #elif defined(USE_WINDOWS_API)
     #define WIOCTL ioctlsocket
 #else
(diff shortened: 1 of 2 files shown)


JacobBarthelmeh:
Thanks Chris, I confirmed it was able to build with Cygwin after the changes.

PR #195, opened 2019-09-07
MSGID_REQUEST_FAILURE to response MSGID_GLOBAL_REQUEST

kojo1:
Response SSH_MSG_REQUEST_FAILURE to Keep-Alive. IETF:draft-ssh-global-requests

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -2970,6 +2970,27 @@ static int DoRequestSuccess(WOLFSSH *ssh, byte *buf, word32 len, word32 *idx)
     
 }
 
+static int DoRequestFailure(WOLFSSH *ssh, byte *buf, word32 len, word32 *idx)
+{
+    word32 dataSz;
+    word32 begin = *idx;
+    int ret = WS_SUCCESS;
+
+    (void)ssh;
+    (void)len;
+
+    WLOG(WS_LOG_DEBUG, "DoRequestFalure, *idx=%d, len=%d", *idx, len);
+    ato32(buf + begin, &dataSz);
+    begin += LENGTH_SZ + dataSz;
+
+    if (ssh->ctx->reqFailureCb != NULL)
+        ret = ssh->ctx->reqFailureCb(ssh, &(buf[*idx]), len, ssh->reqFailureCtx);
+
+    *idx = begin;
+
+    return ret;
+}
+
 static int DoDebug(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 {
     byte  alwaysDisplay;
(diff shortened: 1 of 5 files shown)


src/internal.c
```diff
@@ -7579,6 +7601,38 @@ int SendRequestSuccess(WOLFSSH* ssh, int success)
...
+    int ret = WS_SUCCESS;
+
+    WLOG(WS_LOG_DEBUG, "Entering SendRequestFailure");
+
+    if (ssh == NULL)
+        ret = WS_BAD_ARGUMENT;
+
+    if (ret == WS_SUCCESS)
+        ret = PreparePacket(ssh, MSG_ID_SZ);
+    printf("BundlePacket, ret=%d\n", ret);
```
ejohnstown: Debug print that should be removed.

src/internal.c
```diff
@@ -7579,6 +7601,38 @@ int SendRequestSuccess(WOLFSSH* ssh, int success)
...
+    if (ret == WS_SUCCESS)
+    {
+        output = ssh->outputBuffer.buffer;
+        idx = ssh->outputBuffer.length;
+
+        output[idx++] = MSGID_REQUEST_FAILURE;
+
+        ssh->outputBuffer.length = idx;
+        printf("BundlePacket\n");
```
ejohnstown: Another debug print.

src/internal.c
```diff
@@ -7569,71 +7569,40 @@ int SendUserAuthBanner(WOLFSSH* ssh)
...
-
-    if (ret == WS_SUCCESS)
-        ret = wolfSSH_SendPacket(ssh);
-
-    WLOG(WS_LOG_DEBUG, "Leaving SendRequestSuccess(), ret = %d", ret);
-    return ret;
-}
-
-int SendRequestFailure(WOLFSSH *ssh)
+int SendRequestSuccess(WOLFSSH *ssh, int success)
```
ejohnstown: Can we treat the success parameter as a boolean? We are currently using it with an error code enumeration (WS_SUCCESS) that is getting complemented (~WS_SUCCESS).

src/internal.c
```diff
@@ -3869,7 +3869,7 @@ static int DoGlobalRequest(WOLFSSH* ssh,
         *idx += len;
 
         if (wantReply)
-            ret = SendRequestFailure(ssh);
+            ret = SendRequestSuccess(ssh, ~WS_SUCCESS);
```
ejohnstown: If the second parameter is treated as boolean, ~WS_SUCCESS can just be 0. WS_SUCCESS is supposed to be a named return code, not treated as a number.

src/internal.c
kojo1: Done

src/internal.c
kojo1: Done

kojo1:
Use needs is just for the failure for now. Put both of them,  just in case.

PR #196, opened 2019-09-09
Rejecting password

kojo1:
don't send AuthFailure on rejecting password


ejohnstown:
Why are you changing this? RFC 4252, section 8 on password authentication says when the client sends the SSH_MSG_USERUATH_REQUEST message, "the server responds to this message with success or failure."

PR #198, opened 2019-09-16
SendDisconnect at password reject

kojo1:
example code in echoserver


ejohnstown:
Can you reset the passwdRetry on the start of a new connection? When you fail the password twice, and then reconnect, it doesn't let you fail three times again.

ejohnstown:
I'm rebasing this to master, fixing the conflicts, adding the requested change. It is replace by PR #208.

PR #201, opened 2019-10-09
Echoserver Maintenance

ejohnstown:
1. When printing server worker errors, print out the error string on the stored error code, not the return value.
2. If the error is a socket error, don't try to send a shutdown message to the peer.
3. Update the echoserver to close the connection cleanly when the SFTP peer exits.
4. Added a function to check a WOLFSSL_CHANNEL if it is in eof.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -2104,7 +2104,7 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
         word32 nameLen = (word32)WSTRLEN(dirName);
 
         if (nameLen > MAX_PATH - 3) {
-            WLOG(WS_LOG_DEBUG, "Path name is too long.");
+            WLOG(WS_LOG_SFTP, "Path name is too long.");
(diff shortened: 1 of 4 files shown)


src/wolfsftp.c
```diff
@@ -1007,6 +1010,8 @@ int wolfSSH_SFTP_read(WOLFSSH* ssh)
     int maxSz, ret = WS_SUCCESS;
     WS_SFTP_RECV_STATE* state = NULL;
 
+    WLOG(WS_LOG_DEBUG, "Entering wolfSSH_SFTP_read()");
```
JacobBarthelmeh: please change to be WS_LOG_SFTP log type. Trying to keep all SFTP debugging unique so a quick filter on the connections debugging output using WS_LOG_SFTP will show all SFTP related debug messages.

ejohnstown: What I'd like to see is the logging have a level (WARN, TRACE, DEBUG, etc) and a domain (KEX, SFTP, SCP, etc). Right now it is a bad mashing of the two and the pattern not followed at all.

PR #202, opened 2019-10-10
fix debuging level type

JacobBarthelmeh:


ejohnstown:
I was thinking the WLOG would take a new parameter, domain of type enum wolfSSH_LogDomain. They can be WS_LOG_GENERAL, WS_LOG_SFTP, WS_LOG_SCP, WS_LOG_KEX, WS_LOG_AUTH, etc.

Then we can add a new callback that shadows the existing one that gets the domain value. The old one can be deprecated, but still available.

JacobBarthelmeh:
> I was thinking the WLOG would take a new parameter, domain of type enum wolfSSH_LogDomain. They can be WS_LOG_GENERAL, WS_LOG_SFTP, WS_LOG_SCP, WS_LOG_KEX, WS_LOG_AUTH, etc.
> 
> Then we can add a new callback that shadows the existing one that gets the domain value. The old one can be deprecated, but still available.

Thanks John, that makes sense.

PR #204, opened 2019-10-23
SFTP port to MQX 4.2 (MQX/MFS/RTCS)

cconlon:
This PR adds a port of wolfSSH and SFTP (client and server) to Freescale/NXP MQX 4.2 (Classic).  The compiler used for this port was IAR-EWARM 7.50.2.  Changes include:

- MQX 4.2 SFTP port
- Fix to loop bound on wolfSSH_oct2dec()
- Fix IAR-EWARM compiler warnings

MQX file system operations require that the file system handle be available in addition to the file handle or file name.  To assist in this, a new API has been added that applications running on MQX will need to use to set the general file system handle:

```
wolfSSH_SetFilesystemHandle()
```

NOTE: SCP has not been ported to MQX.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -716,8 +716,8 @@ static int SFTP_SetAttributes(WOLFSSH* ssh, byte* buf, word32 bufSz,
 
 #ifndef NO_WOLFSSH_SERVER
 
-static int SFTP_GetAttributes(const char* fileName, WS_SFTP_FILEATRB* atr,
-        byte link, void* heap);
+static int SFTP_GetAttributes(void* fs, const char* fileName,
+        WS_SFTP_FILEATRB* atr, byte link, void* heap);
 static int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
         WS_SFTP_FILEATRB* atr);
 
@@ -926,7 +926,7 @@ static int wolfSSH_SFTP_RecvRealPath(WOLFSSH* ssh, int reqId, byte* data,
         }
         else {
         #ifndef USE_WINDOWS_API
-            if (WGETCWD(wd, WOLFSSH_MAX_FILENAME) == NULL) {
+            if (WGETCWD(ssh->fs, wd, WOLFSSH_MAX_FILENAME) == NULL) {
                 WLOG(WS_LOG_SFTP, "Unable to get current working directory");
                 if (wolfSSH_SFTP_CreateStatus(ssh, WOLFSSH_FTP_FAILURE, reqId,
                         "Directory error", "English", NULL, &outSz)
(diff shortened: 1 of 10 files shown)


src/wolfsftp.c
```diff
@@ -3561,19 +3648,140 @@ int SFTP_GetAttributes(const char* fileName, WS_SFTP_FILEATRB* atr, byte link,
...
+        atr->flags |= WOLFSSH_FILEATRB_PERM;
+        atr->per |= 0x4000;
+        return WS_SUCCESS;
+    }
+
+    search.ATTRIBUTE = MFS_SEARCH_ANY;
+    search.WILDCARD = (char*)fileName;
+    search.SEARCH_DATA_PTR = &search_data;
+
+    err = ioctl(mfs_ptr, IO_IOCTL_FIND_FIRST_FILE, (uint32_t*)&search);
```
JacobBarthelmeh: can we wrap this in WIOCTL?

src/wolfsftp.c
```diff
@@ -3561,19 +3648,140 @@ int SFTP_GetAttributes(const char* fileName, WS_SFTP_FILEATRB* atr, byte link,
...
+    int err, sz;
+    MQX_FILE_PTR mfs_ptr;
+    MFS_SEARCH_DATA search_data;
+    MFS_SEARCH_PARAM search;
+
+    if (fileName == NULL || atr == NULL) {
+        return WS_BAD_FILE_E;
+    }
+    mfs_ptr = (MQX_FILE_PTR)fs;
+    sz = (int)WSTRLEN(fileName);
```
JacobBarthelmeh: check that sz is at least 3 or greater for when doing the WSTRNCMP later (fileName + sz - 3)

src/wolfsftp.c
```diff
@@ -3561,19 +3648,140 @@ int SFTP_GetAttributes(const char* fileName, WS_SFTP_FILEATRB* atr, byte link,
...
+    if (cur == NULL) {
+        WLOG(WS_LOG_SFTP, "Unknown handle");
+        return WS_BAD_FILE_E;
+    }
+
+    search.ATTRIBUTE = MFS_SEARCH_ANY;
+    search.WILDCARD = cur->name;
+    search.SEARCH_DATA_PTR = &search_data;
+
+    err = ioctl(mfs_ptr, IO_IOCTL_FIND_FIRST_FILE, (uint32_t*)&search);
```
JacobBarthelmeh: wrap in WIOCTL if possible

src/wolfsftp.c
cconlon: This may be challenging, as ioctl here accepts a MQX_FILE_PTR structure as the first argument, where when wolfSSH ssh.c calls WIOCTL for socket operations, it passes an "int fd" as the first argument.  We may need a separate abstraction for one of these use cases if we want to abstract this call out.

src/wolfsftp.c
cconlon: See comment above.

src/wolfsftp.c
cconlon: Good catch, will change.

cconlon: Fixed

PR #213, opened 2019-10-25
be more lenient on CR with protocol exchange

JacobBarthelmeh:


ejohnstown:
When I enable WOLFSSH_STRICT_PROTOCOL_EXCHANGE, and send a protocol ID string that doesn't end in CRLF, the code hangs. It looks like it is waiting for the CRLF explicitly. (I denormalized the code to only send the LF in the protocol ID string.)

JacobBarthelmeh:
> When I enable WOLFSSH_STRICT_PROTOCOL_EXCHANGE, and send a protocol ID string that doesn't end in CRLF, the code hangs. It looks like it is waiting for the CRLF explicitly. (I denormalized the code to only send the LF in the protocol ID string.)

Thanks for the review. Looks like that hang is in master too. The while loop keeps looking for the \r\n I almost think we should remove the while loop here and report back to calling functions instead?

JacobBarthelmeh:
Conversed with John about advancement of IDX being 1 instead of default 2 in the case of just '\n' without '\r\n'. Am working on a code change to account for this.

Issue #221, opened 2019-11-29
Compilation errors when compiling wolfSSH for 32 bit


guidovranken:
I cannot (cross)compile wolfSSH for 32 bit (x86).

I configure using ```setarch i386 ./configure --enable-static --enable-all``` and ```CFLAGS``` contains ```-m32```.

```
src/port.c:125:49: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
            long offset = ((long)shortOffset[1] << 32) | shortOffset[0];
                                                ^  ~~
src/port.c:133:49: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
            long offset = ((long)shortOffset[1] << 32) | shortOffset[0];
                                                ^  ~~
2 errors generated.
src/wolfscp.c:1568:25: error: implicit conversion loses integer precision: 'word64' (aka 'unsigned long long') to '__time_t' (aka 'long') [-Werror,-Wshorten-64-to-32]
        tmp[0].tv_sec = aTime;Makefile:1231: recipe for target 'src/src_libwolfssh_la-port.lo' failed
make[1]: *** [src/src_libwolfssh_la-port.lo] Error 1

make[1]: *** Waiting for unfinished jobs....
                      ~ ^~~~~
src/wolfscp.c:1570:25: error: implicit conversion loses integer precision: 'word64' (aka 'unsigned long long') to '__time_t' (aka 'long') [-Werror,-Wshorten-64-to-32]
        tmp[1].tv_sec = mTime;
                      ~ ^~~~~
2 errors generated.
Makefile:1252: recipe for target 'src/src_libwolfssh_la-wolfscp.lo' failed
make[1]: *** [src/src_libwolfssh_la-wolfscp.lo] Error 1
src/wolfsftp.c:3810:42: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
    atr->sz[1] = (word32)((stats.st_size >> 32) & 0xFFFFFFFF);
                                         ^  ~~
src/wolfsftp.c:3852:42: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
    atr->sz[1] = (word32)((stats.st_size >> 32) & 0xFFFFFFFF);
                                         ^  ~~
src/wolfsftp.c:7489:39: error: implicit conversion loses integer precision: 'word64' (aka 'unsigned long long') to 'long' [-Werror,-Wshorten-64-to-32]
                    WFSEEK(state->fl, offset, 0);
[...]

PR #228, opened 2019-12-05
use off_t type with pwrite/pread

JacobBarthelmeh:
for Issue https://github.com/wolfSSL/wolfssh/issues/221

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,7 @@ AC_PROG_INSTALL
 AC_HEADER_STDC
 AC_CHECK_SIZEOF([long long])
 AC_CHECK_SIZEOF([long])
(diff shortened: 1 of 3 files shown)


JacobBarthelmeh:
System was testing on had off_t be a 64 bit type even when compiling for 32 bit. Also it looks like it can be forced with gcc if defining the macro -D_FILE_OFFSET_BITS=64

ejohnstown:
I tried building this on a 32-bit Ubuntu VM, and the fix still failed to build with the same error.

JacobBarthelmeh:
Thanks for the review John, I added a check for off_t size to configure.ac and retested on a local 32bit Ubuntu VM building with ./configure --enable-all && make check

Issue #233, opened 2020-01-15
how do I enable publickey auth in my SFTP client code?


bagder:
Hi again!

We've now merged initial support for wolfSSH in curl. It however struck me I haven't fixed publickey auth in the curl use of wolfssh (https://github.com/curl/curl/issues/4820), and I find the docs rather lacking in this aspect.

What do I need to do to enable publickey auth? In my test case the server accepts publickey auth but my callback set with `wolfSSH_SetUserAuth()` gets called numerous times but only with `authtype == WOLFSSH_USERAUTH_PASSWORD`:

https://github.com/curl/curl/blob/68403cdbc6325bcbe4e4128c263db5d04a3c77c6/lib/vssh/wolfssh.c#L393-L395

(A secondary issue is that wolfssh git master doesn't build for me after I updated this afternoon (thinking maybe I needed an update), which is surprising since your CI builds all look green but I'll save that for now.)

JacobBarthelmeh:
Hey Daniel,

That's great with the initial support merged in!

Thanks for the report, I see the curl test number in the link. I did a quick sanity check on the password based example bundled in wolfSSH and it went okay, so looks like I will need to dig into it a bit more to find the difference.

```
cd wolfssh
./examples/server/server
ssh -i ./keys/hansel-key-rsa.pem -p 22222 hansel@127.0.0.1
```

Regards,
Jacob

JacobBarthelmeh:
I opened a PR with changes I did for resolving linking while investigating this.

JacobBarthelmeh:
After more investigation this looks like something we need to change in wolfSSH. Dealing with how the flag for public key connections on the client side get handled. You can change this location of code (https://github.com/wolfSSL/wolfssh/blob/master/src/internal.c#L3813) for testing but we will work on a better runtime way for setting public key on the client side

PR #235, opened 2020-01-27
public key vs password is now decision of user callback

JacobBarthelmeh:
Allow user callback to decide if password or public key https://github.com/wolfSSL/wolfssh/issues/233

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3810,8 +3810,6 @@ static int DoUserAuthFailure(WOLFSSH* ssh,
     byte authList[3]; /* Should only ever be password, publickey, hostname */
     word32 authListSz = 3;
     byte partialSuccess;
-    byte authId = ID_USERAUTH_PASSWORD;
-        /* To use public key authentication, change authId. */
     int ret = WS_SUCCESS;
 
     WLOG(WS_LOG_DEBUG, "Entering DoUserAuthFailure()");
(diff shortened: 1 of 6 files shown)


src/internal.c
```diff
@@ -7386,24 +7385,29 @@ int SendUserAuthRequest(WOLFSSH* ssh, byte authId, int addSig)
             WLOG(WS_LOG_DEBUG, "SUAR: Calling the userauth callback");
 
             WMEMSET(&authData, 0, sizeof(authData));
-            authData.type = authId;
 
-            if (authId == ID_USERAUTH_PASSWORD) {
+            authData.username = (const byte*)ssh->userName;
```
JacobBarthelmeh: Now sets the username before calling user callback

src/internal.c
```diff
@@ -7386,24 +7385,29 @@ int SendUserAuthRequest(WOLFSSH* ssh, byte authId, int addSig)
...
+            authData.usernameSz = ssh->userNameSz;
+
+            ret = ssh->ctx->userAuthCb(WOLFSSH_USERAUTH_PUBLICKEY,
```
JacobBarthelmeh: automatically tries calling user callback with public key type first, if callback rejects it then tries with password

src/internal.c
```diff
@@ -7386,24 +7385,29 @@ int SendUserAuthRequest(WOLFSSH* ssh, byte authId, int addSig)
...
                 }
-            }
-            else if (authId == ID_USERAUTH_PUBLICKEY) {
-                ret = ssh->ctx->userAuthCb(WOLFSSH_USERAUTH_PUBLICKEY,
-                        &authData, ssh->userAuthCtx);
-                if (ret != WOLFSSH_USERAUTH_SUCCESS) {
-                    WLOG(WS_LOG_DEBUG, "SUAR: Couldn't get key");
-                    ret = WS_FATAL_ERROR;
+                else {
+                    authId = ID_USERAUTH_PASSWORD;
```
JacobBarthelmeh: sets authId based on success of user callback

JacobBarthelmeh:
Got feedback to check that the client sends NONE first then decides if using password/public key based on server response of whats supported

PR #238, opened 2020-02-07
add SIZEOF_OFF_T guess to wolfsftp.c

JacobBarthelmeh:


src/wolfsftp.c
```diff
@@ -38,6 +38,17 @@
 #endif
 
 
+#ifndef SIZEOF_OFF_T
```
ejohnstown: If you include port.h in wolfsftp.c, you won't need this block here.

JacobBarthelmeh: updated

PR #241, opened 2020-02-11
exit channel in echoserver before sending disconnect

JacobBarthelmeh:
ZD 9884

Fix for when testing scp against echoserver. This change makes the return value of scp client from OpenSSH return 0 on a successful transfer.

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -401,8 +401,20 @@ static THREAD_RETURN WOLFSSH_THREAD server_worker(void* vArgs)
 
     switch (ret) {
         case WS_SCP_COMPLETE:
-            printf("scp file transfer completed\n");
-            ret = 0;
+            {
+                byte buf[1];
+                printf("scp file transfer completed\n");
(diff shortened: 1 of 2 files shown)


ejohnstown:
I'm testing with OpenSSH on macOS Catalina. (OpenSSH_7.9p1, LibreSSL 2.7.3)

    $ scp -P 22222 ulysses.txt jill@localhost:. ; echo $?
    wolfSSH Example Echo Server
    jill@localhost's password: 
    ulysses.txt                                                                                       100% 1549KB  49.6MB/s   00:00    
    channel_by_id: 0: bad id: channel free
    Disconnecting 127.0.0.1 port 22222: ieof packet referred to nonexistent channel 0
    1

The server is just the echoserver. Without the patch, it looks like it closes nicely, but still also returns a 1.

kaleb-himes:
@JacobBarthelmeh @ejohnstown,

ZD 9884: Customer reports these changes fix the issue they were facing.

- KH

ejohnstown:
Adding a sleep(1) between the wolfSSH_stream_exit() and wolfSSL_shutdown() made it work.

PR #243, opened 2020-02-12
VxWorks Customizations

ejohnstown:
Also includes some bug fixes.
1. Add a function that returns a pointer to a session's username string.
2. Add a function to set a "raw" username copied from an incoming packet.
3. Add tests for 1 and 2.
4. Server saves a copy of the username on successful authentication.
5. Fix bug with GetString() where it would choke on a 0 length string.
6. Clean up the termios macros and mode creation.
7. Add some options for when WOLFSSL_VXWORKS is set.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1191,38 +1191,61 @@ int wolfSSH_SetChannelType(WOLFSSH* ssh, byte type, byte* name, word32 nameSz)
     return WS_SUCCESS;
 }
 
-int wolfSSH_SetUsername(WOLFSSH* ssh, const char* username)
+
+int wolfSSH_SetUsernameRaw(WOLFSSH* ssh,
+        const byte* username, word32 usernameSz)
 {
-    char* value = NULL;
-    word32 valueSz = 0;
+    char* newUsername = NULL;
     int ret = WS_SUCCESS;
 
-    if (ssh == NULL || ssh->handshake == NULL ||
-        ssh->ctx->side == WOLFSSH_ENDPOINT_SERVER ||
-        username == NULL) {
-
+    if (ssh == NULL)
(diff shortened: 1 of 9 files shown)


src/ssh.c
```diff
@@ -1191,41 +1191,59 @@ int wolfSSH_SetChannelType(WOLFSSH* ssh, byte type, byte* name, word32 nameSz)
...
+        newUsername = (char*)WMALLOC(usernameSz + 1,
+                ssh->ctx->heap, DYNTYPE_STRING);
+        if (newUsername == NULL)
             ret = WS_MEMORY_E;
+        else {
+            WMEMCPY(newUsername, username, usernameSz);
+            newUsername[usernameSz] = 0;
+            if (ssh->userName != NULL)
+                WFREE(ssh->userName, ssh->ctx->heap, DYNTYPE_STRING);
+            ssh->userName = newUsername;
```
JacobBarthelmeh: case here of a leak if called twice in a row without free'ing the previous pointer

JacobBarthelmeh: also should we make sure a null terminator is set here? hard setting the last value to '\0' for safety

src/ssh.c
```diff
@@ -1191,41 +1191,59 @@ int wolfSSH_SetChannelType(WOLFSSH* ssh, byte type, byte* name, word32 nameSz)
...
+        ret = wolfSSH_SetUsernameRaw(ssh,
+                (const byte*)username, (word32)WSTRLEN(username));
     }
 
     return ret;
 }
 
 
+char* wolfSSH_GetUsername(WOLFSSH* ssh)
+{
```
JacobBarthelmeh: Please add a sanity check for a null pointer on ssh

PR #248, opened 2020-03-02
wolfSCP

ejohnstown:
Add support for an example SCP client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,7 @@ examples/echoserver/echoserver
 examples/server/server
 examples/portfwd/portfwd
 examples/sftpclient/wolfsftp
+examples/scpclient/wolfscp
 
 # test output
 tests/*.test
--- a/examples/include.am
+++ b/examples/include.am
@@ -7,3 +7,4 @@ include examples/server/include.am
 include examples/echoserver/include.am
 include examples/portfwd/include.am
 include examples/sftpclient/include.am
+include examples/scpclient/include.am
--- a/examples/scpclient/include.am
+++ b/examples/scpclient/include.am
@@ -1,0 +1,14 @@
+# vim:ft=automake
+# All paths should be given relative to the root
+
+if BUILD_SCP
+if BUILD_EXAMPLE_CLIENTS
+noinst_PROGRAMS += examples/scpclient/wolfscp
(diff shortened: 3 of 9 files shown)


ejohnstown:
$ ./examples/echoserver/echoserver -d .
$ ./examples/scpclient/wolfscp -u jill -L ./test.txt:.

This will copy the file test.txt from the current directory to the echoserver's current directory. The paths need to be present in the commands. Note, if sending a larger file, you need to set the default window size larger.

JacobBarthelmeh:
Hit a memory leak with the test case.

```
$ ./scripts/get-put.test 

=================================================================
==90682==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 88 byte(s) in 1 object(s) allocated from:
    #0 0x1018767ef  (/usr/local/opt/gcc/lib/gcc/9/libasan.5.dylib:x86_64+0x7b7ef)
    #1 0x1015af263  (/Users/sweetness/Documents/wolfssh/src/.libs/libwolfssh.9.dylib:x86_64+0x3f263)
```

JacobBarthelmeh:
Is rename of file currently not supported? i.e 

```
$ ./examples/scpclient/wolfscp -p 22 -H 192.168.1.149 -u pi2 -S /home/pi2/configure.ac:./new.ac
Password: 
wolfSSH error: Couldn't copy the file.
```

ejohnstown:
I'm looking at the memory leak.

The rename isn't working. But, it isn't working without the PR using just wolfSCP server. I have this on the maintenance to-do list.

ejohnstown:
I'm using commit 805034b of wolfSSL, configured with

    $ ./configure CC="clang -fsanitize=address" --enable-all --prefix=/opt/wolfssl-all --enable-static

and using the PR version of wolfSSH configured with:

    $ ./configure CC="clang -fsanitize=address" --enable-all LDFLAGS=-L/opt/wolfssl-all/lib CPPFLAGS=-I/opt/wolfssl-all/include --disable-shared

I run the test script and I'm not getting any reports. It also looks like that test case is SFTP only.

JacobBarthelmeh:
I think the issues I'm reporting here are unrelated to the this PR. Am in the process of debugging them separately.

PR #252, opened 2020-03-12
Window

ejohnstown:
Changed the send wait for window updates.


JacobBarthelmeh:
Looks great. Only thing is if wolfSSH_CheckReceivePending(WOLFSSH* ssh) gets removed it should also be removed from wolfssh/ssh.h.

PR #253, opened 2020-03-23
Rename

ejohnstown:
Client should get the path and filename processed like the server does.


src/wolfscp.c
```diff
@@ -1570,6 +1570,8 @@ static int wolfSSH_SCP_cmd(WOLFSSH* ssh, const char* localName,
             ret = DoScpSource(ssh);
         }
         else {
+            cmdSz = (word32)WSTRLEN(localName);
+            ParseBasePathHelper(ssh, cmdSz);
```
JacobBarthelmeh: check on return value here

PR #257, opened 2020-05-08
Fuzz Test Fix

ejohnstown:
Fixed an issue where a too large GEX prime group size would cause problems generating the shared secret.
1. Moved the #defines for setting the default DH min/preferred/max values to internal.h.
2. Based the DH e value on the DH max size.
3. Check that the received prime group size is not greater than the max.


JacobBarthelmeh:
Tested against the internal fuzzer report. Resolved the issue, thanks John.

PR #278, opened 2020-09-01
Fixed one time inclusion in header files

jonknoll:
Added #ifndef FILENAME #define FILENAME to include files, and removed #pragma once because it is not recognized by all compilers.

### DIFF
--- a/examples/client/client.h
+++ b/examples/client/client.h
@@ -19,8 +19,6 @@
  */
 
 
-#pragma once
-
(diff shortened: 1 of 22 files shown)


ejohnstown:
Thank you for the change suggestion. We'll handle this separately in our own way. I don't remember why I used the pragma, probably too much coding on a Mac. Using the indef guard follows the pattern we are using in other projects.

PR #279, opened 2020-09-01
Small bugfix in GetStringAlloc() that causes a compiler error.

jonknoll:


ejohnstown:
Thank you for the bug report. This has been fixed in PR #281 along with some other issues found independently.

PR #280, opened 2020-09-02
Fix ID List

ejohnstown:
1. Increase the size of the processed ID list from the peer from 8 to 16. (ZD 9820)


JacobBarthelmeh:
Can we make this dynamic? I had a similar case pop up and think a dynamic list would help with future cases

ejohnstown:
> Can we make this dynamic? I had a similar case pop up and think a dynamic list would help with future cases

I think that's going to complicate it. And we'll end up allocating 8 or 16 bytes. I think that whole section needs a bit of an overhaul anyway.

JacobBarthelmeh (approved):

PR #284, opened 2020-09-17
add example timeout to sftp example

JacobBarthelmeh:
ZD 10833

### DIFF
--- a/examples/sftpclient/sftpclient.c
+++ b/examples/sftpclient/sftpclient.c
@@ -56,26 +56,27 @@ static void err_msg(const char* s)
     #include <sys/time.h>
 
     static char   currentFile[WOLFSSH_MAX_FILENAME+1] = "";
-    static double startTime;
+    static word32 startTime;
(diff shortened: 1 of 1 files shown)


examples/sftpclient/sftpclient.c
```diff
@@ -51,12 +51,54 @@ static void err_msg(const char* s)
...
+
+#ifndef WOLFSSH_NO_TIMESTAMP
+    #include <sys/time.h>
+
+    static char   currentFile[WOLFSSH_MAX_FILENAME+1] = "";
+    static double startTime;
+    #define TIMEOUT_VALUE 120
+
+    double current_time(int);
+    double current_time(int reset)
```
ejohnstown, line 63: Does this have to be a double? Some environments pull in a floating point math library to handle floating point math because the processor doesn't support it. With a word32, and µs resolution, you still get over a day of possible timeout time.

ejohnstown (changes requested):

ejohnstown (approved):

PR #286, opened 2020-09-24
Include File Guards

ejohnstown:
Change file including guard from pragma-once to ifndef-define. This follows the pattern from our other projects.


JacobBarthelmeh:
Let's extend the pattern to the examples also please

JacobBarthelmeh (approved):

PR #291, opened 2020-10-09
add special case for real path and refactor clean_path

JacobBarthelmeh:
ZD 10852


ejohnstown:
Can you add a couple test cases to the unit test?

JacobBarthelmeh:
This also fixes issue 27325 by adding a null check on input argument to clean_path (named wolfSSH_CleanPath now).

PR #295, opened 2020-10-27
in SendKexDhReply(), clear sigKeyBlock before use (fixes ZD11139).

douzzer:


src/internal.c
```diff
@@ -6436,6 +6436,8 @@ int SendKexDhReply(WOLFSSH* ssh)
         return WS_BAD_ARGUMENT;
     }
 
+    memset(&sigKeyBlock, 0, sizeof sigKeyBlock);
```
ejohnstown, line 6439: WMEMSET(..., sizeof(sigKeyBlock));

Mainly so it looks like everything else, more or less. And it'll use the memset wrapper.

ejohnstown (changes requested):

src/internal.c
douzzer: d'oh!

done.

ejohnstown (approved):

PR #298, opened 2020-11-13
More Options

ejohnstown:
1. Added general disable flags for RSA and ECDSA.
2. Replaced HAVE_ECC, NO_RSA, NO_DSA with the general disable flags.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -85,8 +85,14 @@ Flags:
   WOLFSSH_NO_ECDH_SHA2_ED25519
     Set when ED25519 or SHA2-256 are disabled. Set to disable use of ECDHE key
     exchange with prime ED25519. (It just decodes the ID for output.)
+  WOLFSSH_NO_RSA
+    Set when RSA is disabled. Set to disable use of RSA server and user
+    authentication.
   WOLFSSH_NO_SSH_RSA_SHA1
     Set when RSA or SHA1 are disabled. Set to disable use of RSA server
+    authentication.
+  WOLFSSH_NO_ECDSA
+    Set when ECC is disabled. Set to disable use of ECDSA server and user
     authentication.
   WOLFSSH_NO_ECDSA_SHA2_NISTP256
     Set when ECC or SHA2-256 are disabled. Set to disable use of ECDSA server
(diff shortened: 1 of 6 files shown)


src/internal.c
```diff
@@ -4097,8 +4107,10 @@ static int DoUserAuthRequestEcc(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
     WLOG(WS_LOG_DEBUG, "Leaving DoUserAuthRequestEcc(), ret = %d", ret);
     return ret;
 }
+#endif
 
 
+#if !defined(WOLFSSH_NO_RSA) && !defined(WOLFSSH_NO_ECDSA)
```
JacobBarthelmeh, line 4113:
```
src/internal.c:8341:19: error: implicit declaration of function
      'PrepareUserAuthRequestPublicKey' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
            ret = PrepareUserAuthRequestPublicKey(ssh, &payloadSz, &authData,
```

with

./configure --enable-all CFLAGS=-DWOLFSSH_NO_RSA

src/internal.c
```diff
@@ -6409,6 +6428,7 @@ int SendKexDhReply(WOLFSSH* ssh)
                 byte nPad;
             } rsa;
 #endif
+#ifndef WOLFSSH_NO_ECDH_SHA2_NISTP256
```
JacobBarthelmeh, line 6431:
```
src/internal.c:3032:50: error: no member named 'ecc' in 'union (anonymous at
      src/internal.c:2780:9)'
            ret = wc_ecc_init_ex(&sigKeyBlock.sk.ecc.key, ssh->ctx->heap,
```

with

./configure --enable-all CFLAGS=-DWOLFSSH_NO_ECC

JacobBarthelmeh (commented):
Couple of notes:

1) The undef macro before def of macro (that could go in a separate PR if wanted since is not code directly changed in this one)

2) It would be nice to have a safety net for users that try to disable all public key operations. Trying WOLFSSH_NO_ECDSA and WOLFSSH_NO_RSA at the same time. This is a nice to have though....

PR #300, opened 2020-12-04
Use the internal version of `strdup `

dgarske:
For portability this is better. Example: Keil libc doesn't have `strdup `.
ZD 11320


dgarske:
@ejohnstown this is ready to go. The `wstrdup` function is already in the code...

PR #304, opened 2021-01-08
Use WSTRNCMP for comparing session command string

guidovranken:
Resolves OSS-Fuzz issue 27298 (see ZD 11429)


JacobBarthelmeh:
Ok to test

JacobBarthelmeh (approved):

PR #308, opened 2021-01-12
Fix for RSA e/n max size checking

dgarske:
ZD11424


dgarske:
Duplicate fix. Already fixed in PR #303

PR #315, opened 2021-01-27
adjust length for code standard

JacobBarthelmeh:
Using to test CI updates


JacobBarthelmeh:
The updated CI has a compiler that is hitting

```
error: ‘fallthrough’ attribute not followed by ‘;’ [-Werror=attributes]
             #define FALL_THROUGH __attribute__ ((fallthrough))
                                  ^~~~~~~~~~~~~
src/internal.c:1618:21: note: in expansion of macro ‘FALL_THROUGH’
                     FALL_THROUGH
                     ^~~~~~~~~~~~
src/internal.c:1619:13: error: expected identifier or ‘(’ before ‘}’ token
             }
```
adding a commit to fix it here

PR #317, opened 2021-02-02
In ReceiveScpConfirmation reserve room for NULL-terminator.

guidovranken:
ZD 11621


dgarske:
OK to test

JacobBarthelmeh:
Tested with reproducer, thanks @guidovranken

PR #320, opened 2021-02-02
Get Size

ejohnstown:
1. Add a function GetSize() that calls GetUint32() then checks that the value read in plus the data index is still less than the data length.
2. Replaced a few checks of the size of some data with calls to GetSize(). Included are public key type length, public key length, and the signature length in DoUserAuthPublicKey().

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -1770,6 +1770,21 @@ int GetUint32(word32* v, const byte* buf, word32 len, word32* idx)
 }
 
 
+int GetSize(word32* v, const byte* buf, word32 len, word32* idx)
+{
+    int result;
+
+    result = GetUint32(v, buf, len, idx);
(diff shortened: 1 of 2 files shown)


src/internal.c
```diff
@@ -1770,6 +1770,21 @@ int GetUint32(word32* v, const byte* buf, word32 len, word32* idx)
 }
 
 
+int GetSize(word32* v, const byte* buf, word32 len, word32* idx)
+{
+    int result;
+
+    result = GetUint32(v, buf, len, idx);
+    if (result == WS_SUCCESS) {
+        if (*v + *idx > len) {
```
JacobBarthelmeh, line 1779: Can we adjust this here to match the previous comparisons? I know previously some of them were set up to avoid overflows and suspect this is one of the cases. The change would be "(*v > len - *idx)".

PR #324, opened 2021-02-16
SCP: improvements to run on embedded RTOS

danielinux:
Suggest to remove dependencies from snprintf for SCP in server-only mode. 

In fact, `snprintf` calls the libc allocator without the possibility to override using XMALLOC/WMALLOC. When running on an embedded OS with non-standard `malloc` (e.g. freeRTOS), this results in link errors.

The patchset makes the SCP server more embedded friendly through the following changes:

### log.c:
- Removed definition of LOG functions when DEBUG is disabled

### wolfssh/port.h:
- When compiled with `WOLFSSH_USER_IO` don't fall-back to default system calls

### wolfscp.c:
   - ~~added NO_SCP_TIMESTAMP~~
   - ~~added NO_SCP_ENTER_DIRECTORY~~
   - added WSCPFILEHDR compile option to override the default file header function
   - moved some functions out of the server-only build (via NO_SSH_CLIENT)
   - ~~calling ScpCheckForRename only when NO_FILESYSTEM=0~~


src/wolfscp.c
```diff
@@ -1035,15 +1042,15 @@ static int GetScpTimestamp(WOLFSSH* ssh, byte* buf, word32 bufSz,
...
+#ifndef NO_FILESYSTEM
 /* checks for if directory is being renamed in command
  *
  * returns WS_SUCCESS on success
  */
 static int ScpCheckForRename(WOLFSSH* ssh, int cmdSz)
 {
     /* case of file, not directory */
-    char buf[cmdSz + 4];
+    char buf[DEFAULT_SCP_MSG_SZ];
```
JacobBarthelmeh, line 1053: Please add a sanity check on cmdSz before using it in WSTRNCPY with the limited buf size now.

src/wolfscp.c
```diff
@@ -1110,16 +1118,15 @@ static int ScpCheckForRename(WOLFSSH* ssh, int cmdSz)
...
 /* helps with checking if the base path is a directory or file
  * returns WS_SUCCESS on success */
 static int ParseBasePathHelper(WOLFSSH* ssh, int cmdSz)
 {
-    int ret;
-
-    ret = ScpCheckForRename(ssh, cmdSz);
-
+    int ret = 0;
 #ifndef NO_FILESYSTEM
```
JacobBarthelmeh, line 1128: Why can the file in the cmd not be renamed in the case that there is NO_FILESYSTEM?

JacobBarthelmeh (commented):
The file permissions for keys/hansel-key-ecc.pem changed in this PR. Was that intentional?

src/wolfscp.c
danielinux: Added check for the limit on the buf size (against cmdSz + 4)

src/wolfscp.c
danielinux: Indeed `ScpCheckForRename` can be called outside the `#ifndef ...` block. It should be fixed now.

danielinux:
> The file permissions for keys/hansel-key-ecc.pem changed in this PR. Was that intentional?

Trying to use openssh's scp client with a world-readable code generates an error:
```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0660 for 'wolfssh/keys/hansel-key-ecc.pem' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "wolfssh/keys/hansel-key-ecc.pem": bad permissions
```

danielinux:
My bad on the permissions (I did `chmod 0700` instead of `chmod 0600` to test SCP against the FreeRTOS server before committing). It should be fixed now.

danielinux:
Tested against https://github.com/wolfSSL/wolfBoot-examples/pull/7

JacobBarthelmeh:
Thanks @danielinux , I noticed there is an issue with the build using the macro NO_WOLFSSH_CLIENT but part of that issue is also in master. I will open a separate PR with a fix.

JacobBarthelmeh (approved):

PR #332, opened 2021-04-22
User filesystem option

kojo1:
add WOLFSSH_USER_FILESYSTEM
fix OLFSSH_USER_IO conflict
add echoserver project for EWARM with myFilesystemmyFilesystem.c/h under ide/IAR-EWARM/Porjects, for now. May want to be under a generic directory
This is a dummy file system to be replaced with a user specific one.


JacobBarthelmeh (approved):

JacobBarthelmeh:
My only comment on this is the naming of the file looked for with WOLFSSH_USER_FILESYSTEM defined. Maybe setting myFilesystem.h to be user_filesystem.h instead to match like the user_settings.h include from wolfSSL with WOLFSSL_USER_SETTINGS.

PR #333, opened 2021-04-23
fix a leak

kojo1:


src/internal.c
```diff
@@ -3328,8 +3328,11 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 #ifdef WOLFSSH_SMALL_STACK
     if (sigKeyBlock_ptr)
         WFREE(sigKeyBlock_ptr, ssh->ctx->heap, DYNTYPE_PRIVKEY);
+    #ifndef WOLFSSH_NO_ECDSA
+        if (key_ptr)
```
JacobBarthelmeh, line 3332: can you please adjust the spacing here, the indent causes it to look like it is in the if(sigKeyBlock_ptr) block.

JacobBarthelmeh (approved):

PR #337, opened 2021-05-04
PurgePacket

ejohnstown:
Add function to purge a packet in case building one fails.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -7250,6 +7250,9 @@ int SendKexDhReply(WOLFSSH* ssh)
     if (ret == WS_SUCCESS)
         ret = SendNewKeys(ssh);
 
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
There is a lot of other locations where wolfSSH_SendPacket is called an one other location of SendNewKeys, should they also have a check to purge the packet after fail?

ejohnstown:
> There is a lot of other locations where wolfSSH_SendPacket is called an one other location of SendNewKeys, should they also have a check to purge the packet after fail?

SendPacket cleans up after itself. I was trying to catch cases where GrowBuffer() is called and an error causes an exit before the SendPacket.

JacobBarthelmeh (approved):

PR #342, opened 2021-06-16
Add TPM 2.0 support for wolfSSH as a Client

tomoveu:
Intention of this PR is to protect the Private RSA Key of a wolfSSH Client

To achieve this I have gone throw several iterations and the best solution I found is to pass just the TPM key handle to SSH internals. Everything else, like wolfTPM init, TPM key loading and similar, should happen in the Client's code.

Note: TPM Handle is like a pointer to the TPM, which loaded key to use. Once we have this inside the SSH internals, it is clean to call a wolfTPM wrapper. It took me a while, to give up on deeper wolfTPM<>wolfSSH integration. Just the handle is simple and clean.

Project integration - I tried to use the power of automake through _AC_CONFIG_SUBDIRS_. Does not seem to yield complete success as header includes are not found.

Please bear with me @ejohnstown and @JacobBarthelmeh , as this is my first wolfSSH PR and this integration effort was not as straightforward as it may seem.


wolfssh/ssh.h
```diff
@@ -247,6 +247,7 @@ enum WS_FormatTypes {
     WOLFSSH_FORMAT_PEM,
     WOLFSSH_FORMAT_RAW,
     WOLFSSH_FORMAT_SSH,
+    WOLFSSH_FORMAT_TPM,
```
danielinux, line 250: This WOLFSSH_FORMAT_TPM is not used in the client example. Did you plan to make an option to load the key from TPM via wolfSSH_ReadKey_* function ?

tomoveu: This is a leftover from one of the previous tries (code iterations). Thanks for catching it.

Indeed, this was its use. I think this was variant 2 of the integration.

examples/client/client.c
```diff
@@ -792,6 +797,66 @@ static int wolfSSH_AGENT_IO_Cb(WS_AgentIoCbAction action,
...
+
+    if (wolfTPM2_Init(dev, TPM2_IoCb, NULL) != TPM_RC_SUCCESS) {
+#ifdef DEBUG_WOLFSSH
+        printf("TPM 2.0 Device initialization failed\n");
+#endif
+        return WS_FAILURE;
+    }
+
+    /* TPM 2.0 keys live under a Primary Key, acquire such key */
+    if (getPrimaryStoragekey(dev, &storage, TPM_ALG_RSA)) {
```
dgarske, line 813: We should only be using real WolfTPM public API's. Not our test key wrappers. Even if you need to move some of the code from examples/tpm_test_keys.c` into wolfSSH.

dgarske (changes requested):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "wolfTPM"]
+	path = wolfTPM
+	url = https://github.com/wolfSSL/wolfTPM.git
```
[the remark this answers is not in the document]
tomoveu: I think wolfBoot directly compiles from wolfTPM and adds the wolfTPM sources files as object files to its makefile using options.mk. I will move the submodule to the lib folder.

examples/client/client.c
```diff
@@ -55,6 +55,11 @@
     #include <sys/select.h>
 #endif
 
+#ifdef WOLFSSH_TPM
+    #include "../../wolfTPM/wolftpm/tpm2_wrap.h"
```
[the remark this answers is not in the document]
tomoveu: OK. I will make the change.

examples/client/client.c
tomoveu: I was wondering this. Will do.

tomoveu: What to do with the IO Callbacks that are in examples/tpm_io.c(h) ?

I could just use NULL and have native Windows, Linux and TPM Simulator supports. And for baremetal to point to wolfTPM code.

tomoveu:
Changes:

* Fixed errors in code's logic
* wolfTPM is now used as a system library, same way wolfCrypt is used.
* removed wolfTPM as a git submodule (no longer needed)
* wolfTPM test helper functions are copied into examples/tpm and used in client's example, only if WOLFSSH_TPM is enabled
* Added set methods for tpm device and key

configure.ac
```diff
@@ -56,6 +56,7 @@ AC_CHECK_HEADERS([sys/select.h sys/time.h pty.h util.h termios.h])
 AC_CHECK_LIB([network],[socket])
 AC_CHECK_LIB([util],[forkpty])
 AC_CHECK_LIB([wolfssl],[wolfCrypt_Init],,[AC_MSG_ERROR([libwolfssl is required for ${PACKAGE}. It can be obtained from https://www.wolfssl.com/download.html/ .])])
+AC_CHECK_LIB([wolftpm],[wolfTPM2_Init],,[AC_MSG_ERROR([libwolftpm is required for ${PACKAGE}. It can be obtained from https://www.wolfssl.com/download.html/ .])])
```
dgarske, line 59: This is being required all the time. It needs to be in the `AC_ARG_ENABLE([tpm],` section.

examples/tpm/tpm_key_helpers.c
```diff
@@ -0,0 +1,361 @@
+/* helpers.c
```
dgarske, line 1: Please correct comment name. Please add `#ifdef WOLFSSH_TPM` around all this. Please remove and cleanup for functions that are not used.

examples/tpm/tpm_key_helpers.h
```diff
@@ -0,0 +1,71 @@
+/* tpm_test_keys.h
```
dgarske, line 1: Please correct comment name. Please add `#ifdef WOLFSSH_TPM` around all this. Please remove and cleanup for functions that are not used.

examples/tpm/tpm_test.h
```diff
@@ -0,0 +1,514 @@
...
+#ifndef _TPM_TEST_H_
+#define _TPM_TEST_H_
+
+#include <wolftpm/tpm2.h>
+
+#ifdef __cplusplus
+    extern "C" {
+#endif
+
+/* Test Configuration */
```
dgarske, line 32: Please add `#ifdef WOLFSSH_TPM` around all this. Please cleanup and remove items not used for this example. A huge portion of this is not applicable.

src/internal.c
```diff
@@ -8147,9 +8156,16 @@ static int BuildUserAuthRequestRsa(WOLFSSH* ssh,
             else {
                 int sigSz;
                 WLOG(WS_LOG_INFO, "Signing hash with RSA.");
+#ifdef WOLFSSH_TPM
+                wolfTPM2_SignHashScheme(ssh->ctx->tpmDev, ssh->ctx->tpmKey,
```
dgarske, line 8160: I don't see any return code checking on this call? What if there is a TPM failure? What if `ssh->ctx->tpmDev` or `ssh->ctx->tpmKey` are not set?

src/internal.c
```diff
@@ -7109,8 +7112,14 @@ int SendKexDhReply(WOLFSSH* ssh)
                 else {
                     WLOG(WS_LOG_INFO, "Signing hash with %s.",
                             IdToName(ssh->handshake->pubKeyId));
+#ifdef WOLFSSH_TPM
+                    wolfTPM2_SignHashScheme(ssh->ctx->tpmDev, ssh->ctx->tpmKey,
```
dgarske, line 7116: I don't see any return code checking on this call? What if there is a TPM failure? What if `ssh->ctx->tpmDev` or `ssh->ctx->tpmKey` are not set?

src/ssh.c
```diff
@@ -360,6 +360,26 @@ const char* wolfSSH_ErrorToName(int err)
...
+        ssh->ctx->tpmDev = dev;
+}
+
+
+void wolfSSH_SetTpmKey(WOLFSSH* ssh, WOLFTPM2_KEY* key)
+{
+    WLOG(WS_LOG_DEBUG, "Entering wolfSSH_SetTpmKey()");
+
+    if (ssh)
+        ssh->ctx->tpmKey = key;
```
dgarske, line 378: Might also check that `ssh->ctx` is not NULL.

src/ssh.c
```diff
@@ -360,6 +360,26 @@ const char* wolfSSH_ErrorToName(int err)
 }
 
 
+#ifdef WOLFSSH_TPM
+void wolfSSH_SetTpmDev(WOLFSSH* ssh, WOLFTPM2_DEV* dev)
+{
+    WLOG(WS_LOG_DEBUG, "Entering wolfSSH_SetTpmDev()");
+
+    if (ssh)
+        ssh->ctx->tpmDev = dev;
```
dgarske, line 369: Might also check that `ssh->ctx` is not NULL.

examples/client/client.c
dgarske: I don't mind if you use the `example/tpm_io.c` for the other IO callbacks.

examples/tpm/tpm_key_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+#ifndef _TPM_TEST_KEYS_H_
+#define _TPM_TEST_KEYS_H_
+
+#ifndef WOLFTPM2_NO_WRAPPER
+#include <wolftpm/tpm2.h>
+#include <wolftpm/tpm2_wrap.h>
+
+WOLFTPM_LOCAL int readKeyBlob(const char* filename, WOLFTPM2_KEYBLOB* key);
```
dgarske, line 29: The `WOLFTPM_LOCAL` is valid macro to use or have for the wolfSSH examples.

examples/tpm/tpm_key_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+#ifndef _TPM_TEST_KEYS_H_
+#define _TPM_TEST_KEYS_H_
+
+#ifndef WOLFTPM2_NO_WRAPPER
```
dgarske, line 25: All of the TPM support here requires `!WOLFTPM2_NO_WRAPPER`, so please make the build setting detection consistent for all TPM code. You can probably add a single well placed check:

```
#if defined(WOLFSSH_TPM) && defined(WOLFTPM2_NO_WRAPPER)
    /* wolfTPM support requires the wrapper to be enabled */
    #undef WOLFSSH_TPM
#endif
```

examples/client/client.c
dgarske: I would like to see all the code here calling wolfTPM native API's only. Bring in the `getPrimaryStoragekey` and `readKeyBlob` logic here. Eliminate the other files. I would like one copy/paste section of code as a template for this.

dgarske (changes requested):

configure.ac
tomoveu: This check is moved under the conditional check and now the build fails with ... when "--enable-tpm" is set.
```
Undefined symbols for architecture x86_64:
  "_wolfTPM2_SignHashScheme", referenced from:
      _SendKexDhReply in libwolfssh_la-internal.o
      _SendUserAuthRequest in libwolfssh_la-internal.o
```

Any idea why autoconf behaves like this?

tomoveu:
@dgarske all feedback taken in, I have not finished testing. 

Also, I would like the ssh key generation in wolfTPM to be merged first, but there I have to resolve Key attributes issue first.

examples/tpm/tpm_key_helpers.h
tomoveu: Do you think this is required for the code in src/internal.c and ssh.c?

It is no longer required for examples/tpm/... of course, because the file is removed, and only the used code is in the client demo. I was wondering about src/internal.c and ssh.c

tomoveu:
I have two issues to resolve:

1. TPM issue -  turns out I need to use TPM2_RSA_Decrypt with manual padding to use TPM Key for SSH
2. Autoconf - moving the wolfTPM lib check under a conditional check creates one unresolved function

examples/tpm/tpm_io.c
```diff
@@ -0,0 +1,198 @@
+/* tpm_io.c
```
dgarske, line 1: No, I do not want a copy of the example / reference tpm_io code in wolfSSH. That is duplicating a ton of code. This should reference the submodule example code...

examples/client/client.c
```diff
@@ -556,6 +560,133 @@ static inline void ato32(const byte* c, word32* u32)
...
+#if !defined(NO_FILESYSTEM) && !defined(NO_WRITE_TEMP_FILES)
+    XFILE  fp = NULL;
+    size_t fileSz = 0;
+    size_t bytes_read = 0;
+    byte pubAreaBuffer[sizeof(TPM2B_PUBLIC)];
+    int pubAreaSize;
+
+    XMEMSET(key, 0, sizeof(WOLFTPM2_KEYBLOB));
+
+    fp = XFOPEN(filename, "rb");
```
dgarske, line 618: These should be using the wolfSSH port macros WFOPEN, WFSEEK, WFTELL, FREWIND, etc...

examples/client/client.c
```diff
@@ -55,6 +55,10 @@
     #include <sys/select.h>
 #endif
 
+#ifdef WOLFSSH_TPM
+    #include <wolftpm/tpm2_wrap.h>
+    #include "../tpm/tpm_io.h"
```
dgarske, line 60: Use the submodule examples/tpm_io.h

examples/client/client.c
```diff
@@ -556,6 +560,133 @@ static inline void ato32(const byte* c, word32* u32)
...
+            /* Move storage key into persistent NV */
+            rc = wolfTPM2_NVStoreKey(pDev, TPM_RH_OWNER, pStorageKey,
+                TPM2_DEMO_STORAGE_KEY_HANDLE);
+        }
+    #endif
+    }
+    else {
+        /* specify auth password for storage key */
+        pStorageKey->handle.auth.size = sizeof(gStorageKeyAuth)-1;
+        XMEMCPY(pStorageKey->handle.auth.buffer, gStorageKeyAuth,
```
dgarske, line 592: Use the wolfSSH macros WMEMCPY, etc...

examples/client/client.c
```diff
@@ -556,6 +560,133 @@ static inline void ato32(const byte* c, word32* u32)
...
+{
+    int rc = 0;
+#if !defined(NO_FILESYSTEM) && !defined(NO_WRITE_TEMP_FILES)
+    XFILE  fp = NULL;
+    size_t fileSz = 0;
+    size_t bytes_read = 0;
+    byte pubAreaBuffer[sizeof(TPM2B_PUBLIC)];
+    int pubAreaSize;
+
+    XMEMSET(key, 0, sizeof(WOLFTPM2_KEYBLOB));
```
dgarske, line 616: WMEMSET

examples/client/client.c
```diff
@@ -1017,6 +1228,15 @@ THREAD_RETURN WOLFSSH_THREAD client_test(void* args)
     if (ssh == NULL)
         err_sys("Couldn't create wolfSSH session.");
 
+#ifdef WOLFSSH_TPM
+    /* Link the TPM 2.0 device to the SSH session
+     * This way, SSH client operations that can
+     * take advantage of the hardware security
+     * will have access to the TPM 2.0 device */
+    wolfSSH_SetTpmDev(ssh, &tpmDev);
```
dgarske, line 1236: Would a single API for both be easier?

dgarske (changes requested):
The .gitmodules is empty. Needs something like:

```
[submodule "lib/wolfTPM"]
	path = lib/wolfTPM
	url = https://github.com/wolfssl/wolfTPM
```

examples/tpm/tpm_io.c
tomoveu: OK. Will fix.

examples/client/client.c
tomoveu: Initially, I thought like this also. Then, realized there would be an instance, when tpmDev is linked and only the tpmKey used is changed or has to be changed. 

Image the use case

Customer's wolfSSH app
- Init TPM
- Set tpmDEv
- Load Key
- Set tpmKey
- Use Key
- ......
- Unload handle, Load another Key
- Set **new** tpmKey
===== and we can not be sure tpmKey would be exposed in this scenario, so setting just the tpmKey makes sense.

IF, I add GetTpmDev, then Set could be a single call for both device and key

SetTpm(tpmDev, TpmKey)
... user operations ...
GetTpmDev(&tpmDev);
SetTpm(tpmDev, NewTpmKey);

tomoveu:
@ejohnstown current status of this PR: I need the improvements/fixes to the Public Key authentication flow of wolfSSH to complete my work on TPM-backed Public Key authentication.

dgarske:
Closing to open new PR.

PR #347, opened 2021-07-01
M4 Update

ejohnstown:
1. Update the m4 files we updated in wolfSSL.
2. Update the configure.ac.


dgarske (approved):
Reviewed, compared to wolfSSL and tested. Looks good. Thanks John.

PR #355, opened 2021-08-02
fix for RSA public key auth

JacobBarthelmeh:


JacobBarthelmeh:
To test the fix start up the example echo server then use a client with the example key:

```
./examples/echoserver/echoserver
ssh -i ./keys/gretel-key-rsa.pem gretel@127.0.0.1
```

PR #359, opened 2021-08-03
use freshly compiled echoclient

gojimmypi:
although there may be an app called `ssh_client ` as noted in the current [README examples](https://github.com/wolfSSL/wolfssh#examples), the intention is likely to use the freshly compiled `echoclient` app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ to connect to. From the terminal run:
 
 From another terminal run:
 
-    $ ssh_client localhost -p 22222
+    $ ./examples/echoclient/echoclient
 
 The server will send a canned banner to the client:
 


JacobBarthelmeh:
Hi @gojimmypi , thank you for creating a pull request. It's true that the name of the application created by wolfSSH is examples/client/client. The use of name 'ssh_client' is an attempt to keep the command generic, any ssh client supporting the same version of SSH could be used to connect to the wolfSSH server.

gojimmypi:
Hi @JacobBarthelmeh - yes, that's true that the intention is for a generic 'ssh_client', but I think it makes more intuitive sense to use the one just compiled in the example, no? 

Further, using an actual ssh client such as `putty`, I was unsuccessful in connecting to the [example echoserver](https://github.com/wolfSSL/wolfssl/tree/master/examples/echoserver), as noted in https://github.com/wolfSSL/wolfssl/issues/4281

PR #380, opened 2021-12-06
Remove redundant free

ejohnstown:
Remove a redundant free of a hash object. The hash itself it still being updated and will be freed later.


danielinux (changes requested):
Please call wc_HashFree in the session cleanup

danielinux (approved):

PR #383, opened 2022-01-27
Fixes when compiling with an old version of GCC

ejohnstown:
1. Add include config.h to the example and test tool sources. They use some of the types from wolfCrypt's types.h and aren't finding them available. Also moved config.h to the top level source directory.
2. In the version of the system I was building for, functions pread() and pwrite() weren't readily available. Added a DECL check to configure to find if they are in unistd.h, and if not use the home brewed wPread() and wPwrite().
3. Cleaned up a bunch of compiler warnings for things like possible use of uninitialized variables and shadowing global names.

### DIFF
--- a/src/port.c
+++ b/src/port.c
@@ -37,6 +37,13 @@
     #include <stdio.h>
 #endif
 
+/*
+Flags:
+  WOLFSSH_LOCAL_PREAD_PWRITE
+    Defined to use local implementations of pread() and pwrite(). Switched
+    on automatically if pread() or pwrite() aren't found by configure.
+*/
+
 
 #if !defined(NO_FILESYSTEM) && !defined(WOLFSSH_USER_FILESYSTEM)
 int wfopen(WFILE** f, const char* filename, const char* mode)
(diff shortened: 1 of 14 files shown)


dgarske (approved):

src/port.c
```diff
@@ -82,6 +82,15 @@ int wfopen(WFILE** f, const char* filename, const char* mode)
 #endif
 }
 
+
+/* If either pread() or pwrite() are missing, use the local versions. */
+#if ((defined(HAVE_DECL_PREAD) && (HAVE_DECL_PREAD == 0)) || \
```
elms, line 87: Is there a case that `HAVE_DECL_PREAD` or `HAVE_DECL_PWRITE` aren't defined? If so does it make sense in that case to also default to the local version?

JacobBarthelmeh (approved):

src/port.c
ejohnstown: Fixed that. Thanks.

elms (approved):

dgarske (approved):

PR #389, opened 2022-02-25
Missing return value

kojo1:
https://wolfssl.zendesk.com/agent/tickets/13769

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -6762,6 +6762,8 @@ int wolfSSH_SFTP_SendWritePacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
                             ssh->error == WS_WANT_WRITE) {
                         return WS_FATAL_ERROR;
                     }
+                    ssh->error = WS_SFTP_BAD_HEADER;
+                    ret = WS_FATAL_ERROR;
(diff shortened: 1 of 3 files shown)


haydenroche5:
Hi @kojo1. The Windows failure should be resolved once this is merged https://github.com/wolfSSL/wolfssh/pull/390.

src/wolfsftp.c
```diff
@@ -6762,6 +6762,7 @@ int wolfSSH_SFTP_SendWritePacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
                             ssh->error == WS_WANT_WRITE) {
                         return WS_FATAL_ERROR;
                     }
+                    ret = WS_FATAL_ERROR;
```
kojo1, line 6765: Which value would be good for ssh->error?
Do you want to create a new value?

ejohnstown: None of the other calls to `SFTP_GetHeader()` are setting `ssh->error` in this failure case. We should have an error for this. Maybe WS_SFTP_BAD_HEADER.

ejohnstown (approved):

JacobBarthelmeh (approved):

PR #399, opened 2022-03-25
Add tests for scp, even empty file transfer.

anhu:


anhu:
Self assigned because of failing tests.

anhu:
Jenkins is failing on my tests.  But I don't understand why.  I can't repro on my local machine.  Tests pass fine.

ejohnstown (changes requested):
I think you can assume touch is present if they are running the script in bash.

anhu:
> I think you can assume touch is present if they are running the script in bash.

Will be done by Monday.

dgarske (approved):

ejohnstown (approved):

Issue #402, opened 2022-04-06
internal.c error: variable may be used uninitialized in this function


gojimmypi:
I'm compiling the `wolfssh` using the Espressif ESP-IDF. The default compiler settings are such that possibly uninitialized variable warnings are treated as errors. It is probably a good idea not to just ignore the warnings.

I've been able to appease the compiler by simply assigning an initial value of zero. I could submit a PR that does this. Alternatively perhaps the logic needs to be more closely investigated?

```
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssl/wolfssl/wolfcrypt/settings.h:2348:14: warning: #warning "For timing resistance / side-channel attack prevention consider using harden options" [-Wcpp]
             #warning "For timing resistance / side-channel attack prevention consider using harden options"
              ^~~~~~~
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c: In function 'GenerateKeys':
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c:997:49: error: 'sK' may be used uninitialized in this function [-Werror=maybe-uninitialized]
                                   sK->macKey, sK->macKeySz,
                                               ~~^~~~~~~~~~
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c:992:45: error: 'cK' may be used uninitialized in this function [-Werror=maybe-uninitialized]
                               cK->macKey, cK->macKeySz,
                                           ~~^~~~~~~~~~
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c: In function 'SendChannelOpen':
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c:9314:13: error: 'channelTypeSz' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         idx += channelTypeSz;
         ~~~~^~~~~~~~~~~~~~~~
In file included from C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/wolfssh/ssh.h:39,
                 from C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/src/internal.c:32:
C:/Users/gojimmypi/Desktop/esp-idf/components/wolfssh/wolfssh/port.h:384:31: error: 'channelType' may be used uninitialized in this function [-Werror=maybe-uninitialized]
[...]

PR #403, opened 2022-04-06
initialize variables

gojimmypi:
see https://github.com/wolfSSL/wolfssh/issues/402


dgarske:
OK to test

ejohnstown (changes requested):
1. Can you squash this down to one commit?
2. When initializing pointers, use NULL instead of 0.

gojimmypi:
closing. this turned out to be a git train wreck

PR #408, opened 2022-05-05
SABER Level 1 Support in wolfSSH

anhu:
Tested with the following: 

./autogen.sh 
./configure --enable-liboqs
make all check 

Also tested for interop with OQS fork of libssh. See README.md for instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ wolfSSL. The following is the simplest configuration of wolfSSL to
 enable wolfSSH.
 
     $ cd wolfssl
-    $ ./configure [OPTIONS] --enable-ssh
+    $ ./configure [OPTIONS] --enable-wolfssh
     $ make check
     $ sudo make install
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ wolfSSL. The following is the simplest configuration of wolfSSL to
 enable wolfSSH.
 
     $ cd wolfssl
-    $ ./configure [OPTIONS] --enable-ssh
+    $ ./configure [OPTIONS] --enable-wolfssh
```
dgarske, line 16: This option was only added recently in https://github.com/wolfSSL/wolfssl/pull/5100 . For now please leave the `--enable-ssh`.

wolfssh/internal.h
```diff
@@ -467,6 +479,7 @@ typedef struct HandshakeInfo {
 #endif
 
     byte useEcc;
+    byte useSaber;
```
dgarske, line 482: Please wrap with `#ifndef WOLFSSH_NO_SABER_LEVEL1_SHA256`. Also these should be bit-fields. `byte useSaber:1;` and `byte useEcc:1;`. You'll also need all the code with `!ssh->handshake->useSaber)` wrapped in the macro. I don't want code size to grow if saber is not used.

wolfssh/internal.h
```diff
@@ -404,6 +415,7 @@ struct WOLFSSH_CTX {
     byte* privateKey;                 /* Owned by CTX */
     word32 privateKeySz;
     byte useEcc;                      /* Depends on the private key */
+    byte useSaber;                    /* Depends on the private key */
```
dgarske, line 418: Please wrap with `#ifndef WOLFSSH_NO_SABER_LEVEL1_SHA256`. Also these should be bit-fields. `byte useSaber:1;` and `byte useEcc:1;`. You'll also need all the code with `!ssh->handshake->useSaber)` wrapped in the macro. I don't want code size to grow if saber is not used.

src/internal.c
```diff
@@ -7977,15 +8086,43 @@ int SendKexDhInit(WOLFSSH* ssh)
 #else
             ret = WS_INVALID_ALGO_ID;
 #endif /* !defined(WOLFSSH_NO_ECDH) */
+        } else if (ssh->handshake->useSaber) {
```
dgarske, line 8089: Please use:

```
}
#ifndef WOLFSSH_NO_SABER_LEVEL1_SHA256
else if (ssh->handshake->useSaber) {
...
}
#endif
```

dgarske (changes requested):
Please wrap all `useSaber` code in macros. We don't want that code present when Saber / Liboqs is not used.

wolfssh/internal.h
ejohnstown: Do use a bit-field with useSaber, but don't change useEcc yet. That's off topic for this PR. Doing it right when adding it is good. I don't want to change things piecemeal. I'd rather we convert the flags all at once. (The single bit flags aren't the worst memory wasters in wolfSSH, anyway.)

src/internal.c
```diff
@@ -7977,15 +8103,45 @@ int SendKexDhInit(WOLFSSH* ssh)
 #else
             ret = WS_INVALID_ALGO_ID;
 #endif /* !defined(WOLFSSH_NO_ECDH) */
+#ifndef WOLFSSH_NO_SABER_LEVEL1_SHA256
+        } else if (ssh->handshake->useSaber) {
+            OQS_KEM* kem = NULL;
+            ret = 0;
+
+            if (ret == 0) {
```
dgarske, line 8111: `ret` is always 0 here...

wolfssh/internal.h
```diff
@@ -404,6 +417,9 @@ struct WOLFSSH_CTX {
     byte* privateKey;                 /* Owned by CTX */
     word32 privateKeySz;
     byte useEcc;                      /* Depends on the private key */
+#ifndef WOLFSSH_NO_SABER_LEVEL1_SHA256
+    byte useSaber:1;                    /* Depends on the private key */
```
dgarske, line 421: NIT: Line up comments.

dgarske (changes requested):

src/internal.c
anhu: But it looks so nice and consistent.  Anyways, fixed.  Stay tuned.

dgarske (approved):
Looks good! Thanks. Over to John for final.

wolfssh/internal.h
anhu: OH...thats an interesting point.  If this gets expanded to different levels (ie 1, 3, 5) then I can switch this back to BYTE.

ejohnstown (approved):

PR #414, opened 2022-05-19
Espressif component install; updated setup_win.bat, new setup.sh 

gojimmypi:
there's likely no functional difference, as Windows can handle missing, trailing quotes in a path. Still, the trailing quotes were missing and addressed in `f38832c`.

edit: also added minor feature revision to [setup_win.bat](https://github.com/gojimmypi/wolfssh/blob/02347c4eaa6df641db7d85cc6aad639f9b6f05c1/ide/Espressif/ESP-IDF/setup_win.bat#L1) for Windows, and implemented new line-by-line match in [setup.sh](https://github.com/gojimmypi/wolfssh/blob/02347c4eaa6df641db7d85cc6aad639f9b6f05c1/ide/Espressif/ESP-IDF/setup.sh#L1) for Linux

### DIFF
--- a/ide/Espressif/ESP-IDF/setup.sh
+++ b/ide/Espressif/ESP-IDF/setup.sh
@@ -14,13 +14,13 @@
 #    options.h
 #      used with configure builds
 #
-#      This is an older file related an issue that’s been working forever.
+#      This is an older file related an issue that's been working forever.
 #      There should only be a wolfSSL copy right now. It is generated based on configure.
 #
 #    config.h
 #      This is generated per project. The configure script creates it.
 #      The one for wolfSSL is different than the one for wolfSSH
-#      There’s a #define that is added to the Makefile:
+#      There's a #define that is added to the Makefile:
 #
 #      #define HAVE_CONFIG
 #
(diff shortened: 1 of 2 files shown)


dgarske:
Jenkins retest this please

dgarske (changes requested):
Please add the execute bit on `setup.sh`. I had to do a `chmod +x setup.sh` before I could run it

FYI:  I ran the setup_win.bat from the IDF v4.4. Worked great.

gojimmypi:
@dgarske good catch on the executable flag. I developed on WSL which curiously has the executable flags all set, but did not get them pushed to GitHub. Also some trivia: `git add` and `git commit` on Linux appears to _recognize_ the flag change, but `git push` does not. There were a few odd characters and a trailing space that needed cleanup anyhow; that helped actually push the file change. 

At some point, the wolfssl component install should probably also have the parameterized destination option. I've found that rather quite handy to have local components specific to a project as compared to the shared ESP-IDF components..

gojimmypi (commented):
updated with flags & cleanup

dgarske (approved):

PR #416, opened 2022-05-25
Detect if ECC needs RNG

gojimmypi:
The [ecc.c](https://github.com/wolfSSL/wolfssl/blob/c3e4c6b6bc77c4c3d96385563d6134c9e06821bd/wolfcrypt/src/ecc.c#L4389) needs to be supplied a random number for timing resistance features, otherwise we see `err = MISSING_RNG_E`. 

> The wolfSSH library already handles it if you build with `HAVE_WC_ECC_SET_RNG`

In my case, I was using the `WOLFSSL_ESPIDF` in [wolfssl/wolfcrypt/settings.h](https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/settings.h#L299) that defines `ECC_TIMING_RESISTANT` but the API is NOT in all wolfSSL versions. For example older FIPS. Thus, this patch.

credit @dgarske for pointing out this fix.

### DIFF
--- a/wolfssh/settings.h
+++ b/wolfssh/settings.h
@@ -72,6 +72,15 @@ extern "C" {
     #error only SCP server side supported
 #endif
 
+/* Detect if ECC needs RNG */
+#if !defined(HAVE_WC_ECC_SET_RNG) && \
+    defined(ECC_TIMING_RESISTANT) && (!defined(HAVE_FIPS) || \
+    (!defined(HAVE_FIPS_VERSION) || (HAVE_FIPS_VERSION != 2))) && \
+    !defined(HAVE_SELFTEST)
+    /* Enable use of wc_ecc_set_rng */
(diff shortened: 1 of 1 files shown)


wolfssh/settings.h
```diff
@@ -72,6 +72,15 @@ extern "C" {
     #error only SCP server side supported
 #endif
 
+/* Detect if ECC needs RNG */
+#if !defined(HAVE_WC_ECC_SET_RNG) && \
```
dgarske, line 76: For reference this normally comes from the configure.ac here https://github.com/wolfSSL/wolfssh/blob/master/configure.ac#L126, however when building without configure this causes an issue with all ECC signing.

dgarske (approved):

dgarske:
Closing and will open a new one from a fork.

PR #425, opened 2022-06-28
ESP-IDF Build Cleanup

ejohnstown:
1. Update the list of files to leave out of the build. They are from wolfCrypt and aren't needed. Also left out wolfSSH's version of misc.c as it is handled automatically.


dgarske (approved):
Only change are whitespace CRLF and taking aes_asm.S out of the exclude (apparently only cares about .c).

gojimmypi (approved):
confirmed to work properly; removes compiler warnings

dgarske (approved):

PR #434, opened 2022-08-02
handle larger ext sizes with SFTP connect

JacobBarthelmeh:
ZD14588


dgarske (approved):
Customer confirmed fixed.

PR #435, opened 2022-08-05
Alpha version of SSHD support with wolfSSH

JacobBarthelmeh:
This adds the alpha version of SSHD support to wolfSSH. Enabled with --enable-sshd when compiling with autotools and off by default. The core features are implemented and some additional features are still in progress. Some testing has been completed but additional testing for a production ready version of the code is in progress, so far Linux/QNX OS's have been tested, OSX has been built but password support not yet completed.

Once built with --enable-sshd an example of running the SSHD daemon would be the following:

```
sudo ./apps/wolfsshd/wolfsshd -h <host key> -f <sshd_config file> -p 22222
```

To see a help menu use `./apps/wolfsshd/wolfsshd -h`. The program can be run in the foreground using the option '-D'.


configure.ac
```diff
@@ -210,7 +221,9 @@ AC_ARG_ENABLE([distro],
 AS_IF([test "x$ENABLED_DISTRO" = "xyes"],
       [ENABLED_ALL=yes; enable_shared=yes; enable_static=yes])
 AS_IF([test "x$ENABLED_ALL" = "xyes"],
-      [ENABLED_KEYGEN=yes; ENABLED_SCP=yes; ENABLED_SFTP=yes; ENABLED_FWD=yes; ENABLED_SHELL=yes; ENABLED_AGENT=yes])
+      [ENABLED_KEYGEN=yes; ENABLED_SCP=yes; ENABLED_SFTP=yes; ENABLED_FWD=yes; ENABLED_SHELL=yes; ENABLED_AGENT=yes; ENABLED_SSHD=yes])
+AS_IF([test "x$ENABLED_SSHD" = "xyes"],
+      [ENABLED_SHELL=yes;])
```
ejohnstown, line 226: Stray semicolon.

apps/wolfsshd/configuration.h
```diff
@@ -0,0 +1,54 @@
+/* configuration.h
+ *
+ * Copyright (C) 2014-2021 wolfSSL Inc.
```
ejohnstown, line 3: This is all over: Copyright 2022.

apps/wolfsshd/configuration.c
```diff
@@ -0,0 +1,908 @@
...
+
+    /* No value, nothing to do */
+    if (!value || value[0] == '\0') {
+        return WS_BAD_ARGUMENT;
+    }
+
+    /* Ignore trailing whitespace */
+    ptr = value + WSTRLEN(value) - 1;
+    while (ptr != value) {
+        if (!isspace(*ptr)) {
```
ejohnstown, line 448: This is more a question. We use wrappers for malloc and free, and bunch of the string functions. Do we want to be mixing the macro versions and bare calls? Is this expected to be built for Linux only, or for other OSes?

apps/wolfsshd/auth.c
```diff
@@ -0,0 +1,908 @@
...
+            wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Login as root not permited");
+            return WOLFSSH_USERAUTH_FAILURE;
+        }
+    }
+
+    ret = DoCheckUser(usr, auth);
+    /* temporarily elevate permissions */
+    if (ret == WOLFSSH_USERAUTH_SUCCESS &&
+            wolfSSHD_AuthRaisePermissions(auth) != WS_SUCCESS) {
+        wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Failure to raise permissions for auth");
```
ejohnstown, line 593: Wrap string, and trailing whitespace.

apps/wolfsshd/wolfsshd.c
```diff
@@ -0,0 +1,1017 @@
...
+    closelog();
+}
+
+#endif
+
+static void ShowUsage(void)
+{
+    printf("wolfsshd %s\n", LIBWOLFSSH_VERSION_STRING);
+    printf(" -?             display this help and exit\n");
+    printf(" -f <file name> Configuration file to use, default is /etc/ssh/sshd_config\n");
```
ejohnstown, line 146: Wrap string.

apps/wolfsshd/wolfsshd.c
```diff
@@ -0,0 +1,1017 @@
...
+                }
+
+                WFREE(data, NULL, 0);
+                wc_FreeDer(&der);
+            }
+        }
+    }
+
+    /* Load in host public key */
+//    {
```
ejohnstown, line 275: Comment style. This is a good case for `#if 0`

ejohnstown (changes requested):

apps/wolfsshd/configuration.c
JacobBarthelmeh: if using the isspace we should abstract it to a macro, will fix it

ejohnstown (approved):

PR #436, opened 2022-08-12
SFTP List

ejohnstown:
Fixes listing the contents of C:\ on Windows when running the wolfSFTP server. (ZD 14607, but they are using Nucleus)

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -2031,6 +2031,16 @@ static int wsUserAuth(byte authType,
 
 
 #ifdef WOLFSSH_SFTP
+/*
+ * Sets the WOLFSSH object's default SFTP path to the value provided by
+ * defaultSftpPath, or uses the current working directory from where the
+ * echoserver is run. The new default path is cleaned up with the real
+ * path function.
+ *
+ * @param ssh             WOLFSSH object to update
+ * @param defaultSftpPath command line provided default SFTP path
+ * @return                0 for success or error code
+ */
 static int SetDefaultSftpPath(WOLFSSH* ssh, const char* defaultSftpPath)
 {
     char path[WOLFSSH_MAX_FILENAME];
@@ -2039,26 +2049,31 @@ static int SetDefaultSftpPath(WOLFSSH* ssh, const char* defaultSftpPath)
 
     if (defaultSftpPath == NULL) {
     #ifdef USE_WINDOWS_API
-        if (GetCurrentDirectoryA(sizeof(path), path) == 0) {
-            ret = -1;
+        if (GetCurrentDirectoryA(sizeof(path)-1, path) == 0) {
+            ret = WS_INVALID_PATH_E;
         }
     #else
-        if (getcwd(path, sizeof(path)) == NULL) {
-            ret = -1;
+        if (getcwd(path, sizeof(path)-1) == NULL) {
+            ret = WS_INVALID_PATH_E;
         }
     #endif
     }
     else {
-        WSTRNCPY(path, defaultSftpPath, sizeof(path));
-        path[sizeof(path) - 1] = 0;
+        if (WSTRLEN(defaultSftpPath) >= sizeof(path)) {
+            ret = WS_INVALID_PATH_E;
+        }
+        else {
+            WSTRNCPY(path, defaultSftpPath, sizeof(path));
+        }
     }
 
     if (ret == 0) {
         path[sizeof(path) - 1] = 0;
-        wolfSSH_RealPath(NULL, path, realPath, sizeof(realPath));
-        if (wolfSSH_SFTP_SetDefaultPath(ssh, realPath) != WS_SUCCESS) {
-            ret = -1;
-        }
+        ret = wolfSSH_RealPath(NULL, path, realPath, sizeof(realPath));
+    }
+
+    if (ret == WS_SUCCESS) {
+        ret = wolfSSH_SFTP_SetDefaultPath(ssh, realPath);
     }
 
     return ret;
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -2030,6 +2030,42 @@ static int wsUserAuth(byte authType,
 }
 
 
+#ifdef WOLFSSH_SFTP
+static int SetDefaultSftpPath(WOLFSSH* ssh, const char* defaultSftpPath)
+{
+    char path[WOLFSSH_MAX_FILENAME];
+    char realPath[WOLFSSH_MAX_FILENAME];
+    int ret = 0;
+
+    if (defaultSftpPath == NULL) {
+    #ifdef USE_WINDOWS_API
+        if (GetCurrentDirectoryA(sizeof(path), path) == 0) {
+            ret = -1;
+        }
+    #else
+        if (getcwd(path, sizeof(path)) == NULL) {
+            ret = -1;
+        }
+    #endif
+    }
+    else {
+        WSTRNCPY(path, defaultSftpPath, sizeof(path));
+        path[sizeof(path) - 1] = 0;
+    }
+
+    if (ret == 0) {
+        path[sizeof(path) - 1] = 0;
+        wolfSSH_RealPath(NULL, path, realPath, sizeof(realPath));
+        if (wolfSSH_SFTP_SetDefaultPath(ssh, realPath) != WS_SUCCESS) {
+            ret = -1;
+        }
+    }
+
+    return ret;
+}
+#endif
+
+
 static void ShowUsage(void)
 {
     printf("echoserver %s\n", LIBWOLFSSH_VERSION_STRING);
@@ -2451,12 +2487,9 @@ THREAD_RETURN WOLFSSH_THREAD echoserver_test(void* args)
         }
 
     #ifdef WOLFSSH_SFTP
-        if (defaultSftpPath) {
-            if (wolfSSH_SFTP_SetDefaultPath(ssh, defaultSftpPath)
-                    != WS_SUCCESS) {
-                fprintf(stderr, "Couldn't store default sftp path.\n");
-                WEXIT(EXIT_FAILURE);
-            }
+        if (SetDefaultSftpPath(ssh, defaultSftpPath) != 0) {
+            fprintf(stderr, "Couldn't store default sftp path.\n");
+            WEXIT(EXIT_FAILURE);
         }
     #endif
 
--- a/src/port.c
+++ b/src/port.c
@@ -166,6 +166,16 @@ int wfopen(WFILE** f, const char* filename, const char* mode)
 
 #if defined(USE_WINDOWS_API) && (defined(WOLFSSH_SFTP) || defined(WOLFSSH_SCP))
 
+/*
+ * SFTP paths all start with a leading root "/". Most Windows file routines
+ * expect a drive letter when dealing with an absolute path. If the provided
+ * path, f, is of the form "/C:...", adjust the pointer f to point to the "C",
+ * and decrement the file path size, fSz, by one.
+ *
+ * @param f    pointer to a file name
+ * @param fSz  size of f in bytes
+ * @return     pointer to somewhere in f
+ */
 static const char* TrimFileName(const char* f, size_t* fSz)
 {
     if (f != NULL && fSz != NULL && *fSz >= 3 && f[0] == '/' && f[2] == ':') {
--- a/src/port.c
+++ b/src/port.c
@@ -166,6 +166,15 @@ int wfopen(WFILE** f, const char* filename, const char* mode)
 
 #if defined(USE_WINDOWS_API) && (defined(WOLFSSH_SFTP) || defined(WOLFSSH_SCP))
 
+static const char* TrimFileName(const char* f, size_t* fSz)
+{
+    if (f != NULL && fSz != NULL && *fSz >= 3 && f[0] == '/' && f[2] == ':') {
+        f++;
+        *fSz--;
+    }
+    return f;
+}
+
 void* WS_CreateFileA(const char* fileName, unsigned long desiredAccess,
         unsigned long shareMode, unsigned long creationDisposition,
         unsigned long flags, void* heap)
@@ -178,6 +187,8 @@ void* WS_CreateFileA(const char* fileName, unsigned long desiredAccess,
     errno_t error;
 
     fileNameSz = WSTRLEN(fileName);
+    fileName = TrimFileName(fileName, &fileNameSz);
+
     error = mbstowcs_s(&unicodeFileNameSz, NULL, 0, fileName, 0);
     if (error)
         return INVALID_HANDLE_VALUE;
@@ -211,6 +222,8 @@ void* WS_FindFirstFileA(const char* fileName,
     errno_t error;
 
     fileNameSz = WSTRLEN(fileName);
+    fileName = TrimFileName(fileName, &fileNameSz);
+
     error = mbstowcs_s(&unicodeFileNameSz, NULL, 0, fileName, 0);
     if (error)
         return INVALID_HANDLE_VALUE;
@@ -268,6 +281,8 @@ int WS_GetFileAttributesExA(const char* fileName, void* fileInfo, void* heap)
     errno_t error;
 
     fileNameSz = WSTRLEN(fileName);
+    fileName = TrimFileName(fileName, &fileNameSz);
+
     error = mbstowcs_s(&unicodeFileNameSz, NULL, 0, fileName, 0);
     if (error != 0)
         return 0;
@@ -301,6 +316,8 @@ int WS_RemoveDirectoryA(const char* dirName, void* heap)
     errno_t error;
 
     dirNameSz = WSTRLEN(dirName);
+    dirName = TrimFileName(dirName, &dirNameSz);
+
     error = mbstowcs_s(&unicodeDirNameSz, NULL, 0, dirName, 0);
     if (error != 0)
         return 0;
@@ -333,6 +350,8 @@ int WS_CreateDirectoryA(const char* dirName, void* heap)
     errno_t error;
 
     dirNameSz = WSTRLEN(dirName);
+    dirName = TrimFileName(dirName, &dirNameSz);
+
     error = mbstowcs_s(&unicodeDirNameSz, NULL, 0, dirName, 0);
     if (error != 0)
         return 0;
@@ -368,6 +387,7 @@ int WS_MoveFileA(const char* oldName, const char* newName, void* heap)
     errno_t error;
 
     oldNameSz = WSTRLEN(oldName);
+    oldName = TrimFileName(oldName, &oldNameSz);
 
     error = mbstowcs_s(&unicodeOldNameSz, NULL, 0, oldName, 0);
     if (error != 0)
@@ -382,6 +402,8 @@ int WS_MoveFileA(const char* oldName, const char* newName, void* heap)
         oldName, oldNameSz);
 
     newNameSz = WSTRLEN(newName);
+    newName = TrimFileName(newName, &newNameSz);
+
     error = mbstowcs_s(&unicodeNewNameSz, NULL, 0, newName, 0);
     if (error != 0)
         return 0;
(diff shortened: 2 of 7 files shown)


src/wolfsftp.c
```diff
@@ -1577,6 +1553,19 @@ int wolfSSH_SFTP_CreateStatus(WOLFSSH* ssh, word32 status, word32 reqId,
 }
 
 
+static int GetAndCleanPath(const char* dP, const byte* data, word32 sz,
```
haydenroche5, line 1556: Add comments explaining what this function does. dP is also not a descriptive name.

src/wolfsftp.c
```diff
@@ -1692,22 +1670,19 @@ int wolfSSH_SFTP_RecvMKDIR(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
     }
 
-    /* plus one to make sure is null terminated */
-    dir = (char*)WMALLOC(sz + 1, ssh->ctx->heap, DYNTYPE_BUFFER);
-    if (dir == NULL) {
-        return WS_MEMORY_E;
+    ret = GetAndCleanPath(ssh->sftpDefaultPath,
+            data + idx, sz, dir, sizeof(dir));
+    if (ret != WS_SUCCESS) {
+        return WS_BUFFER_E;
```
haydenroche5, line 1676: Why not just return ret, which has the actual error code that got thrown?

tests/api.c
```diff
@@ -1042,33 +1042,63 @@ struct RealPathTestCase realPathTestCases[] = {
...
-                   "         output: %s\n", i, err,
-                   defaultPath, tc->in, tc->exp, checkPath);
-        }
+    WSTRNCPY(testPath, tc->in, sizeof(testPath) - 1);
+    testPath[sizeof(testPath) - 1] = 0;
+    WMEMSET(checkPath, 0, sizeof checkPath);
+    err = wolfSSH_RealPath(path, testPath,
+            checkPath, sizeof checkPath);
+    if (err || WSTRCMP(tc->exp, checkPath) != 0) {
+        printf("RealPath failure (%d)\n"
```
haydenroche5, line 1080: fprintf to stderr preferred for diagnostics and error messages.

haydenroche5 (changes requested):

src/port.c
```diff
@@ -166,6 +166,15 @@ int wfopen(WFILE** f, const char* filename, const char* mode)
 
 #if defined(USE_WINDOWS_API) && (defined(WOLFSSH_SFTP) || defined(WOLFSSH_SCP))
 
+static const char* TrimFileName(const char* f, size_t* fSz)
```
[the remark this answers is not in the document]
ejohnstown: Added comment.

examples/echoserver/echoserver.c
```diff
@@ -2030,6 +2030,42 @@ static int wsUserAuth(byte authType,
...
+#ifdef WOLFSSH_SFTP
+static int SetDefaultSftpPath(WOLFSSH* ssh, const char* defaultSftpPath)
+{
+    char path[WOLFSSH_MAX_FILENAME];
+    char realPath[WOLFSSH_MAX_FILENAME];
+    int ret = 0;
+
+    if (defaultSftpPath == NULL) {
+    #ifdef USE_WINDOWS_API
+        if (GetCurrentDirectoryA(sizeof(path), path) == 0) {
```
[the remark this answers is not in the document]
ejohnstown: GetCurrentDirectory() is the WCHAR* version of the function. GetCurrentDirectoryA() is the char* version.

examples/echoserver/echoserver.c
```diff
@@ -2030,6 +2030,42 @@ static int wsUserAuth(byte authType,
...
+        }
+    #endif
+    }
+    else {
+        WSTRNCPY(path, defaultSftpPath, sizeof(path));
+        path[sizeof(path) - 1] = 0;
+    }
+
+    if (ret == 0) {
+        path[sizeof(path) - 1] = 0;
```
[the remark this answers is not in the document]
ejohnstown: I'm going to treat the other copy of that line as the duplicate.

examples/echoserver/echoserver.c
```diff
@@ -2030,6 +2030,42 @@ static int wsUserAuth(byte authType,
...
+            ret = -1;
+        }
+    #else
+        if (getcwd(path, sizeof(path)) == NULL) {
+            ret = -1;
+        }
+    #endif
+    }
+    else {
+        WSTRNCPY(path, defaultSftpPath, sizeof(path));
```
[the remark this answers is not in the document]
ejohnstown: Yeah, we should probably check that. At this point we know we aren't going to have a good path, that it will be truncated. We don't need to try to use it in a function to check it later. Also, the truncated path might point to some other file.

examples/echoserver/echoserver.c
```diff
@@ -2030,6 +2030,42 @@ static int wsUserAuth(byte authType,
...
+        path[sizeof(path) - 1] = 0;
+    }
+
+    if (ret == 0) {
+        path[sizeof(path) - 1] = 0;
+        wolfSSH_RealPath(NULL, path, realPath, sizeof(realPath));
```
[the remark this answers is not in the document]
ejohnstown: Done.

examples/echoserver/echoserver.c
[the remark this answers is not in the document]
ejohnstown: Duplicate removed.

src/ssh.c
```diff
@@ -2274,57 +2274,60 @@ int wolfSSH_ChannelGetEof(WOLFSSH_CHANNEL* channel)
...
-    }
-    else if (in[0] == '/') {
-        out[0] = '/';
-    }
-    else {
-        WSTRNCPY(out, currentPath, outSz);
+    out[0] = '/';
+    if (inSz == 0 || (!IS_DELIM(in[0]) && !IS_WINPATH(inSz, in))) {
+        if (defaultPath != NULL) {
+            WSTRNCPY(out, defaultPath, outSz);
```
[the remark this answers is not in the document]
ejohnstown: I thought this was the purpose of strncpy().

src/wolfsftp.c
ejohnstown: Renamed dP and commented.

src/wolfsftp.c
ejohnstown: Makes sense.

tests/api.c
ejohnstown: Changing it.

src/ssh.c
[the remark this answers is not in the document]
haydenroche5: Yeah, I was off on my out of bounds comment, but should we still check if strlen(defaultPath) < outSz? Basically, do we need to check if defaultPath fits in its entirety in out?

src/wolfsftp.c
```diff
@@ -1210,40 +1208,18 @@ static int wolfSSH_SFTP_RecvRealPath(WOLFSSH* ssh, int reqId, byte* data,
                 return WS_BAD_FILE_E;
             }
         #endif
-        }
-        WSTRNCAT(wd, "/", WOLFSSH_MAX_FILENAME);
-        WSTRNCAT(wd, r, WOLFSSH_MAX_FILENAME);
-        WMEMCPY(r, wd, WOLFSSH_MAX_FILENAME);
+        ret = wolfSSH_RealPath(wd, r, s, sizeof s);
```
haydenroche5, line 1211: So in the non-Windows case we give a buffer of 0s (wd is memset to 0s) as the default path?

src/wolfsftp.c
```diff
@@ -1577,6 +1553,31 @@ int wolfSSH_SFTP_CreateStatus(WOLFSSH* ssh, word32 status, word32 reqId,
...
+ * @param s           destination buffer for the Real Path
+ * @param sSz         size of s in bytes
+ * @return            0 for success or negative error code
+ */
+static int GetAndCleanPath(const char* defaultPath,
+        const byte* data, word32 sz, char* s, word32 sSz)
+{
+    char r[WOLFSSH_MAX_FILENAME];
+
+    if (sz > sizeof r) sz = sizeof r;
```
haydenroche5, line 1573: Shouldn't this be an error? Why truncate?

src/wolfsftp.c
```diff
@@ -1577,6 +1553,31 @@ int wolfSSH_SFTP_CreateStatus(WOLFSSH* ssh, word32 status, word32 reqId,
...
+    WMEMCPY(r, data, sz);
+    r[sz] = '\0';
```
haydenroche5, line 1575: If sz == sizeof(r) this will write out of bounds, I think.

src/wolfsftp.c
```diff
@@ -1900,13 +1881,10 @@ int wolfSSH_SFTP_RecvOpen(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
         return WS_BUFFER_E;
     }
 
-    /* plus one to make sure is null terminated */
-    dir = (char*)WMALLOC(sz + 1, ssh->ctx->heap, DYNTYPE_BUFFER);
-    if (dir == NULL) {
-        return WS_MEMORY_E;
+    if (GetAndCleanPath(ssh->sftpDefaultPath,
```
haydenroche5, line 1884: Would it be better to return the error from GetAndCleanPath rather than converting it to WS_BUFFER_E?

src/wolfsftp.c
```diff
@@ -2301,27 +2268,55 @@ int wolfSSH_SFTP_RecvOpenDir(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
-            realName, sizeof(realName), &isDir, ssh->ctx->heap);
-    if (findHandle == INVALID_HANDLE_VALUE || !isDir) {
+        /* get directory handle - see if directory exists */
+        findHandle = (HANDLE)WS_FindFirstFileA(name,
+                realName, sizeof(realName), &isDir, ssh->ctx->heap);
+        if (findHandle == INVALID_HANDLE_VALUE || !isDir) {
 
-        WLOG(WS_LOG_SFTP, "Error with opening directory");
-        WFREE(dirName, ssh->ctx->heap, DYNTYPE_BUFFER);
+            WLOG(WS_LOG_SFTP, "Error with opening directory");
```
haydenroche5, line 2307: Would be helpful to add the directory path to this log.

src/wolfsftp.c
```diff
@@ -2547,40 +2542,26 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
-            WSTRNCAT(buf, dirName, bufSz + 1);
-            tmpSz = WSTRLEN(buf);
-
-            /* add delimiter between path and file/dir name */
-            if (tmpSz + 1 < bufSz) {
-                buf[tmpSz] = WS_DELIM;
-                buf[tmpSz+1] = '\0';
-            }
-
+            WSTRNCAT(r, dirName, sizeof(r));
```
haydenroche5, line 2550: What if everything we're copying into r here doesn't fit? Won't r be garbled?

src/wolfsftp.c
```diff
@@ -2642,37 +2623,24 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
+        char  s[WOLFSSH_MAX_FILENAME];
 
-        /* add delimiter between path and file/dir name */
-        if ((tmpSz + 1 < bufSz) && (WMEMCMP(buf + tmpSz, (byte*)WS_DELIM, 1) != 0)) {
-            buf[tmpSz] = WS_DELIM;
-            buf[tmpSz+1] = '\0';
-        }
-        WSTRNCAT(buf, out->fName, bufSz + 1);
+        r[0] = '\0';
+        WSTRNCAT(r, dirName, sizeof(r));
```
haydenroche5, line 2630: Same question as above.

src/wolfsftp.c
```diff
@@ -2695,24 +2663,33 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
         word32 nameLen = (word32)WSTRLEN(dirName);
 
-        if (nameLen > MAX_PATH - 3) {
+        if (nameLen > MAX_PATH - 2) {
             WLOG(WS_LOG_SFTP, "Path name is too long.");
             return WS_FATAL_ERROR;
         }
         WSTRNCPY(name, dirName, MAX_PATH);
-        WSTRNCAT(name, "\\*", MAX_PATH);
+        WSTRNCAT(name, "/*", MAX_PATH);
```
haydenroche5, line 2692: It looks like this is intentional, but I just want to check that we _do_ in fact want to use / instead of \\, even though this is Windows.

src/wolfsftp.c
```diff
@@ -2817,43 +2794,25 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
-        buf = (char*)WMALLOC(bufSz + 1, out->heap, DYNTYPE_SFTP);
-        if (buf == NULL) {
-            WFREE(out->fName, out->heap, DYNTYPE_SFTP);
-            return WS_MEMORY_E;
-        }
-        buf[0] = '\0';
-        WSTRNCAT(buf, dirName, bufSz + 1);
-        tmpSz = (int)WSTRLEN(buf);
+        r[0] = '\0';
+        WSTRNCAT(r, dirName, sizeof(r));
```
haydenroche5, line 2801: Same question as earlier about this pattern.

src/wolfsftp.c
```diff
@@ -2899,38 +2858,24 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
+        char  s[WOLFSSH_MAX_FILENAME];
 
-        /* add delimiter between path and file/dir name */
-        if (tmpSz + 1 < bufSz) {
-            buf[tmpSz] = WS_DELIM;
-            buf[tmpSz+1] = '\0';
-        }
-        WSTRNCAT(buf, out->fName, bufSz + 1);
+        r[0] = '\0';
+        WSTRNCAT(r, dirName, sizeof(r));
```
haydenroche5, line 2865: Same question as earlier about this pattern.

src/wolfsftp.c
```diff
@@ -3887,41 +3826,26 @@ int wolfSSH_SFTP_RecvRename(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
-    old = (char*)WMALLOC(sz + 1, ssh->ctx->heap, DYNTYPE_BUFFER);
-    if (old == NULL) {
-        ret = WS_MEMORY_E;
-    }
-    if (ret == WS_SUCCESS) {
-        WMEMCPY(old, data + idx, sz); idx += sz;
-        old[sz] = '\0';
+    if (GetAndCleanPath(ssh->sftpDefaultPath,
+                data + idx, sz, old, sizeof(old)) < 0) {
+        ret = WS_FATAL_ERROR;
```
haydenroche5, line 3831: Does the stuff below need to be guarded by ret == WS_SUCCESS in the event we hit this error?

wolfssh/internal.h
```diff
@@ -717,6 +717,11 @@ struct WOLFSSH {
     struct WS_SFTP_SEND_WRITE_STATE* sendWriteState;
     struct WS_SFTP_GET_HANDLE_STATE* getHandleState;
     struct WS_SFTP_RENAME_STATE* renameState;
+#ifdef USE_WINDOWS_API
+    char driveList[26];
```
haydenroche5, line 721: 26 for each possible drive letter? Maybe map that to a constant/define just to be extra clear.

haydenroche5 (changes requested):

src/wolfsftp.c
ejohnstown: It would. Changing.

src/wolfsftp.c
ejohnstown: This should be an error.

src/wolfsftp.c
ejohnstown: Returning an error here instead.

ejohnstown: And updated incase we try using a provided path that is as big as r.

src/wolfsftp.c
ejohnstown: I see what I didn't do here. This worked with the echoserver because the echoserver is always setting the default path. Updating.

src/wolfsftp.c
ejohnstown: Did both instances of this error.

src/wolfsftp.c
ejohnstown: Yes it should be. And I see another pattern that needs addressing related to this.

src/ssh.c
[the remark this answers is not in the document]
ejohnstown: Added that check and a couple others. While strncpy should truncate appropriately and not overrun the buffer, a truncated path is known bad. Error sooner before giving out a path that is known bad and might collide with a file we don't want to actually access.

src/wolfsftp.c
ejohnstown: This is intentional. While back-slashes are the common directory separator in Windows, the file APIs accept paths with forward-slashes.

src/wolfsftp.c
ejohnstown: Returning an error if everything won't fit into r, and freeing the out filename.

src/wolfsftp.c
ejohnstown: Returning an error if everything won't fit into r, and freeing the out filename.

src/wolfsftp.c
ejohnstown: Returning an error if everything won't fit into r, and freeing the out filename.

src/wolfsftp.c
ejohnstown: Returning an error if everything won't fit into r, and freeing the out filename.

wolfssh/internal.h
ejohnstown: Got it.

src/wolfsftp.c
```diff
@@ -2545,16 +2564,26 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
...
-        r[0] = '\0';
         if (!special) { /* do not add dir name in special case */
-            WSTRNCAT(r, dirName, sizeof(r));
-            WSTRNCAT(r, "/", sizeof(r));
+            if (WSTRLEN(dirName) + out->fSz + 2 > (sizeof r)) {
+                WLOG(WS_LOG_SFTP, "Path length too large");
+                WFREE(out->fName, out->heap, DYNTYPE_SFTP);
+                return WS_FATAL_ERROR;
+            }
+            WSNPRINTF(r, sizeof(r), "%s/%s", dirName, out->fName);
```
haydenroche5, line 2573: Do we need to add a null terminator at the end of r?

src/wolfsftp.c
```diff
@@ -2626,13 +2655,16 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
         char  r[WOLFSSH_MAX_FILENAME];
         char  s[WOLFSSH_MAX_FILENAME];
 
-        r[0] = '\0';
-        WSTRNCAT(r, dirName, sizeof(r));
-        WSTRNCAT(r, "/", sizeof(r));
-        WSTRNCAT(r, out->fName, sizeof(r));
-        r[sizeof(r) - 1] = 0;
+        if ((WSTRLEN(dirName) + WSTRLEN(out->fName) + 2) > sizeof(r)) {
+            WLOG(WS_LOG_SFTP, "Path length too large");
```
haydenroche5, line 2659: Not sure how we intend to use WS_LOG_SFTP but am curious if this and other error logs should use WS_LOG_ERROR instead.

haydenroche5 (changes requested):

haydenroche5 (approved):

paulwratt:
@ejohnstown thanks for the effort required to get this completed

PR #441, opened 2022-08-19
add chroot jailing to sshd

JacobBarthelmeh:


JacobBarthelmeh:
retest this please

ejohnstown (changes requested):
The `chroot()` call fails. When it is called, the EUID is for user sshd, which doesn't have permission to call `chroot()`. I added in a `seteuid(0);` right before calling `chroot()` and it succeeded, but we got an "[SSHD] Invalid user name found" error returned by the server.

apps/wolfsshd/configuration.c
```diff
@@ -169,17 +196,101 @@ WOLFSSHD_CONFIG* wolfSSHD_ConfigNew(void* heap)
...
+
+    newConf = wolfSSHD_ConfigNew(conf->heap);
+    if (newConf != NULL) {
+        if (conf->banner) {
+            ret = CreateString(&newConf->banner, conf->banner,
+                                        (int)WSTRLEN(conf->banner),
+                                        newConf->heap);
+        }
+
+        if (ret == WS_SUCCESS && conf->chrootDir) {
```
ejohnstown, line 216: This code is duplicated down below at line 258.

ejohnstown (changes requested):

ejohnstown (approved):

PR #442, opened 2022-08-19
Add support for OpenSSH-style certificates.

haydenroche5:
This commit adds support for presenting an OpenSSH-style RSA SHA-1 cert as the
host public key. It also adds support for receiving the same type of cert for
user authentication.


JacobBarthelmeh:
https://github.com/wolfSSL/wolfssh/pull/561

PR #445, opened 2022-08-25
wolfSSHd Wildcard Config Includes

ejohnstown:
1. Fix the wildcard config file include.
2. Update the guard flags so macOS can also use wildcards.
3. Change the user priviledge separating setting to a bitfield.
4. Add test_configuration test to gitignore.
5. Process the included config files in strcmp order.

### DIFF
--- a/apps/wolfsshd/configuration.c
+++ b/apps/wolfsshd/configuration.c
@@ -437,185 +437,204 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
     const char *prefix = NULL;
     int prefixLen = 0;
     int found = 0;
+    int ret = WS_SUCCESS;
 
     /* No value, nothing to do */
     if (!value || value[0] == '\0') {
-        return WS_BAD_ARGUMENT;
+        ret = WS_BAD_ARGUMENT;
     }
 
-    /* Ignore trailing whitespace */
-    ptr = value + WSTRLEN(value) - 1;
-    while (ptr != value) {
-        if (WISSPACE(*ptr)) {
-            ptr--;
-        }
-        else {
-            break;
-        }
-    }
-
-    /* Find wildcards */
-    ptr2 = ptr;
-    while (ptr2 != value) {
-        if (*ptr2 == '*') {
-            /* Wildcard found */
-            found = 1;
-            if (ptr != ptr2) {
-                postfix = ptr2 + 1;
+    if (ret == WS_SUCCESS) {
+        /* Ignore trailing whitespace */
+        ptr = value + WSTRLEN(value) - 1;
+        while (ptr != value) {
+            if (WISSPACE(*ptr)) {
+                ptr--;
             }
-            break;
-        }
-        if (*ptr2 == '/') {
-            /* Found slash before wildcard directory-wildcards not supported */
-            break;
-        }
-        ptr2--;
-    }
-    ptr = ptr2;
-
-    /* Use wildcard */
-    if (found) {
-#if defined(__unix__) || defined(__unix) || \
-    (defined(__APPLE__) && defined(__MACH__))
-        int ret;
-        struct dirent *dir;
-        WDIR d;
-        char *path;
-        char *filepath = (char*)WMALLOC(PATH_MAX, conf->heap, DYNTYPE_PATH);
-
-        if (filepath == NULL) {
-            return WS_MEMORY_E;
+            else {
+                break;
+            }
         }
 
-        /* Back find the full path */
+        /* Find wildcards */
+        ptr2 = ptr;
         while (ptr2 != value) {
+            if (*ptr2 == '*') {
+                /* Wildcard found */
+                found = 1;
+                if (ptr != ptr2) {
+                    postfix = ptr2 + 1;
+                }
+                break;
+            }
             if (*ptr2 == '/') {
+                /* Found slash before wildcard directory-wildcards
+                 * not supported */
                 break;
             }
             ptr2--;
         }
+        ptr = ptr2;
 
-        if (ptr2 != value) {
-            path = (char*)WMALLOC(ptr2 - value + 1, conf->heap, DYNTYPE_PATH);
-            if (path == NULL) {
-                WFREE(filepath, conf->heap, DYNTYPE_PATH);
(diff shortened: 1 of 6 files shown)


haydenroche5 (changes requested):
Is it feasible to add tests to test_configuration.c for this?

haydenroche5:
Can we use the X/W functions (e.g. WREADDIR as opposed to readdir) throughout the Include stuff?

JacobBarthelmeh (approved):
The added test requires --enable-sftp on (./configure --enable-sshd without sftp gets compile errors). Otherwise looks great.

apps/wolfsshd/configuration.c
```diff
@@ -503,49 +505,80 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
             prefixLen = (int)(ptr - value);
         }
 
-        d = opendir(path);
-        if (d) {
-            while ((dir = readdir(d)) != NULL) {
-                if (dir->d_type == DT_DIR) {
-                    /* Skip sub-directories */
-                    continue;
+        if (!WOPENDIR(NULL, conf->heap, &d, path)) {
```
haydenroche5, line 508: Let's change the `memcpy`s above to `WMEMCPY`s. Also move the `} else {` onto a new line.

apps/wolfsshd/configuration.c
```diff
@@ -503,49 +505,80 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
...
-                        if ((int)WSTRLEN(dir->d_name) <= prefixLen) {
-                            continue;
-                        }
-                        if (WSTRNCMP(dir->d_name, prefix, prefixLen) != 0) {
-                            continue;
+            }
+            WREWINDDIR(&d);
+
+            if (fileCount > 0) {
+                fileNames = (char**)WMALLOC(fileCount * sizeof(char*), NULL, 0);
```
haydenroche5, line 522: Can use `conf->heap` for the heap argument to `WMALLOC` here and elsewhere.

apps/wolfsshd/configuration.c
```diff
@@ -474,10 +475,11 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
...
     if (found) {
-#ifdef __unix__
+#if defined(__unix__) || defined(__unix) || \
+    (defined(__APPLE__) && defined(__MACH__))
         int ret;
         struct dirent *dir;
-        DIR *d;
+        WDIR d;
         char *path;
         char *filepath = (char*)WMALLOC(PATH_MAX, NULL, 0);
```
haydenroche5, line 484: We need to check the return of this `WMALLOC`.

haydenroche5 (approved):

apps/wolfsshd/configuration.c
ejohnstown: A little later, the path variable is also malloced without a check, and isn't freed. Fixing.

apps/wolfsshd/configuration.c
ejohnstown: And added dynamic types.

apps/wolfsshd/configuration.c
ejohnstown: Check

apps/wolfsshd/configuration.c
```diff
@@ -490,69 +496,114 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
...
+                    WSNPRINTF(filepath, PATH_MAX, "%s/%s", path,
+                             fileNames[i]);
+                }
+                ret = wolfSSHD_ConfigLoad(conf, filepath);
+                if (ret != WS_SUCCESS) {
+                    WCLOSEDIR(&d);
+                    WFREE(fileNames, conf->heap, DYNTYPE_PATH);
+                    WFREE(filepath, conf->heap, DYNTYPE_PATH);
+                    WFREE(path, conf->heap, DYNTYPE_PATH);
+                    return ret;
```
haydenroche5, line 592: Would prefer this function have a single return at the end, which'll save us having to do all the freeing at multiple different return sites, but not critical.

apps/wolfsshd/test/test_configuration.c
```diff
@@ -153,12 +230,20 @@ int main(int argc, char** argv)
     (void)argc;
     (void)argv;
 
+    CleanupWildcardTest();
+    ret = SetupWildcardTest();
+    if (ret != 0) {
+        return 1;
```
haydenroche5, line 236: Prefer to return `WS_FATAL_ERROR` or another negative value for consistency.

haydenroche5 (changes requested):

apps/wolfsshd/configuration.c
```diff
@@ -474,85 +475,133 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
...
+                else {
+                    WMEMCPY(path, ".", 1);
+                    path[1] = '\0';
+                    prefix = value;
+                    prefixLen = (int)(ptr - value);
+                }
+            }
+        }
+
+        if (!WOPENDIR(NULL, conf->heap, &d, path)) {
```
haydenroche5, line 526: Needs `if (ret == WS_SUCCESS)`.

apps/wolfsshd/configuration.c
```diff
@@ -474,85 +475,133 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value)
...
+            WCLOSEDIR(&d);
         }
         else {
             /* Bad directory */
-            WFREE(filepath, NULL, 0);
-            return WS_BAD_ARGUMENT;
+            ret = WS_INVALID_PATH_E;
         }
-        WFREE(filepath, NULL, 0);
+        WFREE(path, conf->heap, DYNTYPE_PATH);
```
haydenroche5, line 603: Do  we need to check that path and filepath aren't NULL before doing these two `WFREE`s?

haydenroche5 (changes requested):

haydenroche5 (approved):

haydenroche5 (approved):

PR #454, opened 2022-09-07
Fix DoUserAuthRequestEccCert to only call mp_clear on vars that have been init'd

embhorn:
> never call mp_clear on something that hasn't been initialised.
The implementation in sp_int.c (as does integer.c) uses the 'used' field to determine the number of elements to zero out.
The tfm.c implementation just clears everything without reading values and so we may have got away with it in the past.

Fixes an issue from ZD14775

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -4726,31 +4726,54 @@ static int DoUserAuthRequestEcc(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
     }
 
     if (ret == WS_SUCCESS) {
-#ifdef WOLFSSH_SMALL_STACK
-    key_ptr = (ecc_key*)WMALLOC(sizeof(ecc_key), ssh->ctx->heap,
-            DYNTYPE_PUBKEY);
-    sig_r_ptr = (mp_int*)WMALLOC(sizeof(mp_int), ssh->ctx->heap, DYNTYPE_MPINT);
-    sig_s_ptr = (mp_int*)WMALLOC(sizeof(mp_int), ssh->ctx->heap, DYNTYPE_MPINT);
-    if (key_ptr == NULL || sig_r_ptr == NULL || sig_s_ptr == NULL)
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -4726,31 +4726,48 @@ static int DoUserAuthRequestEcc(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
...
+        sig_s_ptr = (mp_int*)WMALLOC(sizeof(mp_int), ssh->ctx->heap, DYNTYPE_MPINT);
+        if (sig_s_ptr == NULL) {
+            ret = WS_MEMORY_E;
+        }
+        else
+    #else
+        sig_s_ptr = &sig_s;
+    #endif
         if (mp_init(sig_s_ptr) != MP_OKAY) {
+            sig_s_ptr = NULL;
```
JacobBarthelmeh, line 4766: should sig_s_ptr be free'd in the WOLFSSH_SMALL_STACK case before set to NULL?

src/internal.c
```diff
@@ -4726,31 +4726,48 @@ static int DoUserAuthRequestEcc(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
...
+        sig_r_ptr = (mp_int*)WMALLOC(sizeof(mp_int), ssh->ctx->heap, DYNTYPE_MPINT);
+        if (sig_r_ptr == NULL) {
+            ret = WS_MEMORY_E;
+        }
+        else
+    #else
+        sig_r_ptr = &sig_r;
+    #endif
         if (mp_init(sig_r_ptr) != MP_OKAY) {
+            sig_r_ptr = NULL;
```
JacobBarthelmeh, line 4751: free'ing of sig_r_ptr in small stack case

src/internal.c
embhorn: Yep, that's right... Pushed update.

JacobBarthelmeh (approved):

PR #459, opened 2022-09-12
Fix wolfSSH client: freeing uninitialized data

embhorn:
Fixes ZD14816


src/internal.c
```diff
@@ -4025,15 +4025,17 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
             }
         }
 
-        if (sigKeyBlock_ptr->useRsa) {
+        if (ret == WS_SUCCESS) {
```
JacobBarthelmeh, line 4028: Will the RSA or ECC structure get free'd in the case that wc_SignatureVerify fails or any function call between the initialization and this free case?

JacobBarthelmeh (changes requested):

JacobBarthelmeh (approved):

PR #462, opened 2022-09-20
Fuzzing and Scan Build Fixes

ejohnstown:
1. Fuzzing found a bad read when processing an X.509 certificate chain. Use the correct bounds when checking value reads.
2. Fix a potential NULL dereference with the ssh-agent. Move the check inside an everything-ok check.
3. Fix a couple dead writes.

(ZD 14803)

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3301,58 +3301,50 @@ static int ParseAndVerifyCert(WOLFSSH* ssh, byte* in, word32 inSz,
     int ret;
     word32 l = 0, m = 0;
     word32 ocspCount = 0;
-    byte*  ocspBuf   = NULL;
-    word32 ocspBufSz = 0;
     word32 certCount = 0;
-    byte*  certPt    = NULL;
+    byte*  certChain = NULL;
     word32 certChainSz = 0;
 
     /* Skip the name */
     ret = GetSize(&l, in, inSz, &m);
     m += l;
 
-    /* Get the cert count */
-    ret = GetUint32(&certCount, in, inSz, &m);
     if (ret == WS_SUCCESS) {
-        WLOG(WS_LOG_INFO, "Peer sent certificate count of %d", certCount);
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -3301,31 +3301,29 @@ static int ParseAndVerifyCert(WOLFSSH* ssh, byte* in, word32 inSz,
...
+        /* Get the cert count */
+        ret = GetUint32(&certCount, in, inSz, &m);
     }
 
     if (ret == WS_SUCCESS) {
         word32 count;
 
-        certPt = in + m;
-        m = 0;
+        WLOG(WS_LOG_INFO, "Peer sent certificate count of %d", certCount);
```
haydenroche5, line 3320: certCount is unsigned int so %d should be %u.

src/internal.c
```diff
@@ -3301,31 +3301,29 @@ static int ParseAndVerifyCert(WOLFSSH* ssh, byte* in, word32 inSz,
...
+        certChain = in + m;
+
         for (count = certCount; count > 0; count--) {
             word32 certSz = 0;
 
-            ret = GetSize(&certSz, certPt, inSz, &m);
+            ret = GetSize(&certSz, in, inSz, &m);
             WLOG(WS_LOG_INFO, "Adding certificate size %d", certSz);
```
haydenroche5, line 3327: Same issue with %d here. Would also prefer this WLOG get moved after the break on ret != WS_SUCCESS block.

haydenroche5 (changes requested):

src/internal.c
ejohnstown: got it

src/internal.c
ejohnstown: got it

haydenroche5 (approved):

JacobBarthelmeh (approved):

PR #463, opened 2022-09-21
Improve FATFS: correct lifetime on ff_open/close

danielinux:


wolfssh/port.h
```diff
@@ -262,6 +266,15 @@ extern "C" {
         return (int)ret;
     }
     #define WFOPEN(f, fn, m) ff_fopen((f),(fn),(m))
+
+    static inline int ff_close(WFILE *f)
```
ejohnstown, line 270: This conflicts with the earlier prototype.

```
port.h:1085:9: error: conflicting types for 'ff_close'
 1085 |     int ff_close(int fd);
      |         ^~~~~~~~
```

ejohnstown (changes requested):

PR #464, opened 2022-09-22
update test include path, needed for Yocto builds

JacobBarthelmeh:


haydenroche5:
@JacobBarthelmeh Can you explain why this isn't working?

```
apps_wolfsshd_test_test_configuration_CPPFLAGS     = $(AM_CPPFLAGS) -DWOLFSSH_SSHD -DWOLFSSHD_UNIT_TEST -Iapps/wolfsshd/
``

Just curious.

JacobBarthelmeh:
I think Yocto does an out of directory build. Can reproduce the include error without Yocto by doing:

mkdir build
cd build
../configure --enable-sshd

When in another directory like that then apps/wolfsshd does not point to the location where configuration.h is.

haydenroche5:
If we do your solution, let's remove the unnecessary `-I` directive in the include.am. Or, you can keep the source code as is, and modify the include.am to use `$(srcdir)`.

haydenroche5 (approved):

PR #467, opened 2022-10-07
Multiple Server Keys

ejohnstown:
1. Allow user to add multiple server keys to the CTX up to a limit.
2. Clear the list of server's keys when releasing the CTX.
3. Update the API test case.
4. Server key algo list based on keys loaded.
5. Server uses key requested by client.
6. Change echoserver to load ECC and RSA key, -E option to select order.
7. Added a function to identify keys, and another to identify certs.
8. Added a function to add keys or certs to the list of keys and certs.
9. Fix return value for MAC Algo C2S match fail to the correct value.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -408,7 +408,7 @@ const char* GetErrorString(int err)
         case WS_CERT_KEY_SIZE_E:
             return "key size too small error";
 
-        case WS_CTX_KEY_COUNT:
+        case WS_CTX_KEY_COUNT_E:
             return "trying to add too many keys";
 
         default:
@@ -538,6 +538,8 @@ static const word32 cannedBannerSz = sizeof(cannedBanner) - 1;
 
 WOLFSSH_CTX* CtxInit(WOLFSSH_CTX* ctx, byte side, void* heap)
 {
+    word32 idx;
+
     WLOG(WS_LOG_DEBUG, "Entering CtxInit()");
 
     if (ctx == NULL)
@@ -571,6 +573,10 @@ WOLFSSH_CTX* CtxInit(WOLFSSH_CTX* ctx, byte side, void* heap)
     ctx->windowSz = DEFAULT_WINDOW_SZ;
     ctx->maxPacketSz = DEFAULT_MAX_PACKET_SZ;
 
+    for (idx = 0; idx < WOLFSSH_MAX_PVT_KEYS; idx++) {
+        ctx->privateKeyId[idx] = ID_NONE;
+    }
+
     return ctx;
 }
 
@@ -591,7 +597,7 @@ void CtxResourceFree(WOLFSSH_CTX* ctx)
             }
             #ifdef WOLFSSH_CERTS
             if (ctx->cert[i] != NULL) {
-                WFREE(ctx->cert[i], ctx->heap, DYNTYPE_PRIVKEY);
+                WFREE(ctx->cert[i], ctx->heap, DYNTYPE_CERT);
                 ctx->cert[i] = NULL;
                 ctx->certSz[i] = 0;
             }
@@ -806,42 +812,35 @@ union wolfSSH_key {
  * @param heap      heap to use for memory allocation
  * @return          keyId as int, WS_MEMORY_E, WS_UNIMPLEMENTED_E
  */
-static int IdentifyKey(const byte* in, word32 inSz, int isPrivate, void* heap)
+int IdentifyKey(const byte* in, word32 inSz, int isPrivate, void* heap)
 {
     union wolfSSH_key *key = NULL;
     int keyId = ID_UNKNOWN;
-    word32 scratch;
-    int ret = WS_SUCCESS;
+    word32 idx;
+    int ret;
     int dynType = isPrivate ? DYNTYPE_PRIVKEY : DYNTYPE_PUBKEY;
 
     key = (union wolfSSH_key*)WMALLOC(sizeof(union wolfSSH_key), heap, dynType);
-    if (key == NULL) {
-        ret = WS_MEMORY_E;
-    }
 
 #ifndef WOLFSSH_NO_RSA
-    if (ret == WS_SUCCESS) {
+    if (key != NULL) {
         /* Check RSA key */
         if (keyId == ID_UNKNOWN) {
-            scratch = 0;
+            idx = 0;
             ret = wc_InitRsaKey(&key->rsa, NULL);
 
             if (ret == 0) {
                 if (isPrivate) {
-                    ret = wc_RsaPrivateKeyDecode(in, &scratch,
-                            &key->rsa, inSz);
+                    ret = wc_RsaPrivateKeyDecode(in, &idx, &key->rsa, inSz);
                 }
                 else {
-                    ret = wc_RsaPublicKeyDecode(in, &scratch,
-                            &key->rsa, inSz);
+                    ret = wc_RsaPublicKeyDecode(in, &idx, &key->rsa, inSz);
                 }
 
-                /* If decode was successful, this was an RSA key. */
+                /* If decode was successful, this is an RSA key. */
                 if (ret == 0) {
                     keyId = ID_SSH_RSA;
                 }
-                /* in case we need to check for another key flavor */
-                ret = WS_SUCCESS;
             }
 
             wc_FreeRsaKey(&key->rsa);
@@ -849,22 +848,21 @@ static int IdentifyKey(const byte* in, word32 inSz, int isPrivate, void* heap)
     }
 #endif
 #ifndef WOLFSSH_NO_ECDSA
-    if (ret == WS_SUCCESS) {
+    if (key != NULL) {
+        /* Check ECDSA key */
         if (keyId == ID_UNKNOWN) {
-            scratch = 0;
+            idx = 0;
             ret = wc_ecc_init_ex(&key->ecc, heap, INVALID_DEVID);
 
             if (ret == 0) {
                 if (isPrivate) {
-                    ret = wc_EccPrivateKeyDecode(in, &scratch,
-                            &key->ecc, inSz);
+                    ret = wc_EccPrivateKeyDecode(in, &idx, &key->ecc, inSz);
                 }
                 else {
-                    ret = wc_EccPublicKeyDecode(in, &scratch,
-                            &key->ecc, inSz);
+                    ret = wc_EccPublicKeyDecode(in, &idx, &key->ecc, inSz);
                 }
 
-                /* If decode was successful, this was an RSA key. */
+                /* If decode was successful, this is an ECDSA key. */
                 if (ret == 0) {
                     switch (wc_ecc_get_curve_id(key->ecc.idx)) {
                         case ECC_SECP256R1:
@@ -878,8 +876,6 @@ static int IdentifyKey(const byte* in, word32 inSz, int isPrivate, void* heap)
                             break;
                     }
                 }
-                /* in case we need to check for another key flavor */
-                ret = WS_SUCCESS;
             }
 
             wc_ecc_free(&key->ecc);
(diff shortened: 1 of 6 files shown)


src/internal.c
```diff
@@ -408,6 +408,9 @@ const char* GetErrorString(int err)
         case WS_CERT_KEY_SIZE_E:
             return "key size too small error";
 
+        case WS_CTX_KEY_COUNT:
```
haydenroche5, line 411: Should we add an _E suffix to match the convention?

src/internal.c
```diff
@@ -576,17 +579,32 @@ void CtxResourceFree(WOLFSSH_CTX* ctx)
...
+        for (i = 0; i < ctx->privateKeyCount; i++) {
+            if (ctx->privateKey[i] != NULL) {
+                ForceZero(ctx->privateKey[i], ctx->privateKeySz[i]);
+                WFREE(ctx->privateKey[i], ctx->heap, DYNTYPE_PRIVKEY);
+                ctx->privateKey[i] = NULL;
+                ctx->privateKeySz[i] = 0;
+            }
+            #ifdef WOLFSSH_CERTS
+            if (ctx->cert[i] != NULL) {
+                WFREE(ctx->cert[i], ctx->heap, DYNTYPE_PRIVKEY);
```
haydenroche5, line 594: Should this use DYNTYPE_CERT?

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+ * @param inSz      size of key
+ * @param isPrivate indicates private or public key
+ * @param heap      heap to use for memory allocation
+ * @return          keyId as int, WS_MEMORY_E, WS_UNIMPLEMENTED_E
+ */
+static int IdentifyKey(const byte* in, word32 inSz, int isPrivate, void* heap)
+{
+    union wolfSSH_key *key = NULL;
+    int keyId = ID_UNKNOWN;
+    word32 scratch;
```
haydenroche5, line 813: Recommend renaming to idx.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+                if (isPrivate) {
+                    ret = wc_EccPrivateKeyDecode(in, &scratch,
+                            &key->ecc, inSz);
+                }
+                else {
+                    ret = wc_EccPublicKeyDecode(in, &scratch,
+                            &key->ecc, inSz);
+                }
+
+                /* If decode was successful, this was an RSA key. */
```
haydenroche5, line 867: RSA -> ECDSA.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+                if (ret == 0) {
+                    switch (wc_ecc_get_curve_id(key->ecc.idx)) {
```
haydenroche5, line 869: Might make sense to put this into a function. This snippet is also used in wolfSSH_ReadKey_buffer.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+
+    if (key != NULL) {
+        WFREE(key, heap, dynType);
+    }
+
+    if (keyId == ID_UNKNOWN) {
+        ret = WS_UNIMPLEMENTED_E;
+    }
+
+    return ret == WS_SUCCESS ? keyId : ret;
```
haydenroche5, line 898: I think it would be more readable if we add an else to the if at like 894 and set ret to keyId. Then just return ret without the ternary operator.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+
+    if (ret == 0) {
+        ret = IdentifyKey(key, keySz, 0, heap);
+        if (ret > 0) {
+            keyId = ret;
+            ret = WS_SUCCESS;
+        }
+    }
+
+    if (key != NULL) {
```
haydenroche5, line 946: WFREE should check for NULL, so this check can be omitted, I think.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+ * public key to IdentifyKey.
+ *
+ * @param in        certificate to identify
+ * @param inSz      size of certificate
+ * @param heap      heap to use for memory allocation
+ * @return          keyId as int, WS_MEMORY_E, WS_UNIMPLEMENTED_E
+ */
+static int IdentifyCert(const byte* in, word32 inSz, void* heap)
+{
+    struct DecodedCert cert;
```
haydenroche5, line 915: DecodedCert is a pretty large struct. May be a good idea to preemptively add the WOLFSSH_SMALL_STACK logic to this function.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+        }
+    }
+
+    if (ret == 0) {
+        ret = wc_GetPubKeyDerFromCert(&cert, key, &keySz);
+    }
+
+    if (ret == 0) {
+        ret = IdentifyKey(key, keySz, 0, heap);
+        if (ret > 0) {
```
haydenroche5, line 940: I think you can omit this if block entirely and just return ret on line 951.

src/internal.c
```diff
@@ -776,19 +794,231 @@ union wolfSSH_key {
...
+    if (destIdx >= WOLFSSH_MAX_PVT_KEYS) {
+        ret = WS_CTX_KEY_COUNT;
+    }
+    else {
+        if (ctx->privateKeyId[destIdx] == keyId) {
+            if (isKey) {
+                if (ctx->privateKey[destIdx] != NULL) {
+                    ForceZero(ctx->privateKey[destIdx],
+                            ctx->privateKeySz[destIdx]);
+                    WFREE(ctx->privateKey[destIdx], heap, dynamicType);
```
haydenroche5, line 976: What's dynamicType? Not seeing the declaration. Same problem elsewhere.

src/internal.c
```diff
@@ -7592,6 +7660,30 @@ static INLINE void CopyNameList(byte* buf, word32* idx,
...
 }
 
+
+static word32 BuildNameList(char* buf, word32 bufSz,
+        const byte* src, word32 srcSz)
+{
+    const char* name;
+    word32 nameSz, idx;
+
+    (void)bufSz;
```
haydenroche5, line 7670: Why take in bufSz if we aren't going to use it?

src/internal.c
```diff
@@ -576,17 +579,32 @@ void CtxResourceFree(WOLFSSH_CTX* ctx)
 {
     WLOG(WS_LOG_DEBUG, "Entering CtxResourceFree()");
 
-    if (ctx->privateKey) {
-        ForceZero(ctx->privateKey, ctx->privateKeySz);
-        WFREE(ctx->privateKey, ctx->heap, DYNTYPE_PRIVKEY);
+    if (ctx->privateKeyCount > 0) {
```
haydenroche5, line 582: Do we need corresponding init code in CtxInit? e.g. settings the private key IDs array to unknown, etc.

haydenroche5 (changes requested):

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: I forgot to add the length checking. I'm going to use it.

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

src/internal.c
ejohnstown: Check

ejohnstown:
retest this please

src/internal.c
haydenroche5: Looks like this one is still outstanding.

src/internal.c
```diff
@@ -7592,6 +7674,34 @@ static INLINE void CopyNameList(byte* buf, word32* idx,
...
+    const char* name;
+    word32 nameSz, idx;
+
+    idx = 0;
+    do {
+        name = IdToName(*src);
+        nameSz = (word32)WSTRLEN(name);
+
+        if (nameSz + 1 + idx > bufSz) {
+            break;
```
haydenroche5, line 7690: This will truncate the name list silently, I think. Should we change this function's return type to int and return a negative value in this case? Then wherever we call BuildNameList we check that the return isn't negative before proceeding.

src/internal.c
```diff
@@ -7592,6 +7674,34 @@ static INLINE void CopyNameList(byte* buf, word32* idx,
...
+
+        WMEMCPY(buf + idx, name, nameSz);
+        idx += nameSz;
+        buf[idx++] = ',';
+
+        src++;
+        srcSz--;
+    } while (srcSz > 0);
+
+    return idx - 1; /* Take off the size of last ','. */
```
haydenroche5, line 7701: If we return idx-1, wouldn't the caller expect the null terminator to come right after the last name? But instead, there'll be a comma there. I think we just need to not add the comma on the final iteration. Also perhaps add the null terminator explicitly to the buffer.

src/internal.c
```diff
@@ -7755,14 +7839,19 @@ int SendKexInit(WOLFSSH* ssh)
...
+    char* keyAlgoNames = NULL;
 
     WLOG(WS_LOG_DEBUG, "Entering SendKexInit()");
 
     if (ssh == NULL)
         ret = WS_BAD_ARGUMENT;
 
+    if (ssh->ctx->side == WOLFSSH_ENDPOINT_SERVER &&
+            ssh->ctx->privateKeyCount == 0) {
+        ret = WS_BAD_ARGUMENT;
```
haydenroche5, line 7852: Could be a good place for a debug message saying no private key loaded.

src/internal.c
```diff
@@ -7899,6 +7943,10 @@ int SendKexInit(WOLFSSH* ssh)
         }
     }
 
+    if (keyAlgoNames != NULL) {
```
haydenroche5, line 7946: WFREE should do the NULL check.

haydenroche5 (changes requested):

src/internal.c
ejohnstown: IdentifyKey() is the function. wolfSSH_ReadKey_buffer() calls IdentifyKey().

haydenroche5: I see now. Thanks.

src/internal.c
ejohnstown: check

src/internal.c
ejohnstown: check

src/internal.c
ejohnstown: check

haydenroche5 (approved):

JacobBarthelmeh (changes requested):

src/internal.c
```diff
@@ -776,19 +800,239 @@ union wolfSSH_key {
...
+    if (key == NULL) {
+        ret = WS_MEMORY_E;
+    }
+    else if (keyId == ID_UNKNOWN) {
+        ret = WS_UNIMPLEMENTED_E;
+    }
+    else {
+        ret = keyId;
+    }
+    WFREE(key, heap, dynType);
```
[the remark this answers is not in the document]
ejohnstown: WFREE is an alias for XFREE, and XFREE checks for NULL when building for FreeRTOS.

src/internal.c
```diff
@@ -7899,6 +7963,8 @@ int SendKexInit(WOLFSSH* ssh)
         }
     }
 
+    WFREE(keyAlgoNames, ssh->ctx->heap, DYNTYPE_STRING);
```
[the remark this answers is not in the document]
ejohnstown: WFREE is an alias for XFREE, and XFREE checks for NULL when building for FreeRTOS.

JacobBarthelmeh (approved):

PR #477, opened 2022-12-09
fix for x509 use, add cert use to sftp client

JacobBarthelmeh:
Fixes verification of the host when using X509 certificates:

The following commands should print out the IP check of the hosts certificate on the client side (with it using X509). Without this PR it is not using certificates with the connection, which can be seen by changing the clients -A CA certificate and the connection still being established.
```
sudo ./apps/wolfsshd/wolfsshd -f sshd_config -D -d

./examples/client/client -i ./keys/fred-key.der -J ./keys/fred-cert.der -A ./keys/ca-cert-ecc.der
```

Adds X509 support to SFTP example client.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -138,6 +138,7 @@ AS_IF([test "x$ax_enable_debug" = "xyes"],
        AM_CPPFLAGS="-DNDEBUG $AM_CPPFLAGS"])
 
 AX_PTHREAD([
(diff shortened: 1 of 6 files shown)


ejohnstown (changes requested):
Please add a more detailed description of the change and a reference to a support ticket or reason for the change. (I try to put ZD #XXXX or call out a Github issue.)

ejohnstown (approved):

PR #491, opened 2023-02-01
SFTP Stall Fix

ejohnstown:
SFTP Stall Fix

1. Fixed issue where a SFTP connection would stall while upload a large file.
2. Remove a little duplicate code.
3. Increased the channel receive buffer size to 128k.
4. Fixed the max packet size to 32k.
5. Clear the error register when calling stream read.

To test: Enable SFTP. Start the echoserver. Connect to the echoserver with WinSCP. Upload a 1GB file to the echoserver. If you back out the size change commit, it should stall while uploading. (The size change commit improves throughput.)

(ZD 14607)

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1007,48 +1007,7 @@ int wolfSSH_stream_peek(WOLFSSH* ssh, byte* buf, word32 bufSz)
 }
 
 
-/* moves the window for more room
- * returns WS_SUCCESS on success */
-static int wolfSSH_stream_adjust_window(WOLFSSH* ssh)
-{
-    word32  usedSz;
-    word32  bytesToAdd;
-    int     ret;
-    WOLFSSH_BUFFER* inputBuffer;
-
-    inputBuffer = &ssh->channelList->inputBuffer;
-    usedSz      = inputBuffer->length - inputBuffer->idx;
-    bytesToAdd  = inputBuffer->idx;
-
-    WLOG(WS_LOG_DEBUG, "Making more room: %u", usedSz);
-    WLOG(WS_LOG_DEBUG, "  Current index into buffer = %u", inputBuffer->idx);
-    WLOG(WS_LOG_DEBUG, "  Current max index for available data = %u",
-        inputBuffer->length);
-    WLOG(WS_LOG_DEBUG, "  Current total buffer size = %u",
-        inputBuffer->bufferSz);
-    if (usedSz) {
-        WLOG(WS_LOG_DEBUG, "  ...moving %d used bytes down", usedSz);
-        WMEMMOVE(inputBuffer->buffer, inputBuffer->buffer + bytesToAdd, usedSz);
-        inputBuffer->length = usedSz;
-        inputBuffer->idx = 0;
-    }
-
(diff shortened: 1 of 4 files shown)


wolfssh/internal.h
```diff
@@ -341,10 +341,10 @@ enum {
     #define DEFAULT_HIGHWATER_MARK ((1024 * 1024 * 1024) - (32 * 1024))
 #endif
 #ifndef DEFAULT_WINDOW_SZ
-    #define DEFAULT_WINDOW_SZ 16384
+    #define DEFAULT_WINDOW_SZ (128 * 1024)
```
JacobBarthelmeh, line 344: With this change the default listed in the README.md also needs updated

wolfssh/internal.h
```diff
@@ -341,10 +341,10 @@ enum {
...
 #endif
 #ifndef DEFAULT_MAX_PACKET_SZ
-    #define DEFAULT_MAX_PACKET_SZ (16 * 1024)
+    #define DEFAULT_MAX_PACKET_SZ 32768
```
JacobBarthelmeh, line 347: A comment on the choice of size should be added here, something to the effect of default max packet size defined in RFC XXX

src/ssh.c
```diff
@@ -2044,8 +2006,8 @@ static int _UpdateChannelWindow(WOLFSSH_CHANNEL* channel)
 
     inputBuffer = &channel->inputBuffer;
 
-    if (!channel->ssh->isKeying &&
-        (inputBuffer->length > inputBuffer->bufferSz / 2)) {
+    if ((inputBuffer->length > inputBuffer->bufferSz / 2) ||
+         (channel->windowSz == 0)) {
```
JacobBarthelmeh, line 2010: Changing the default highwater mark for more re-keying fails to send the full file now with the wolfsftp example.


```
$ ./configure CPPFLAGS="-DDEFAULT_HIGHWATER_MARK=16000" --enable-sftp && make

jak@ubuntu:~/Documents/wolfssh$ ls -la large-test.txt
-rw-rw-r-- 1 jak jak 2140640000 Dec 21 10:47 large-test.txt
jak@ubuntu:~/Documents/wolfssh$ ./examples/sftpclient/wolfsftp -u jill
Password: 
wolfSSH sftp> pwd
/home/jak/Documents/wolfssh
wolfSSH sftp> put large-test.txt test.txt
pushing large-test.txt to /home/jak/Documents/wolfssh/test.txt
Processed    32768	 bytes in 0 seconds
wolfSSH sftp> quit
jak@ubuntu:~/Documents/wolfssh$ ls -la test.txt 
-rw-r--r-- 1 jak jak 32768 Feb  1 08:17 test.txt
```

JacobBarthelmeh (changes requested):

src/ssh.c
ejohnstown: This is an issue with the client, not the server. The library wasn't clearing the error register, so a successful return was getting seen as a failure, terminating the connection very early.

JacobBarthelmeh (approved):

PR #500, opened 2023-02-28
Rename to ecdh-nistp256-kyber-512-sha256

anhu:
Used to be ecdh-sha2-nistp256-kyber-512-sha256 .

PR #510, opened 2023-04-21
ecc_p256-kyber_level1 interop with OQS OpenSSH

anhu:
The implementation now complies with the following draft: 

https://www.ietf.org/id/draft-kampanakis-curdle-ssh-pq-ke-01.html

We implement the method as defined by the following name in that draft: 

ecdh-nistp256-kyber-512r3-sha256-d00@openquantumsafe.org

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -113,7 +113,7 @@ Flags:
   WOLFSSH_NO_ECDSA_SHA2_NISTP521
     Set when ECC or SHA2-512 are disabled. Set to disable use of ECDSA server
     authentication with prime NISTP521.
-  WOLFSSH_NO_ECDH_SHA2_NISTP256_KYBER_LEVEL1_SHA256
+  WOLFSSH_NO_ECDH_NISTP256_KYBER_LEVEL1_SHA256
     Set when there is no liboqs integration. Set to disable use of ECDHE with
     prime NISTP256 hybridized with post-quantum Kyber Level1 KEM.
   WOLFSSH_NO_AES_CBC
@@ -1194,7 +1194,8 @@ int GenerateKey(byte hashId, byte keyId,
                 byte* key, word32 keySz,
                 const byte* k, word32 kSz,
                 const byte* h, word32 hSz,
-                const byte* sessionId, word32 sessionIdSz)
+                const byte* sessionId, word32 sessionIdSz,
+                byte doKeyPad)
 {
     word32 blocks, remainder;
     wc_HashAlg hash;
@@ -1220,7 +1221,10 @@ int GenerateKey(byte hashId, byte keyId,
         return WS_BAD_ARGUMENT;
     }
 
-    if (k[0] & 0x80) kPad = 1;
+    /* Data can be define as string and mpint. (see Section 5 of RFC4251).
+     * This padding is required in the case of an mpint, but not in the case of
+     * a string. */
+    if (doKeyPad && (k[0] & 0x80)) kPad = 1;
     c32toa(kSz + kPad, kSzFlat);
 
     blocks = keySz / digestSz;
@@ -1304,7 +1308,7 @@ int GenerateKey(byte hashId, byte keyId,
 }
 
 
-static int GenerateKeys(WOLFSSH* ssh, byte hashId)
+static int GenerateKeys(WOLFSSH* ssh, byte hashId, byte doKeyPad)
 {
     Keys* cK = NULL;
     Keys* sK = NULL;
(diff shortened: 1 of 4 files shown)


anhu:
[csosto-pk](https://github.com/csosto-pk) Please have a look.  This PR is a squashed version of https://github.com/wolfSSL/wolfssh/pull/500 .

src/internal.c
```diff
@@ -3711,6 +3714,10 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ecc_key key_s;
     #endif
 #endif
+#ifndef WOLFSSH_NO_ECDH_NISTP256_KYBER_LEVEL1_SHA256
+    byte ss_hashSz = 0;
```
JacobBarthelmeh, line 3718: Lets make variables follow camel case, right now we have a mix from not being strict enough. But our preference is camel case and the other snake case mix variable names will eventually be refactored.

src/internal.c
```diff
@@ -1220,7 +1221,7 @@ int GenerateKey(byte hashId, byte keyId,
         return WS_BAD_ARGUMENT;
     }
 
-    if (k[0] & 0x80) kPad = 1;
+    if (doKeyPad && (k[0] & 0x80)) kPad = 1;
```
JacobBarthelmeh, line 1224: Lets add a comment here on why we are not doing the padding. Something citing "Thus, K_PQ and K_CL are encoded as fixed-length byte arrays, not as integers. Byte arrays are defined in Section 5 of [[RFC4251](https://www.ietf.org/id/draft-kampanakis-curdle-ssh-pq-ke-01.html#RFC4251)]." Which is why I currently am assuming that we are skipping the padding here in special cases.

src/internal.c
```diff
@@ -4195,8 +4235,15 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         }
     }
 
-    if (ret == WS_SUCCESS)
-        ret = GenerateKeys(ssh, enmhashId);
+    if (ret == WS_SUCCESS) {
+        ret = GenerateKeys(ssh, enmhashId,
+#ifndef WOLFSSH_NO_ECDH_NISTP256_KYBER_LEVEL1_SHA256
+                           !ssh->handshake->useEccKyber
```
JacobBarthelmeh, line 4241: Can we make this more generic? Something along the lines of useKeyPadding. If we expand supported post quantum algorithms in the future than we are not adding more ifdef's and specific flag checks here.

JacobBarthelmeh: It would be helpful too with removing the magic number '1' in the else case of this macro guard.

src/internal.c
```diff
@@ -8704,6 +8753,8 @@ int SendKexDhReply(WOLFSSH* ssh)
                 privKey = (ecc_key*)WMALLOC(sizeof(ecc_key), heap,
                         DYNTYPE_PRIVKEY);
                 if (pubKey == NULL || privKey == NULL) {
+                    WFREE(pubKey, heap, DYNTYPE_PUBKEY);
```
JacobBarthelmeh, line 8756: was this leaking before? both appear to be free'd here https://github.com/wolfSSL/wolfssh/pull/510/files#diff-1c4e2f5adfa1ad30618e78ff459b2c0758ecf34278459ad0a8d58db4fec622eaR8802

src/internal.c
```diff
@@ -8782,52 +8835,78 @@ int SendKexDhReply(WOLFSSH* ssh)
...
+                        ssh->handshake->e) != OQS_SUCCESS) {
+                        ret = WS_PUBKEY_REJECTED_E;
+                    }
                 }
 
-                if (ret == 0)
+                if (ret == 0) {
+                    fSz -= kem->length_ciphertext;
+                    ssh->kSz -= kem->length_shared_secret;
+                } else {
```
JacobBarthelmeh, line 8866: coding style standards of:

```
if () {
}
else {
}
```

src/ssh.c
```diff
@@ -1718,7 +1718,7 @@ int wolfSSH_KDF(byte hashId, byte keyId,
 {
     WLOG(WS_LOG_DEBUG, "Entering wolfSSH_KDF()");
     return GenerateKey(hashId, keyId, key, keySz, k, kSz, h, hSz,
-                       sessionId, sessionIdSz);
+                       sessionId, sessionIdSz, 1);
```
JacobBarthelmeh, line 1721: Magic number of 1 here, lets comment it or use a named variable/define

src/internal.c
anhu: Wow!!  Great eye!

anhu:
Rebase and force push as suggested by colleague.

src/ssh.c
ejohnstown: We're not using TRUE/FALSE as constants. I'm OK just having a bare 1 for true instead of a name.

anhu:
Jenkins retest please.

anhu:
jenkins retest this please

anhu:
jenkins retest this please

JacobBarthelmeh (approved):

PR #517, opened 2023-05-15
RFC 8332

ejohnstown:
1. Add server auth support for RSA signing with SHA2-256 and SHA2-512.


src/internal.c
```diff
@@ -8982,28 +8970,24 @@ int SendKexDhReply(WOLFSSH* ssh)
...
+            ret = HashUpdate(hash, hashId, ssh->k, ssh->kSz);
         }
 
         /* Save the exchange hash value H, and session ID. */
         if (ret == 0) {
-            ret = wc_HashFinal(&ssh->handshake->hash,
-                               enmhashId, ssh->h);
-            wc_HashFree(&ssh->handshake->hash, enmhashId);
-            ssh->handshake->hashId = WC_HASH_TYPE_NONE;
+            ssh->hSz = wc_HashGetDigestSize(hashId);
```
JacobBarthelmeh, line 8985: This can return a BAD_FUNC_ARG value (-173). It should never happen since earlier 'update' calls were done with the same hashId but we should sanity check the return value before use later in the WMEMCPY call.

JacobBarthelmeh (changes requested):
This breaks support for password login when wolfSSHd is set to use x509.

To reproduce:

```
cd wolfssh && ./configure --enable-all && make
sudo ./apps/wolfsshd/wolfsshd -f ./sshd_config_sec -D -p 22222
```

From another terminal run: `ssh -p22222 <usr>@127.0.0.1`

The contents of the sshd_config_sec file are :

```
Port 22222
Protocol 2
LoginGraceTime 600
PermitRootLogin yes
PasswordAuthentication yes
PermitEmptyPasswords no
UsePrivilegeSeparation no
UseDNS no

TrustedUserCAKeys wolfssh/keys/ca-cert-ecc.pem     
HostKey wolfssh/keys/server-key.pem                
HostCertificate wolfssh/keys/server-cert.pem
```

The expected result is wolfSSHd handling both a x509 client connection and a password auth connection inter-oping with other SSH client.

tests/api.c
```diff
@@ -363,85 +359,85 @@ static void test_wolfSSH_CTX_UsePrivateKey_buffer(void)
...
-    AssertIntNE(lastKeySz, ctx->privateKeySz[0]);
+    AssertIntEQ(0, (lastKey == ctx->privateKey[0].key));
+    AssertIntNE(lastKeySz, ctx->privateKey[0].keySz);
 #endif
 
 #ifndef WOLFSSH_NO_SSH_RSA_SHA1
-    lastKey = ctx->privateKey[ctx->privateKeyCount];
-    lastKeySz = ctx->privateKeySz[ctx->privateKeyCount];
+    lastKey = ctx->privateKey[ctx->privateKeyCount].key;
+    lastKeySz = ctx->privateKeySz[ctx->privateKeyCount].keySz;
```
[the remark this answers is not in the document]
ejohnstown: Good catch.

src/internal.c
ejohnstown: Fixed.

src/internal.c
```diff
@@ -624,47 +629,46 @@ void CtxResourceFree(WOLFSSH_CTX* ctx)
...
-            der = (byte*)WMALLOC(ctx->privateKeySz[idx], ctx->heap,
-                DYNTYPE_PRIVKEY);
+            der = (byte*)WMALLOC(pvtKey->keySz, ctx->heap, DYNTYPE_PRIVKEY);
             if (der != NULL) {
                 int ret;
-                WMEMCPY(der, ctx->privateKey[idx], ctx->privateKeySz[idx]);
-                ret = SetHostPrivateKey(ctx, keyId, 1, der,
-                            ctx->privateKeySz[idx], DYNTYPE_PRIVKEY);
+                WMEMCPY(der, pvtKey->key, pvtKey->keySz);
+                ret = SetHostPrivateKey(ctx, pvtKey->publicKeyFmt, 1,
```
JacobBarthelmeh, line 670: This is still failing the password logon attempt to a wolfSSHd running with X509 support. The reason is that pvtKey->publicKeyFmt here has been updated to be an X509 type. This SetHostPrivateKey though should be using the initial publicKeyFmt at the start of the function which is a non X509 type. That way there is a ID_X509_* version and a ID_* version in the list.

JacobBarthelmeh (changes requested):

src/internal.c
ejohnstown: This isn't user authentication. This is a server authentication issue. It looks like wolfSSHd should be offering both x509v3-ecdsa-sha2-nistp256 with the certificate and ecdsa-sha2-nistp256 using the public key from the certificate.

JacobBarthelmeh (approved):

PR #512, opened 2023-04-24
wolfSSHd port to Windows

JacobBarthelmeh:
What's Left:
- [x] SCP on Windows
- [x] Public key / x509 authentication (currently only testing with password auth)
- [x] Proper management of security level / access of connection (right now the create process uses current process token with higher privileges rather than user token)
    - [x] SHELL
    - [x] SFTP
    - [x] SCP
- [ ] Better handling of waiting for data with shell, now it uses a lot of processing power while looping on checking if there is data to read/write
- [X] SFTP connection
- [X] Password auth shell connection
- [X] Initial build and wolfSSHD VS project
- [ ] Additional testing
    - [x] Memory leak testing
    - [ ] CI build tie in
    - [ ] static analysis testing

PR #519, opened 2023-05-19
break between SSH accept and SCP operation

JacobBarthelmeh:
setup chroot with scp operations too, breaks after SSH handshake and returns to application

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -435,6 +435,67 @@ static int SetupChroot(WOLFSSHD_CONFIG* usrConf)
     return ret;
 }
 
+#ifdef WOLFSSH_SCP
+static int SCP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
+    WPASSWD* pPasswd, WOLFSSHD_CONFIG* usrConf)
+{
+    int ret   = WS_SUCCESS;
+    int error = WS_SUCCESS;
+    int select_ret = 0;
+
+    /* temporarily elevate permissions to get users information */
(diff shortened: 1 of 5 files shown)


ejohnstown (changes requested):
I ran the wolfSSHd server. Then copied a file.

```
% scp -P22222 configure john@localhost:/home/john/configure
wolfSSHD
configure                                                                                         100%  633KB  17.7MB/s   00:00
```

It just hangs there.

JacobBarthelmeh:
I believe the hang with scp transfer happens in master too. When doing the Windows port of SCP I had noticed this and did a fix here (https://github.com/wolfSSL/wolfssh/pull/512/commits/d60c55b9a316a8e8fa24afaf52a441e562857cae#diff-ee22b49af3445d6389e6f9da16021ab72231506e131d6a4c357997a70fbbe061R1376)

That fix could probably be moved out of the Windows port PR though. Should it go here or in a new PR?

ejohnstown:
> I believe the hang with scp transfer happens in master too. When doing the Windows port of SCP I had noticed this and did a fix here ([d60c55b#diff-ee22b49af3445d6389e6f9da16021ab72231506e131d6a4c357997a70fbbe061R1376](https://github.com/wolfSSL/wolfssh/commit/d60c55b9a316a8e8fa24afaf52a441e562857cae#diff-ee22b49af3445d6389e6f9da16021ab72231506e131d6a4c357997a70fbbe061R1376))
> 
> That fix could probably be moved out of the Windows port PR though. Should it go here or in a new PR?

I'd put it in here, then rebase the Windows port PR when this is merged.

ejohnstown (approved):

PR #520, opened 2023-05-19
fix empty password support regression

JacobBarthelmeh:
Fix for sshd_config option `PermitEmptyPasswords yes`


JacobBarthelmeh:
Need to investigate case where shadow password is unavailable

ejohnstown (approved):

PR #531, opened 2023-06-28
fix for building client on windows with shell support

JacobBarthelmeh:
Reported directly to me. When building with WOLFSSH_SHELL on Windows the client project currently fails. This adds a macro guard around the include that is not found.


JacobBarthelmeh:
Can be tested with VS adding WOLFSSH_SHELL to ide/winvs/user_settings.h and removing the define for SINGLE_THREADED then building the VS solution.

dgarske (approved):

PR #535, opened 2023-07-07
Made dirlist part of WolfSSH struct.

asali23:
For concurrent access by the same user from different applications, the global variable dirlist can cause exception.


JacobBarthelmeh:
closing in favor of this PR https://github.com/wolfSSL/wolfssh/pull/539

PR #537, opened 2023-07-12
better handling of want read/write cases with SFTP + SSHD

JacobBarthelmeh:
fix for report given by direct email


JacobBarthelmeh:
Can be reproduced running wolfsshd on qnx (probably linux too) and using sftp to try and 'get' a specific file. Contact me for copy of the file if testing the exact reproducing case.

PR #541, opened 2023-07-21
Asynchronous user authentication.

falemagn:
This patch makes it possible to do asynchronous user authentication, by returning WOLFSSH_USERAUTH_WOULD_BLOCK from the authentication callback.


JacobBarthelmeh:
ok to test

falemagn:
@JacobBarthelmeh the functionality has only been tested server-side. To make use of it, one needs to return WOLFSSH_USERAUTH_WOULD_BLOCK and then at a later time trigger a new read from the socket. Since the auth request doesn't get discarded when returning WOLFSH_USERAUTH_WOULD_BLOCK, that invokes the auth callback with the same request again.

ejohnstown:
Superseded by PR #695.

PR #545, opened 2023-07-21
wolfSSH_CTX_SetSshProtoIdStr

falemagn:
Allow to modify the Protocol identifier on a per-context basis.


JacobBarthelmeh:
ok to test

src/ssh.c
```diff
@@ -1628,6 +1628,16 @@ int wolfSSH_CTX_SetBanner(WOLFSSH_CTX* ctx,
     return WS_SUCCESS;
 }
 
+WOLFSSH_API int wolfSSH_CTX_SetSshProtoIdStr(WOLFSSH_CTX* ctx,
```
JacobBarthelmeh, line 1631: please remove the leading WOLFSSH_API macro here on function implementation and add a null check for 'ctx' before de-reference.

JacobBarthelmeh (changes requested):

src/ssh.c
falemagn: done

JacobBarthelmeh (approved):

PR #546, opened 2023-07-21
Convert WS_SUCCESS to string.

falemagn:
Simply return a string representation of WS_SUCCESS too.


JacobBarthelmeh:
ok to test

JacobBarthelmeh (approved):

PR #550, opened 2023-07-21
Support partial packet reads during the SFTP server-side initialization.

falemagn:
Added new state and related buffer. Should eventually be replicated for the client


JacobBarthelmeh:
ok to test

JacobBarthelmeh (approved):

PR #551, opened 2023-07-21
Implemented WOLFSSH_FTP_FSETSTAT

falemagn:
Added support for the SSH_FXP_FSETSTAT packet type.


JacobBarthelmeh:
ok to test

JacobBarthelmeh (dismissed):

wolfssh/wolfsftp.h
```diff
@@ -254,6 +254,8 @@ WOLFSSH_LOCAL int wolfSSH_SFTP_RecvSetSTAT(WOLFSSH* ssh, int reqId, byte* data,
         word32 maxSz);
 WOLFSSH_LOCAL int wolfSSH_SFTP_RecvFSTAT(WOLFSSH* ssh, int reqId, byte* data,
         word32 maxSz);
+WOLFSSH_API int wolfSSH_SFTP_RecvFSetSTAT(WOLFSSH* ssh, int reqId, byte* data,
```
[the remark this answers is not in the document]
falemagn: I see you made it. Resolving the conversation.

src/wolfsftp.c
```diff
@@ -4946,6 +4994,81 @@ int wolfSSH_SFTP_RecvSetSTAT(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
+    byte  type  = WOLFSSH_FTP_OK;
+
+    if (ssh == NULL) {
+        return WS_BAD_ARGUMENT;
+    }
+
+    WLOG(WS_LOG_SFTP, "Receiving WOLFSSH_FTP_FSETSTAT");
+
+    /* get file handle */
+    ato32(data + idx, &sz); idx += UINT32_SZ;
```
SparkiDev, line 5027: Can maxSz be less than 4?

SparkiDev (changes requested):

src/wolfsftp.c
falemagn: I assume you're referring to the line below the one you've commented on.

`maxSz` is a parameter of the function, and it's basically the size of the buffer, as far as I can tell. It's the same kind of code that is used throughout the `wolfsftp.c` file. And indeed you code I suppose you're referring to  has been copied and pasted from other functions that need to get the file handle, which predate this PR, like [`wolfSSH_SFTP_RecvClose`](https://github.com/falemagn/wolfssh/blob/cc5a037f649c6d333172e6009c570b0a4b7b95ce/src/wolfsftp.c#L3621).

SparkiDev: The line I commented is pulling 4 bytes of data from the buffer.

If the maxSz is guaranteed to be more than 4 bytes, then you don't need a check to whether the buffer is big enough. Otherwise a check is needed to ensure there are 4 bytes to read.

falemagn: I understand now. If that were an issue, then it would be an issue also in [wolfSSH_SFTP_RecvClose](https://github.com/falemagn/wolfssh/blob/cc5a037f649c6d333172e6009c570b0a4b7b95ce/src/wolfsftp.c#L3620) and all other functions that do the same thing. I didn't delve into the code that deep, I just assumed that if it's correct elsewhere, then it's correct there.

JacobBarthelmeh:
Some additional sanity checks have now been added in to wolfSSH master.

JacobBarthelmeh (approved):

PR #552, opened 2023-07-21
Make wolfSSH_SFTPNAME_readdir overridable 

falemagn:
Make wolfSSH_SFTPNAME_readdir overridable when defining WOLFSSH_USER_FILESYSTEM and SFTP_Name_readdir.

SFTP_Name_readdir has to be defined as a macro, but it can be defined to expand to its own name if also a function with the same name is defined.

What matters is that it takes 3 arguments:

    1) the filesystem context as first argument;
    2) WDIR* as second argument;
    3) WS_SFTPNAME* as third argument

On successful execution, it returns WS_SUCCESS and the WS_SFTPNAME structure pointed by the third argument will be filled with the relevant info, otherwise a WS_* error code is returned.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -2997,7 +2997,6 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
     {
         char  r[WOLFSSH_MAX_FILENAME];
         char  s[WOLFSSH_MAX_FILENAME];
-
         if ((WSTRLEN(dirName) + WSTRLEN(out->fName) + 2) > sizeof(r)) {
             WLOG(WS_LOG_SFTP, "Path length too large");
             WFREE(out->fName, out->heap, DYNTYPE_SFTP);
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
ok to test

src/wolfsftp.c
```diff
@@ -2870,6 +2870,42 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
     return WS_SUCCESS;
 }
 
+#elif defined(WOLFSSH_USER_FILESYSTEM) && defined(SFTP_Name_readdir)
```
JacobBarthelmeh, line 2873: Would like to see the overridable macro match the naming scheme of all caps with WOLFSSH_ prepended. And near the top of the if-else list not restricted to WOLFSSH_USER_FILESYSTEM.

falemagn: Not sure what you mean with "near the top". Before my patch only `SFTP_GetAttributes` and `SFTP_GetAttributes_Handle` seemed to be overridable, but only if `WOLFSSH_USER_FILESYSTEM` is defined, I tried to stick to that, plus find a way to selectively override the specific function I needed.

Can you clarify?

JacobBarthelmeh: Around here for the top of the if else macro chain, https://github.com/wolfSSL/wolfssh/pull/552/files#diff-03800f8cf7db0a809cc8ba47f57f1e807aa193440295d3822eadbd4255c97118L2528. If someone is building then with say Windows, they can then still override the SFTP_NAME_READDIR

falemagn: @JacobBarthelmeh please, check if it's ok now.

ejohnstown:
Closing for updated PR #691.

PR #553, opened 2023-07-21
Correctly send out the CHANNEL_OPEN_FAIL message on failure.

falemagn:
Currently the implementation doesn't support having session channels coexist with other channels (including other sessions channels themselves), hence administratively forbid opening the channel in case such a situation arises.


JacobBarthelmeh:
ok to test

JacobBarthelmeh (approved):

PR #555, opened 2023-07-21
Added support for Ed25519 private keys and and X25519 curve.

falemagn:


JacobBarthelmeh:
ok to test

JacobBarthelmeh (changes requested):
Thanks @falemagn ! It looks like some macro guards are missing. The Jenkins tests fail when trying to link to wolfssl built with `./configure --enable-ssh --enable-keygen`. Also when trying with wolfssl built with `./configure --enable-ssh --enable-keygen --enable-ed25519 --enable-curve25519` I see some streaming API with ed25519 that needs guarded.

Running wolfSSH `make check` linked to wolfSSL configured with `./configure --enable-ssh --enable-keygen --enable-ed25519 --enable-curve25519 CPPFLAGS=-DWOLFSSL_ED25519_STREAMING_VERIFY` fails for me.

JacobBarthelmeh:
With the implementation requiring ed25519 streaming, ed25519 should not be an option if linked against a wolfSSL library that does not have the streaming feature enabled. Similar to how in wolfssh/internal.h there is a check for if wolfSSL has the NO_RSA macro, and if so disabled RSA code in wolfSSH. There would be a macro like WOLFSSH_NO_ED25519 that picks up on if the streaming macro from wolfssl/option.h is found.

```
  89 #ifdef NO_RSA
  90     #undef WOLFSSH_NO_RSA
  91     #define WOLFSSH_NO_RSA
  92 #endif
```

falemagn:
> With the implementation requiring ed25519 streaming, ed25519 should not be an option if linked against a wolfSSL library that does not have the streaming feature enabled. Similar to how in wolfssh/internal.h there is a check for if wolfSSL has the NO_RSA macro, and if so disabled RSA code in wolfSSH. There would be a macro like WOLFSSH_NO_ED25519 that picks up on if the streaming macro from wolfssl/option.h is found.
> 
> ```
>   89 #ifdef NO_RSA
>   90     #undef WOLFSSH_NO_RSA
>   91     #define WOLFSSH_NO_RSA
>   92 #endif
> ```

I've implemented that, but some tests are still not passing and since I cannot access them I am not sure what is going on. Could you give more feedback, please?

falemagn:
> I've implemented that, but some tests are still not passing and since I cannot access them I am not sure what is going on. Could you give more feedback, please?

@JacobBarthelmeh, after further analysis, it looks like the api test is failing for lack of Ed25519 support in the client side. Is that correct, as far as you can tell?

ejohnstown:
I'm taking this over. I going to squash your commits into one commit. I'll rebasing it to the current master branch. Since one of my coworkers has added x25519 key agreement already, so I'll remove that.

ejohnstown:
Will you mind if I force push an update to this branch?

falemagn:
> Will you mind if I force push an update to this branch?

Be my guest, so I need to do anything to allow you?

ejohnstown:
I'll try soon. I'm still adding the client-side support for Ed25519. I do have the rebase, squash, and cleanup of the server-side done enough. I pruned the Curve25519 KEX as it already exists in wolfSSH.

ejohnstown:
I squashed this PR into a single commit. I rebased it to the recent master branch fixing the conflicts. I added the missing parts for reading Ed25519 key files and for using in both the client and server. I removed the changes for Curve25519 key exchange as it is already in master added by @anhu.

PR #556, opened 2023-07-21
All the filesystem-related W* macros accept a filesystem context pointer as first parameter

falemagn:
For better portability.


JacobBarthelmeh:
ok to test

JacobBarthelmeh:
Have been reviewing this, and looked it over a couple times so far. My only concern is backwards compatibility if in some case the applications had overridden the W* file types with their own, then they would need to update their application when updating the wolfSSH version used.

falemagn:
> Have been reviewing this, and looked it over a couple times so far. My only concern is backwards compatibility if in some case the applications had overridden the W* file types with their own, then they would need to update their application when updating the wolfSSH version used.

Our own application needs a filesystem context in all cases, how do you suggest to move forward? This topic is also covered by issue #558.

JacobBarthelmeh (approved):

ejohnstown (changes requested):
Is there a chance you can pass the ssh->fs ctx into all the filesystem functions, and add an accessor to set the fs ctx?

ejohnstown:
> @ejohnstown I am not sure I understand. Which filesystem functions are we talking about?

I meant the function-style macros you added. For the fs parameter you passed in NULL, rather than the intended ssh->fs.

>The accessor, as far as I can see and understand, is already there: [`wolfSSH_SetFilesystemHandle`](https://github.com/wolfSSL/wolfssh/blob/f7958e10eebe14579a16c0c05960cc6c2231183e/src/ssh.c#L206).

Whoops. I see the function `wolfSSH_SetFilesystemHandle()`. That's my mistake.

ejohnstown:
I don't think it is going far enough. You should be passing in ssh->fs for the first parameter of all those macros. I'll edit it.

Also, there's a conflict with the wolfscp source. I'll fix that, too.

ejohnstown:
This should be rebased with the current master branch. And it breaks the wolfSSHd build. The changes aren't that bad.

ejohnstown (approved):

ejohnstown:
I am going to merge this despite the tests failed. I'm going to PR a fix right after merging this.

PR #563, opened 2023-08-04
fix file offest on Windows and change offset increment

JacobBarthelmeh:
Reported via direct email. On Windows the high offset was not getting set with SFTP leading to not stopping when writing files larger than max word32 size (over 4G). When fixing that I also changed the assign function to not depend on overflow (undefined behavior) for incrementing the high offset value.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -13217,12 +13217,20 @@ int wolfSSH_oct2dec(WOLFSSH* ssh, byte* oct, word32 octSz)
 }
 
 
+/* not using UINT_MAX here because word32 should always be 32 bits where an
+ * int type can change depending on platform */
+#define WS_WORD32_MAX_SZ 4294967295U
+
 /* addend1 += addend2 */
 void AddAssign64(word32* addend1, word32 addend2)
 {
-    word32 tmp = addend1[0];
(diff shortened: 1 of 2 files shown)


src/internal.c
```diff
@@ -13217,12 +13217,20 @@ int wolfSSH_oct2dec(WOLFSSH* ssh, byte* oct, word32 octSz)
 }
 
 
+/* not using UINT_MAX here because word32 should always be 32 bits where an
+ * int type can change depending on platform */
+#define WS_WORD32_MAX_SZ 4294967295U
```
dgarske, line 13222: Is it too ugly to use something like `#define WS_WORD32_MAX_SZ (word32)((1ULL << (sizeof(word32)*4)) - 1)`?

JacobBarthelmeh: Something like that would work. I get the idea, doing a bitshift to larger than the max size minus 1. Will make the change.

JacobBarthelmeh: Looking for WOLFSSL_BIT_SIZE (8) I see we have WOLFSSL_MAX_32BIT (0xffffffU), which is what I tried to find initially. Will try to use that instead of recreating it.

dgarske (approved):

PR #572, opened 2023-09-06
use WLOG for debug messages when checking peer IP address

JacobBarthelmeh:


ejohnstown (changes requested):
DefaultLoggingCB() outputs the newline at the end of the string. You don't need one in the error string.

ejohnstown (approved):

PR #575, opened 2023-09-11
fix WOLFSSH_API set for wolfSSH_SetTerminalResizeCtx

JacobBarthelmeh:


wolfssh/ssh.h
```diff
@@ -287,8 +287,8 @@ WOLFSSH_API int wolfSSH_ChangeTerminalSize(WOLFSSH* ssh, word32, word32,
     word32, word32);
 typedef int (*WS_CallbackTerminalSize)(WOLFSSH*, word32, word32, word32,
     word32, void*);
-void wolfSSH_SetTerminalResizeCb(WOLFSSH*, WS_CallbackTerminalSize);
-void wolfSSH_SetTerminalResizeCtx(WOLFSSH*, void*);
+WOLFSSH_API void wolfSSH_SetTerminalResizeCb(WOLFSSH*, WS_CallbackTerminalSize);
```
ejohnstown, line 290: Going forward, we must put the variable names in the prototypes in the header files. One of the checking tools errs on the variable names not matching between declaration and definition.

One of these days we need to update them all to match, but do it right for new items, please.

ejohnstown (changes requested):

ejohnstown (approved):

PR #579, opened 2023-09-13
fix DLL VS build configurations for wolfsshd and wolfsftp-client proj…

JacobBarthelmeh:
…ects

Fixes a DLL Release build of wolfsshd and wolfsftp-client projects being unable to find the wolfssl.lib file due to build configuration settings.


ejohnstown (changes requested):
s/DDL/DLL/ ?

ejohnstown (approved):

PR #581, opened 2023-09-15
Fix Terminal Window Size Update

ejohnstown:
Change sizeof a pointer to a string to WSTRLEN(). It was only copying part of the string "window-change" to the channel status message.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -13025,7 +13025,7 @@ int SendChannelTerminalResize(WOLFSSH* ssh, word32 columns, word32 rows,
     }
 
     if (ret == WS_SUCCESS) {
-        typeSz = (word32)sizeof(cType) - 1;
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh (dismissed):

src/internal.c
```diff
@@ -13025,7 +13025,7 @@ int SendChannelTerminalResize(WOLFSSH* ssh, word32 columns, word32 rows,
     }
 
     if (ret == WS_SUCCESS) {
-        typeSz = (word32)sizeof(cType) - 1;
+        typeSz = (word32)strlen(cType);
```
JacobBarthelmeh, line 13028: WSTRLEN?

ejohnstown: Good point.

JacobBarthelmeh (approved):

PR #583, opened 2023-09-18
Read PEM Keys

ejohnstown:
1. Update wolfSSH_ReadKey_buffer() to decode PEM keys.
2. Add detection of PEM files to wolfSSH_ReadKey_file().
3. Add parameter labels to the prototypes of the ReadKey functions.

Test Cases
----------

```
$ ./configure && make

$ ./examples/echoserver/echoserver -e
$ ./examples/client/client -u hansel -i keys/hansel-key-ecc.der -j keys/hansel-key-ecc.pub
$ ./examples/client/client -u hansel -i keys/hansel-key-ecc.pem -j keys/hansel-key-ecc.pub

$ ./examples/echoserver/echoserver
$ ./examples/client/client -u hansel -i keys/hansel-key-rsa.der -j keys/hansel-key-rsa.pub
$ ./examples/client/client -u hansel -i keys/hansel-key-rsa.pem -j keys/hansel-key-rsa.pub
```

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1479,7 +1479,8 @@ union wolfSSH_key {
 /* Reads a key from the buffer in to out. If the out buffer doesn't exist
    it is created. The type of key is stored in outType. It'll be a pointer
    to a constant string. Format indicates the format of the key, currently
-   either SSH format (a public key) or ASN.1 in DER format (a private key). */
+   either SSH format (a public key) or ASN.1 in DER or PEM format (a
+   private key). */
 int wolfSSH_ReadKey_buffer(const byte* in, word32 inSz, int format,
         byte** out, word32* outSz, const byte** outType, word32* outTypeSz,
         void* heap)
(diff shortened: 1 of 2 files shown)


src/ssh.c
```diff
@@ -1562,8 +1563,49 @@ int wolfSSH_ReadKey_buffer(const byte* in, word32 inSz, int format,
...
+        }
+
+        /* If it is PEM, convert to ASN1 then process. */
+        ret = wc_KeyPemToDer(in, inSz, newKey, newKeySz, NULL);
+        if (ret > 0) {
+            newKeySz = (word32)ret;
+            ret = WS_SUCCESS;
+        }
+        else {
+            ret = WS_ERROR;
```
dgarske, line 1590: Perhaps it would be helpful to use/capture or log the actual failure return code from wc_KeyPemToDer

src/ssh.c
```diff
@@ -1562,8 +1563,49 @@ int wolfSSH_ReadKey_buffer(const byte* in, word32 inSz, int format,
...
+        }
+
+        if (ret == WS_SUCCESS) {
+            ret = IdentifyKey(newKey, newKeySz, 1, heap);
+            if (ret > 0) {
+                *outSz = newKeySz;
+                *outType = (const byte*)IdToName(ret);
```
dgarske, line 1597: Should there be a NULL check for `*outType`?

dgarske (changes requested):

src/ssh.c
ejohnstown: Updating.

src/ssh.c
ejohnstown: `outType` is checked for NULL. It is a pointer to a string pointer. We're assigning `*outType` a string pointer.

dgarske (approved):

PR #596, opened 2023-09-29
fix speed of SFTP with SSHd when handling 'get' command

JacobBarthelmeh:
Before this change even when just sending data there would still be a wait on 'select' which killed throughput performance.


ejohnstown:
Jenkins retest this please

ejohnstown (approved):

PR #604, opened 2023-10-16
wolfSSH Client with OpenSSH-format Keys

ejohnstown:
1. Move the KeySignature struct around in internal.c so it can be used for a couple tasks.
2. Add decoder for the OpenSSH-format keys.
3. Add an identification function for the new key.
4. Update ReadKey to handle the new format.


src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+    byte keySigId;
+    word32 sigSz;
+    const char *name;
+    word32 nameSz;
+    union {
 #ifndef WOLFSSH_NO_RSA
-    RsaKey rsa;
+        struct {
+            RsaKey key;
+            byte e[256];
```
JacobBarthelmeh, line 863: this should be a max rsa macro instead of hard set

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+        } rsa;
 #endif
 #ifndef WOLFSSH_NO_ECDSA
-    ecc_key ecc;
+        struct {
+            ecc_key key;
+            word32 keyBlobSz;
+            const char *keyBlobName;
+            word32 keyBlobNameSz;
+            byte q[256];
```
JacobBarthelmeh, line 877: needs to be a macro instead of hard set max q size

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+
+static int GetOpenSshKeyRsa(RsaKey* key,
+        const byte* buf, word32 len, word32* idx)
+{
+    const byte* val = NULL;
+    word32 valSz = 0;
+    mp_int m;
+
+    GetMpint(&valSz, &val, buf, len, idx); /* n */
+    mp_read_unsigned_bin(&key->n, val, valSz);
```
JacobBarthelmeh, line 1054: checks on return value

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+
+
+static int GetOpenSshKeyEcc(ecc_key* key,
+        const byte* buf, word32 len, word32* idx)
+{
+    const byte *name = NULL, *priv = NULL, *pub = NULL;
+    word32 nameSz = 0, privSz = 0, pubSz = 0;
+    int ret;
+
+    GetStringRef(&nameSz, &name, buf, len, idx); /* curve name */
```
JacobBarthelmeh, line 1086: checks on return values

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+
+static int GetOpenSshKey(WS_KeySignature *key,
+        const byte* buf, word32 len, word32* idx)
+{
+    const char AuthMagic[] = "openssh-key-v1";
+    const byte* str = NULL;
+    word32 keyCount = 0, strSz = 0, i;
+    int ret = WS_SUCCESS;
+
+    if (strcmp(AuthMagic, (const char*)buf) != 0) {
```
JacobBarthelmeh, line 1106: strcmp should be WSTRCMP

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+                                    /* key comment */
+                    }
+                }
+                /* Padding: Add increasing digits to pad to the nearest
+                 * block size. Default block size is 8, but depends on
+                 * the encryption algo. The private key chunk's length,
+                 * and the length of the comment delimit the end of the
+                 * encrypted blob. No added padding required. */
+                if (ret == WS_SUCCESS) {
+                    if (strSz % 8 == 0) {
```
JacobBarthelmeh, line 1185: comment mentions default is 8 but if it can change based on algo then the value 8 should be a macro or variable

src/internal.c
```diff
@@ -844,110 +850,400 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
...
+                        if (strSz - subIdx > 0) {
+                            /* The padding starts at 1. */
+                            check2 = strSz - subIdx;
+                            for (check1 = 1;
+                                 check1 <= check2;
+                                 check1++, subIdx++) {
+                                if (check1 != str[subIdx]) {
```
JacobBarthelmeh, line 1192: can subIdx here be pass the end of str?

src/ssh.c
```diff
@@ -1479,137 +1476,231 @@ union wolfSSH_key {
...
+            WLOG(WS_LOG_DEBUG, "DER private key output size too small");
+            return WS_BUFFER_E;
+        }
+        newKey = *out;
+    }
+    *outSz = inSz;
+    WMEMCPY(newKey, in, inSz);
+
+    ret = IdentifyAsn1Key(in, inSz, 1, heap);
+    if (ret > 0) {
```
JacobBarthelmeh, line 1571: should we be free'ing newKey if ret is less than 0 here?

src/ssh.c
```diff
@@ -1479,137 +1476,231 @@ union wolfSSH_key {
...
+        void* heap)
+{
+    int ret = WS_SUCCESS;
+    byte* newKey = NULL;
+    word32 newKeySz = inSz; /* binary will be smaller than PEM */
+
+    WOLFSSH_UNUSED(heap);
+
+    if (*out == NULL) {
+        newKey = (byte*)WMALLOC(newKeySz, heap, DYNTYPE_PRIVKEY);
```
JacobBarthelmeh, line 1592: please check free of newKey in error cases

src/ssh.c
```diff
@@ -1479,137 +1476,231 @@ union wolfSSH_key {
...
+        if (*outSz < inSz) {
+            WLOG(WS_LOG_DEBUG, "PEM private key output size too small");
+            return WS_BUFFER_E;
         }
+        newKey = *out;
+    }
 
-        /* If it is PEM, convert to ASN1 then process. */
-        ret = wc_KeyPemToDer(in, inSz, newKey, newKeySz, NULL);
+    in += strlen(PrivBeginOpenSSH);
```
JacobBarthelmeh, line 1657: use WSTRLEN here for port layer

src/ssh.c
```diff
@@ -1479,137 +1476,231 @@ union wolfSSH_key {
...
+static int DoOpenSshKey(const byte* in, word32 inSz, byte** out,
+        word32* outSz, const byte** outType, word32* outTypeSz,
+        void* heap)
+{
+    int ret = WS_SUCCESS;
+    byte* newKey = NULL;
+    word32 newKeySz = inSz; /* binary will be smaller than PEM */
+
+    if (*out == NULL) {
+        newKey = (byte*)WMALLOC(newKeySz, heap, DYNTYPE_PRIVKEY);
```
JacobBarthelmeh, line 1643: check on free of newKey with error cases

JacobBarthelmeh (changes requested):

src/internal.c
ejohnstown: This should be actual size based on the key. This value isn't needed until after we've read the key into the RsaKey.

src/internal.c
ejohnstown: This should be actual size based on the key. This value isn't needed until after we've read the key into the ecc_key.

src/internal.c
ejohnstown: Check.

src/ssh.c
ejohnstown: Check.

src/ssh.c
ejohnstown: Check.

src/ssh.c
ejohnstown: Check.

src/ssh.c
ejohnstown: Check.

src/internal.c
ejohnstown: Check.

src/internal.c
ejohnstown: Check.

src/internal.c
ejohnstown: Check. Changed it to MIN_BLOCK_SZ.

src/internal.c
ejohnstown: I'm changing one of the earlier comparisons, I was subtracting subIdx and strSz and making sure it was greater than 0; I needed to make sure strSz is greater than subIdx. subIdx should not pass the end of str. check2 is based on the length of the str and subIdx. subIdx is smaller than strSz.

src/internal.c
ejohnstown: I double checked and these values weren't being used. Removed them.

src/internal.c
ejohnstown: I double checked and these values weren't being used. Removed them.

JacobBarthelmeh (changes requested):
Code changes look good. Thanks John! Only other thing is I see there is not a test yet. Hoping that we can add one of the keys to the wolfssh/keys directory and a basic test case.

JacobBarthelmeh (dismissed):

JacobBarthelmeh (approved):

PR #614, opened 2023-11-01
send error status on exit

JacobBarthelmeh:


ejohnstown (approved):

apps/wolfsshd/test/error_return.sh
```diff
@@ -0,0 +1,31 @@
...
+PUBLIC_KEY="./keys/hansel-key-ecc.pub"
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo "expecting host and port as arguments"
+    echo "./error_return.sh 127.0.0.1 22222"
+    exit -1
+fi
+
+echo "$TEST_CLIENT -c 'ls error' -u $USER -i $PRIVATE_KEY -j $PUBLIC_KEY -h \"$1\" -p \"$2\""
+$TEST_CLIENT -c 'ls error' -u $USER -i $PRIVATE_KEY -j $PUBLIC_KEY -h "$1" -p "$2"
```
ejohnstown, line 20: I would like to see the command run change to
```
bash -c "(exit 2)"
```
Which would start the bash shell which starts a sub shell that exits immediately with the specified result code. This works on the Mac, unlike `ls error`, which doesn't return 2 on the Mac (BSD).

ejohnstown (changes requested):

ejohnstown (approved):

PR #617, opened 2023-11-14
do not require locking file with SFTP on Windows

JacobBarthelmeh:
Fix for dirty read on Windows with SFTP.
ZD16961


ejohnstown (approved):
Good find.

PR #623, opened 2023-11-28
fix for grace period login and add test case

JacobBarthelmeh:
ZD17083

### DIFF
--- a/apps/wolfsshd/test/run_all_sshd_tests.sh
+++ b/apps/wolfsshd/test/run_all_sshd_tests.sh
@@ -63,6 +63,7 @@ fi
 # these tests require setting up an sshd
 if [ "$USING_LOCAL_HOST" == 1 ]; then
     run_test "sshd_forcedcmd_test.sh"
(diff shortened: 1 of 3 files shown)


ejohnstown (changes requested):
```
ubuntu@wolfssh:~/wolfssh/apps/wolfsshd/test$ ./create_authorized_test_file.sh
ubuntu@wolfssh:~/wolfssh/apps/wolfsshd/test$ sudo ./sshd_login_grace_test.sh 127.0.0.1 22222
rm: cannot remove 'log.txt': No such file or directory
SSHD running on PID 5829
 rejecting password type with root in favor of pub key
Password: cat: ./log.txt: No such file or directory
Grace period not hit
```

1. Need a `if -f ./log.txt` before removing it.
2. The script changes directories up three levels, but doesn't come back down before trying to cat log.txt. Use pushd and popd.
3. Check out the program `shellcheck` sometime. It makes good suggestions.

JacobBarthelmeh:
Retest this please Jenkins

PR #625, opened 2023-11-30
Known Hosts Update

ejohnstown:
The issue was visible when configuring for sshclient and scp. The client would perform an invalid write while processing the known hosts file. I don't know what scp had to do with it, but it was necessary to enable it. There was definitely a bug and it has been fixed.

1. Move setting a nul termination on the knownHosts data until after checking the size is reasonable.
2. A temporary keySz variable was getting used to get the length of the key type value, but it wasn't used to copy the value. Deleted it and used the other sz value.
3. Fix the leaking of the known hosts filename.


ejohnstown:
Please retest this.

ejohnstown:
Retest this please Jenkins

JacobBarthelmeh (approved):

PR #638, opened 2023-12-28
Remove Example Server

ejohnstown:
The example server was never used. All new behaviors were added to the echoserver only. Now with wolfSSHd, the server example can finally be removed.

1. Remove the example server from the build.
2. Remove the file notes.md.
3. Update the readme.


dgarske (approved):

douzzer (approved):
LGTM

PR #640, opened 2023-12-28
wolfSSHd Terminal

ejohnstown:
Some cleanups and improvements to the code in wolfSSHd that starts the user's shell. Uses the modes in the client's terminal settings provided in the PTY request. Also configures some of the inherited variables in the new shell.


douzzer (approved):
LGTM

PR #642, opened 2023-12-28
Gate GetOpenSshKeyRsa with WOLFSSH_NO_RSA

gojimmypi:
`GetOpenSshKeyRsa()` is not used with `#define WOLFSSH_NO_RSA` and thus the compiler complains it is unused when RSA is disabled.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -1048,6 +1048,7 @@ static INLINE int GetMpintToMp(mp_int* mp,
 }
 
 
+#ifndef WOLFSSH_NO_RSA
 /*
  * For the given RSA key, calculate p^-1 and q^-1. wolfCrypt's RSA
  * code expects them, but the OpenSSH format key doesn't store them.
@@ -1073,7 +1074,6 @@ static INLINE int CalcRsaInverses(RsaKey* key)
     return ret;
 }
 
-#ifndef WOLFSSH_NO_RSA
 /*
  * Utility for GetOpenSshKey() to read in RSA keys.
  */
@@ -1107,6 +1107,8 @@ static int GetOpenSshKeyRsa(RsaKey* key,
 }
 #endif
 
+
+#if !defined(WOLFSSH_NO_ECDSA) && !defined(WOLFSSH_NO_ECC)
 /*
  * Utility for GetOpenSshKey() to read in ECDSA keys.
  */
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -1073,7 +1073,7 @@ static INLINE int CalcRsaInverses(RsaKey* key)
     return ret;
 }
 
-
+#ifndef WOLFSSH_NO_RSA
```
ejohnstown, line 1076: Move this up to include the CalcRsaInverses function.

ejohnstown (changes requested):
Add the similar guard around GetOpenSshKeyEcdsa() as well.

src/internal.c
gojimmypi: done

gojimmypi:
> Add the similar guard around GetOpenSshKeyEcdsa() as well.

This one was a little more interesting. I assume you meant `GetOpenSshKeyEcc()`. Some related things seemed to need manual attention, such as turning on a curve.

Here's a segment of the [user_settings.h](https://github.com/gojimmypi/wolfssh/blob/component-manager/ide/Espressif/ESP-IDF/examples/wolfssl_echoserver/components/wolfssl/include/user_settings.h) I used for testing. Please confirm sanity:

```
#define MY_USE_ECC 0
#define MY_USE_RSA 1

/* We can use either or both ECC and RSA, but must use at least one. */
#if MY_USE_ECC || MY_USE_RSA
    #if MY_USE_ECC
        /* ---- ECDSA / ECC ---- */
        #define HAVE_ECC
        #define HAVE_CURVE25519
        #define HAVE_ED25519

        /*
        #define HAVE_ECC384
        #define CURVE25519_SMALL
        */
    #else
        #define WOLFSSH_NO_ECC
        /* WOLFSSH_NO_ECDSA is typically defined automatically,
         * here for clarity: */
        #define WOLFSSH_NO_ECDSA
    #endif

    #if MY_USE_RSA
        /* ---- RSA ----- */
        /* #define RSA_LOW_MEM */

        /* DH disabled by default, needed if ECDSA/ECC also turned off */
        #define HAVE_DH
    #else
        #define WOLFSSH_NO_RSA
    #endif
#else
    #error "Either RSA or ECC must be enabled"
#endif
```

ejohnstown (approved):

Issue #633, opened 2023-12-18
Espressif fixes: MISC_WARNING not expanded, vars not initialized, not used


gojimmypi:
A variety of minor issues crop up when compiling in the Espressif ESP-IDF environment.

- The [#warning MISC_WARNING](https://github.com/wolfSSL/wolfssh/blob/7f52e23a5ac41a9652c1625bec24f758484f08c1/src/misc.c#L59) is manifested exactly like that, and not with the `MISC_WARNING` message:

```
ples/wolfssh_server/build/bootloader
[989/1010] Building C object esp-idf/wolfssh/CMakeFiles/__idf_wolfssh.dir/src/misc.c.obj/mnt/c/workspace/wolfssl_scripts-gojimmypi/espressif/wolfssh-component-publish/examples/wolfssh_server/components/wolfssh/src/misc.c:59:10: warning: #warning MISC_WARNING [-Wcpp]
   59 |         #warning MISC_WARNING
      |          ^~~~~~~
[995/1010] Building C object esp-idf/wolfssh/CMakeFiles/__idf_wolfssh.dir/src/ssh.c.objFAILED: esp-idf/wolfssh/CMakeFiles/__idf_wolfssh.dir/src/ssh.c.obj
```

* Some vars not uses / not initialized:

```
/mnt/c/workspace/wolfssl_scripts-gojimmypi/espressif/wolfssh-component-publish/examples/wolfssh_server/components/wolfssh/src/ssh.c:1493:20: warning: 'PrivSuffix' defined but not used [-Wunused-variable]
 1493 | static const char* PrivSuffix = " PRIVATE KEY-----";
      |                    ^~~~~~~~~~
/mnt/c/workspace/wolfssl_scripts-gojimmypi/espressif/wolfssh-component-publish/examples/wolfssh_server/components/wolfssh/src/ssh.c:1491:20: warning: 'PrivBeginPrefix' defined but not used [-Wunused-variable]
 1491 | static const char* PrivBeginPrefix = "-----BEGIN ";
      |                    ^~~~~~~~~~~~~~~
/mnt/c/workspace/wolfssl_scripts-gojimmypi/espressif/wolfssh-component-publish/examples/wolfssh_server/components/wolfssh/src/ssh.c: In function 'DoSshPubKey':
[...]

PR #643, opened 2023-12-29
Improved "misc.c does not need to be compiled" message

gojimmypi:
As noted in https://github.com/wolfSSL/wolfssh/issues/633,  when building with the Espressif ESP-IDF, and the `misc.c` is not excluded in the `CMakeLists.txt`, I see this message:

```
[2/9] Building C object esp-idf/wolfssh/CMakeFiles/__idf_wolfssh.dir/C_/workspace/wolfssh-gojimmypi/src/misc.c.obj
C:/workspace/wolfssh-gojimmypi/src/misc.c:76:10: warning: #warning MISC_WARNING [-Wcpp]
   76 |         #warning MISC_WARNING
      |          ^~~~~~~
```

In particular, note the `MISC_WARNING` was previously defined as a descriptive string, but that string is not displayed at compile time.

This PR adds some macro logic to detect if `_Pragma` is supported, and if so the contents of `MISC_WARNING` is string-ified and displayed as appropriate.

Although this might have been simplified by simply using the same string in multiple places, there may be value in having the single definition of `MISC_WARNING` available elsewhere as needed. 

Using this change, the warning is manifested like this in my environment:

```
[2/9] Building C object esp-idf/wolfssh/CMakeFiles/__idf_wolfssh.dir/C_/workspace/wolfssh-gojimmypi/src/misc.c.obj
C:/workspace/wolfssh-gojimmypi/src/misc.c:83:9: note: '#pragma message: misc.c does not need to be compiled when using inline (NO_INLINE not defined))'
   83 |         _Pragma(MISC_TOSTRING(message(MISC_WARNING)))
      |         ^~~~~~~
[3/9] Building C object esp-idf
```

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -53,21 +53,11 @@
 /* Check for if compiling misc.c when not needed. */
 #if !defined(WOLFSSH_MISC_INCLUDED) && !defined(NO_INLINE) && \
     !defined(WOLFSSH_IGNORE_FILE_WARN)
-    #define MISC_WARNING "misc.c does not need to be compiled when using inline (NO_INLINE not defined))"
-    #define MISC_STRINGIFY(x) #x
-    #define MISC_TOSTRING(x) MISC_STRINGIFY(x)
-    #ifdef __STDC_VERSION__
-      #if __STDC_VERSION__ >= 199901L
-        #define PRAGMA_SUPPORTED 1
-      #endif
-    #endif
 
-    #if (defined(__GNUC__) || defined(__clang__)) && defined(PRAGMA_SUPPORTED)
-        _Pragma(MISC_TOSTRING(message(MISC_WARNING)))
-    #elif  defined(_MSC_VER)
-        #pragma message("warning: " MISC_WARNING)
+    #ifndef _MSC_VER
+        #warning "misc.c does not need to be compiled when using inline (NO_INLINE not defined))"
     #else
-        #warning MISC_WARNING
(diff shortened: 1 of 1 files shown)


dgarske (dismissed):
Any reason to not just use `WOLFSSL_IGNORE_FILE_WARN` in your user_settings.h? That option covers several other places too.

gojimmypi:
> Any reason to not just use `WOLFSSL_IGNORE_FILE_WARN` in your user_settings.h? That option covers several other places too.

Yes: my final solution was to exclude the `misc.c` in the build and not trigger the warning at all. The issue is that when I first saw the "MISC_WARNING" it was not very intuitive. This PR prints the intended message as desired. If there's going to be a warning, it should print the proper text.

ejohnstown:
I did the change for the MISC_WARNING string. I don't know why I did it that way, but it should probably go back to just being a string on the #warning directive. I suppose we could switch to using _Pragma, but that's going to be the first one. I don't care about duplicating the string there, it isn't compiled code.

ejohnstown (approved):

gojimmypi:
>  I suppose we could switch to using _Pragma, but that's going to be the first one.

I'm ok to go back to the simpler code & duplicate the compile-time text warning. There's otherwise no compelling reason to introduce new things. Just let me know which you prefer.

gojimmypi:
I've reverted the defined `MISC_WARNING` macro string to the simpler, duplicated text.

ejohnstown (approved):

PR #653, opened 2024-02-02
fix for build on windows, and fix for sftp file transfer from windows

JacobBarthelmeh:


douzzer (approved):
Thanks for getting this working.  `curX` and `curY` were never actually a thing...

Issue #659, opened 2024-02-22
Runtime configurability of key signature algorithms for KEX


falemagn:
The available algorithms can currently be configured only at compile time. It would be useful to have the possibility to compile them in, but enable/disable them at runtime in the context of KEX.

So that:
1. The server_host_key_algorithms namelist in SSH_MSG_KEXINIT includes the algorithm
2. If the algorithm is agreed upon by the client and server, then it's used to sign the key

The specific use case that prompted this feature request is the usage of ssh-rsa host keys: we'd like to be able to configure at runtime whether or not they can be signed with ssh-rsa or ssh-sha2-256 (and in the future with ssh-sha2-512).

dgarske:
See ZD 17568

falemagn:
Cannot access that

dgarske:
I was just adding that information for the assignment to @ejohnstown. @falemagn that's the ticket you filed in our Zendesk.

falemagn:
There must be a misunderstanding. As said, I've got no access to that ticket, because I have personally not filed it. If this conversation must be moved there, I suppose you need to do something to involve me in it.

ejohnstown:
You need to set the preprocessor define `WOLFSSH_YES_SSH_RSA_SHA1`. The code for `ssh-rsa` using SHA-1 is still available.

We disabled SHA-1 when OpenSSH announced that they were disabling it. SHA-1 is broken and provides no security. We decided to disable it as well.

falemagn:
It's clear that it can be enabled at compile time - that's the premise of [this very issue](https://github.com/wolfSSL/wolfssh/issues/659#issue-2148654026):

> The available algorithms can currently be configured only at compile time.

What I am proposing is to be able to disable/enable **at runtime** the  algorithms that have been compiled in, just like it happens with OpenSSH.

Btw, the [release note](https://www.wolfssl.com/wolfssh-1-4-14-release/) of the version that disabled the SHA1 signature didn't mention it. Was it mentioned anywhere else?

ejohnstown:
I'm going to close this issue since the PR adds the requested functions. I believe it achieves the desired goal.

PR #661, opened 2024-03-01
Cipher List Configuration

ejohnstown:
1. Add APIs to change the cipher lists used in the KEX init messages.
2. Add APIs to query the possible built in algorithms.
3. Reinstate the SHA-1 algorithm, but soft-disable it. It isn't offered by default but the user can set it for use.
(ZD 17568, Issue #659)

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1932,21 +1932,6 @@ int wolfSSH_ReadKey_file(const char* name,
 #endif
 
 
-int wolfSSH_CheckAlgoName(const char* name)
-{
-    int ret = WS_INVALID_ALGO_ID;
-
-    if (name) {
-        word32 nameSz = (word32)WSTRLEN(name);
-        if (NameToId(name, nameSz) != ID_UNKNOWN) {
-            ret = WS_SUCCESS;
-        }
-    }
-
-    return ret;
-}
-
-
 int wolfSSH_CTX_SetAlgoListKex(WOLFSSH_CTX* ctx, const char* list)
 {
     int ret = WS_SSH_CTX_NULL_E;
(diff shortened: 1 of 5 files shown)


src/ssh.c
```diff
@@ -1935,6 +1931,296 @@ int wolfSSH_ReadKey_file(const char* name,
...
+    return list;
+}
+
+
+int wolfSSH_CTX_SetAlgoListKeyAccepted(WOLFSSH_CTX* ctx, const char* list)
+{
+    int ret = WS_SSH_CTX_NULL_E;
+
+    if (ctx) {
+        ctx->algoListKeyAccepted = list;
```
JacobBarthelmeh, line 2140: NULL sanity checks on 'list' for these Setter functions.

JacobBarthelmeh: Should validation of the list be done? As is anything could be sent, for example if using `wolfSSH_CTX_SetAlgoListKey(ctx, "i-like-cheese")` the client then gets:

```
$ ssh -p 22222 jill@127.0.0.1
Unable to negotiate with 127.0.0.1 port 22222: no matching host key type found. Their offer: i-like-cheese
```

JacobBarthelmeh (changes requested):

src/ssh.c
ejohnstown: OpenSSH allows for a lot of extra options for setting algorithms, like "+name" to just add name to an existing list. I wanted these setters to be dead simple, and put the smarts in the application. (Kept simple for embedded devices.) That's why I added the functions to check values.

I also feel NULL is a valid value. It isn't a particularly useful value, but it is valid.

JacobBarthelmeh (approved):

PR #664, opened 2024-03-06
Use ENABLE_VIRTUAL_TERMINAL_PROCESSING on Windows treminal

JacobBarthelmeh:
ZD17595

Previously if setting a Windows terminal to "Windows Console Host", running the client.exe application and connecting to wolfSSHd (also running on Windows) the virtual terminal control sequences were not all handled well. Leading to the cursor jumping around in the terminal. This change makes use of Windows processing of virtual terminal control sequences instead.

### DIFF
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -404,6 +404,20 @@ static THREAD_RET readPeer(void* in)
     FD_SET(fd, &errSet);
 
 #ifdef USE_WINDOWS_API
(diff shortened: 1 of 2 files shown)


ejohnstown (approved):
Ran an A/B test between master and this PR and it works as described.

ejohnstown:
I wasn't sure what you meant by "setting a Windows terminal to 'Windows Console Host'", so I ran `conhost.exe` and used the client from there. Some hand-holding comments in the future would help.

PR #665, opened 2024-03-11
update libwolfssl search for older tools and static library

JacobBarthelmeh:
ZD17497

Issue of autotools finding a static libwolfssl was reported on Ubuntu 20 and reproduced on Ubuntu 18. On newer Ubuntu versions (and autotools) such as Ubuntu 22 it is able to find the static library without this change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -138,6 +138,11 @@ if test ${HAVE_WOLFSSL} = "0" ; then
     fi
 fi
 
+if test ${HAVE_WOLFSSL} = "1" ; then
+    LIBS="-lwolfssl ${LIBS}"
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
Steps to reproduce on Ubuntu 18 :

```
sudo rm -rf /usr/local/lib/libwolfssl*
cd wolfssl
./configure --enable-ssh --enable-static --disable-shared && make && sudo make install
cd ../wolfssh
./autogen.sh
./configure
```

JacobBarthelmeh:
Possibly should be solved by adding -lpthread and -lm when doing the search for libwolfssl.a on older Ubuntu OS's.

PR #673, opened 2024-03-26
Change references in README.md from John to Fred.

anhu:
Manually tested and double checked the text.


ejohnstown (changes requested):
It's a good change, except that those directions don't work. I'm going to add a commit to fix that.

ejohnstown (approved):

douzzer (approved):

PR #674, opened 2024-04-09
Changing Channels

ejohnstown:
This set of commits changes how the server code handles a connection accept and channel opening. Originally, opening a channel is part of the accept state machine. The server code now completes the accept after user authentication. This PR introduces a set of callback function hooks for channel opening, end-of-file, and closing, success and failure, as well as various channel requests. On top of this, the echoserver and wolfSSHd are updated to take advantage of the new callbacks.


apps/wolfsshd/wolfsshd.c
```diff
@@ -1771,81 +1829,59 @@ static void* HandleConnection(void* arg)
...
-                    if (ret == WS_SCP_INIT) {
-                    #ifdef WOLFSSH_SCP
-                        ret = SCP_Subsystem(conn, ssh, pPasswd, usrConf);
-                    #else
-                        err_sys("SCP not compiled in. Please use "
-                                "--enable-scp");
-                    #endif
-                    }
-                    break;
+        do {
```
JacobBarthelmeh, line 1832: needs sanity check on ret value to account for error in getting user info, before entering do while loop

JacobBarthelmeh: will this handle a peer just closing down the socket and disconnecting?

apps/wolfsshd/wolfsshd.c
```diff
@@ -1771,81 +1829,59 @@ static void* HandleConnection(void* arg)
...
+            err_sys("SCP not compiled in. Please use --enable-scp");
+        #endif
+        }
+        else if (conn->doExec) {
+            if (ret == WS_SUCCESS) {
+                wolfSSH_Log(WS_LOG_INFO, "[SSHD] Entering exec session [%s]",
+                    wolfSSH_GetSessionCommand(ssh));
+                SHELL_Subsystem(conn, ssh, pPasswd, usrConf,
+                        wolfSSH_GetSessionCommand(ssh));
             }
```
JacobBarthelmeh, line 1885: would be nice to have a final else case to log that an unknown/unsupported channel type was recieved

src/wolfsftp.c
```diff
@@ -1126,22 +1126,6 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
     if (ssh->error == WS_WANT_READ || ssh->error == WS_WANT_WRITE)
         ssh->error = WS_SUCCESS;
 
-    /* check accept is done, if not call wolfSSH accept */
-    if (ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED) {
```
JacobBarthelmeh, line 1130: I like having the easy one API call for SFTP accept. If removing it though should we also change the wolfSSH_SFTP_connect to match?

JacobBarthelmeh (changes requested):

src/wolfsftp.c
ejohnstown: I would like the client-connect side to match the way the server-accept side works at some point.

I believe the SFTP accept needs to happen after the handshake is complete. The wolfSSH_accept() should only do KEX and user authentication. After userauth, any sort of channel can be open. In fact, channel open doesn't specify the kind of channel. That's all channel requests when the channel has been allocated.

I guess an alternative could be to have wolfSSH_SFTP_accept() call wolfSSH_accept() first thing and if the SSH connection is established and authenticated, return success. Then roll into doing the SFTP accept actions. I don't think SFTP should know about the WOLFSSH's internal state and where its handshake is at.

apps/wolfsshd/wolfsshd.c
ejohnstown: wolfSSHd's channel request callbacks set those flags after the channel is created, it's guaranteed. Having a channel pointer and none of the three "do" flags sets is definitely an error condition. Adding the else clause.

apps/wolfsshd/wolfsshd.c
ejohnstown: If the client application authenticates the user then closes the socket, we should just close down. The client should also be able to send a MSG_DISCONNECT and the server should close out the connection.

ejohnstown: And you are right about the check on the return value.

JacobBarthelmeh: I probably should have been more clear about the peer just closing down. Disconnected as in the Ethernet wire was pulled or internet connection shut off. I think this do while loop would continue spinning on in that case since the channel is never opened?

src/wolfsftp.c
JacobBarthelmeh: marking as resolved, we will need to update the documentation and add the change behavior to the next release. Having the clean boundary between internal SSH operations and post handshake operations is nice though.

ejohnstown: I'm thinking this is a v1.5 thing. And I agree with the documentation update. I have another branch that works off this to make the servers even better.

ejohnstown:
This PR is now broken into several others, and one incomplete branch.

PR #682, opened 2024-04-26
Static Memory Pools

ejohnstown:
Update wolfSSH to use wolfCrypt's static memory allocation pools. Most of wolfSSH already passes around a heap value that is passed into the WOLFSSH_CTX as a NULL. The echoserver and sftpclient examples have been updated with static pool descriptors that work on macOS and Linux builds when only SFTP is available. Some other changes were made in the client common code to include a heap when loading keys, and the other clients were updated for this.
(Fixes 17726)

### DIFF
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -780,7 +780,7 @@ THREAD_RETURN WOLFSSH_THREAD client_test(void* args)
         err_sys("If setting priv key, need pub key.");
     }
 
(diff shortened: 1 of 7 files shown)


JacobBarthelmeh (approved):
Would be nice to have an --enable option for the feature. That can come later though. Tried it out.

Thanks John!

Issue #679, opened 2024-04-19
Error when building wolfssh with only WOLFSSH_NO_ECDSA or WOLFSSH_NO_ECDH


oleg35hp:
wolfssh/internal.h:598

```
#if !defined(WOLFSSH_NO_ECDSA) && !defined(WOLFSSH_NO_ECDH)
        ecc_key ecc;
#endif
```

should be

```
#if !defined(WOLFSSH_NO_ECDSA) || !defined(WOLFSSH_NO_ECDH)
        ecc_key ecc;
#endif
```

ejohnstown:
The actual solution is to check for `#ifndef WOLFSSH_NO_ECDH`. I have a fix in the works, and I'll link it where when it is a PR.

PR #684, opened 2024-04-30
Fix Guards

ejohnstown:
1. In `KexDhReply()`, make the ECDH vs ECDSA guards consistent and correct them for ECDH.
2. In the private key agreement handshake info, change the guard for the ecc key to only be on ECDH.
Fixes issue #679.


ejohnstown:
PR #686 supersedes this PR.

PR #685, opened 2024-05-01
Kyber Fix

ejohnstown:
1. Fix a couple spots where a flag to use kyber wasn't getting set correctly.
2. Changed a check to kyber back to the original source, a local variable.


anhu (dismissed):
Thank you!!

JacobBarthelmeh (approved):

PR #692, opened 2024-05-15
Kyber Testing

ejohnstown:
Add a workflow for a kyber testing. It's only a make check, but ECDSA+Kyber is the default KEX.

### DIFF
--- a/.github/workflows/kyber.yml
+++ b/.github/workflows/kyber.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ '*' ]
 
+env:
+  LIBOQS_REF: 0.10.0
+  WOLFSSL_REF: v5.7.0-stable
+
 jobs:
   build_liboqs:
     name: Build liboqs
@@ -13,11 +17,11 @@ jobs:
     timeout-minutes: 4
     steps:
       - name: Checking cache for liboqs
-        uses: actions/cache/restore@v4
+        uses: actions/cache@v4
         id: cache-liboqs
         with:
           path: opt/liboqs/
-          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
+          key: wolfssh-kyber-liboqs-${{ env.LIBOQS_REF }}-${{ runner.os }}
           lookup-only: true
 
       - name: Checkout liboqs
@@ -25,7 +29,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: open-quantum-safe/liboqs
-          ref: 0.10.0
+          ref: ${{ env.LIBOQS_REF }}
           path: liboqs
 
       - name: Build and install liboqs
@@ -38,24 +42,17 @@ jobs:
           make
           make install
 
-      - name: Stash liboqs in cache
-        if: steps.cache-liboqs.outputs.cache-hit != 'true'
-        uses: actions/cache/save@v4
-        with:
-          path: opt/liboqs/
-          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
-
   build_wolfssl:
     name: Build wolfssl
     runs-on: ubuntu-latest
     timeout-minutes: 4
     steps:
       - name: Checking cache for wolfssl
-        uses: actions/cache/restore@v4
+        uses: actions/cache@v4
         id: cache-wolfssl
         with:
           path: opt/wolfssl/
-          key: wolfssh-wolfssl-v5.7.0-stable-${{ runner.os }}
+          key: wolfssh-kyber-wolfssl-${{ env.WOLFSSL_REF }}-${{ runner.os }}
           lookup-only: true
 
       - name: Checkout, build, and install wolfssl
@@ -63,18 +60,17 @@ jobs:
         uses: wolfSSL/actions-build-autotools-project@v1
         with:
           repository: wolfssl/wolfssl
-          ref: v5.7.0-stable
+          ref: ${{ env.WOLFSSL_REF }}
           path: wolfssl
-          configure: --prefix=${{ github.workspace }}/opt/wolfssl --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
+          configure: --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
           check: false
           install: true
 
-      - name: Stash wolfssl in cache
+      - shell: bash
         if: steps.cache-wolfssl.outputs.cache-hit != 'true'
-        uses: actions/cache/save@v4
-        with:
-          path: opt/wolfssl/
-          key: wolfssh-wolfssl-v5.7.0-stable-${{ runner.os }}
+        run: |
+          mkdir -p opt
+          mv build-dir opt/wolfssl
 
   build_wolfssh:
     name: Build wolfssh
@@ -83,17 +79,17 @@ jobs:
     needs: [build_wolfssl, build_liboqs]
     steps:
       - name: Checking cache for liboqs
-        uses: actions/cache/restore@v4
+        uses: actions/cache@v4
         with:
           path: opt/liboqs/
-          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
+          key: wolfssh-kyber-liboqs-${{ env.LIBOQS_REF }}-${{ runner.os }}
           fail-on-cache-miss: true
 
       - name: Checking cache for wolfssl
-        uses: actions/cache/restore@v4
+        uses: actions/cache@v4
         with:
           path: opt/wolfssl/
-          key: wolfssh-wolfssl-v5.7.0-stable-${{ runner.os }}
+          key: wolfssh-kyber-wolfssl-${{ env.WOLFSSL_REF }}-${{ runner.os }}
           fail-on-cache-miss: true
 
       - name: Checkout, build, and test wolfssh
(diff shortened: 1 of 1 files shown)


.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+          ./configure --prefix=${{ github.workspace }}/opt --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
+          make
+          make install
+
+      - name: Upload wolfssl
+        uses: actions/upload-artifact@v4
+        with:
+          name: wolfssh-wolfssl
+          path: ${{ github.workspace}}/opt/
+          retention-days: 3
```
julek-wolfssl, line 55: There is no need to split out wolfssl building since its a one (maybe many in the future) to one relationship with wolfssh version.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+    branches: [ '*' ]
+
+jobs:
+  build_liboqs:
+    name: Build liboqs
+    runs-on: ubuntu-latest
+    timeout-minutes: 4
+    steps:
+      - name: Install liboqs source
+        run: git clone --depth 1 --branch "0.10.0" "https://github.com/open-quantum-safe/liboqs"
```
julek-wolfssl, line 16: Use actions/checkout to get repos.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+          path: ${{ github.workspace }}/opt/
+          retention-days: 3
+
+  build_wolfssl:
+    name: Build wolfssl
+    runs-on: ubuntu-latest
+    timeout-minutes: 4
+    steps:
+      - name: Install wolfssl source
+        run: git clone --depth 1 "https://github.com/wolfssl/wolfssl"
```
julek-wolfssl, line 40: Use actions/checkout to get repos.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+          path: ${{ github.workspace }}/opt/
+
+      - name: Download liboqs
+        uses: actions/download-artifact@v4
+        with:
+          name: wolfssh-liboqs
+          path: ${{ github.workspace }}/opt/
+
+      - name: Install wolfSSH
+        run: git clone --depth 1 "https://github.com/wolfssl/wolfssh"
```
julek-wolfssl, line 76: Use actions/checkout to get repos.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Install wolfssl source
+        run: git clone --depth 1 "https://github.com/wolfssl/wolfssl"
+
+      - name: Build and install wolfssl
+        working-directory: wolfssl
+        run: |
+          autoreconf -ivf
+          ./configure --prefix=${{ github.workspace }}/opt --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
+          make
+          make install
```
julek-wolfssl, line 48: When building with autotools use the wolfSSL/actions-build-autotools-project action.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Build wolfSSH
+        working-directory: wolfssh
+        run: |
+          autoreconf -ivf
+          ./configure --with-liboqs=${{ github.workspace }}/opt LDFLAGS=-L${{ github.workspace }}/opt/lib CPPFLAGS=-I${{ github.workspace }}/opt/include
+          make
+
+      - name: Run wolfssh tests
+        working-directory: wolfssh
+        run: make check
```
julek-wolfssl, line 87: Also use wolfSSL/actions-build-autotools-project@v1 to build and test.

julek-wolfssl (changes requested):

.github/workflows/kyber.yml
ejohnstown: I prefer to split it out, and I'm caching it. I'm testing against a specific version of wolfSSL. I'm using a cache, specifically, instead of an artifact.

julek-wolfssl: In that case that is a good choice.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,105 @@
...
+          cmake -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/opt/liboqs -DOQS_MINIMAL_BUILD=KEM_kyber_512 -DOQS_USE_OPENSSL=0 ..
+          make
+          make install
+
+      - name: Stash liboqs in cache
+        if: steps.cache-liboqs.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: opt/liboqs/
+          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
```
julek-wolfssl, line 46: You can just use the `actions/cache` to check and save the cache. It will automatically update the cache if the path is present.

```suggestion
      - name: Checking cache for liboqs
        uses: actions/cache@v4
        id: cache-liboqs
        with:
          path: opt/liboqs/
          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
          lookup-only: true

      - name: Checkout liboqs
        if: steps.cache-liboqs.outputs.cache-hit != 'true'
        uses: actions/checkout@v4
        with:
          repository: open-quantum-safe/liboqs
          ref: 0.10.0
          path: liboqs

      - name: Build and install liboqs
        if: steps.cache-liboqs.outputs.cache-hit != 'true'
        working-directory: liboqs
        run: |
          mkdir build
          cd build
          cmake -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/opt/liboqs -DOQS_MINIMAL_BUILD=KEM_kyber_512 -DOQS_USE_OPENSSL=0 ..
          make
          make install
```

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,105 @@
...
+          configure: --prefix=${{ github.workspace }}/opt/wolfssl --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
+          check: false
+          install: true
+
+      - name: Stash wolfssl in cache
+        if: steps.cache-wolfssl.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: opt/wolfssl/
+          key: wolfssh-wolfssl-v5.7.0-stable-${{ runner.os }}
```
julek-wolfssl, line 77: Same as above.
```suggestion
      - name: Checking cache for wolfssl
        uses: actions/cache@v4
        id: cache-wolfssl
        with:
          path: opt/wolfssl/
          key: wolfssh-wolfssl-v5.7.0-stable-${{ runner.os }}
          lookup-only: true

      - name: Checkout, build, and install wolfssl
        if: steps.cache-wolfssl.outputs.cache-hit != 'true'
        uses: wolfSSL/actions-build-autotools-project@v1
        with:
          repository: wolfssl/wolfssl
          ref: v5.7.0-stable
          path: wolfssl
          configure: --prefix=${{ github.workspace }}/opt/wolfssl --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
          check: false
          install: true
```

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,105 @@
...
+    name: Build liboqs
+    runs-on: ubuntu-latest
+    timeout-minutes: 4
+    steps:
+      - name: Checking cache for liboqs
+        uses: actions/cache/restore@v4
+        id: cache-liboqs
+        with:
+          path: opt/liboqs/
+          key: wolfssh-liboqs-0.10.0-${{ runner.os }}
```
julek-wolfssl, line 20: For static versions of liboqs and wolfssl, I recommend setting environmental variables at the top to make it easier to update in the future. Look here for an example: https://github.com/wolfSSL/wolfssl/blob/db383519199dc9fab626df96da9ead1441bc02d7/.github/workflows/hostap-vm.yml#L15-L16. It is referenced in the action with `${{ env.LINUX_REF }}`. I would recommend something like this:
```
env:
  LIBOQS_REF: 0.10.0
  WOLFSSL_REF: v5.7.0-stable
```

julek-wolfssl (changes requested):
Very nice fixes. Almost there.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,105 @@
...
+          lookup-only: true
+
+      - name: Checkout, build, and install wolfssl
+        if: steps.cache-wolfssl.outputs.cache-hit != 'true'
+        uses: wolfSSL/actions-build-autotools-project@v1
+        with:
+          repository: wolfssl/wolfssl
+          ref: v5.7.0-stable
+          path: wolfssl
+          configure: --prefix=${{ github.workspace }}/opt/wolfssl --enable-wolfssh --enable-cryptonly --disable-examples --disable-crypttests
```
julek-wolfssl, line 68: The prefix won't work because the action sets its own prefix https://github.com/wolfSSL/actions-build-autotools-project/blob/905111b5a8be134fed11e01362427534f36febb7/action.yml#L85. The location is `$GITHUB_WORKSPACE/build-dir`. You can move the dir after this step if you still want it at `$GITHUB_WORKSPACE/opt/wolfssl`.

julek-wolfssl (changes requested):

julek-wolfssl (approved):
Nice job!

anhu (approved):

dgarske (approved):

PR #698, opened 2024-05-22
Enable sha1 when WOLFSSH_NO_SHA1_SOFT_DISABLE is enabled

miyazakh:
sha1 is only enabled when user intentionally enables NO_SHA1_SOFT_DISABLE

ZD#17793

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -13490,7 +13490,8 @@ static int PrepareUserAuthRequestPublicKey(WOLFSSH* ssh, word32* payloadSz,
         #ifndef WOLFSSH_NO_RSA_SHA2_256
             algoId[algoIdSz++] = ID_RSA_SHA2_256;
         #endif
-        #ifndef WOLFSSH_NO_SSH_RSA_SHA1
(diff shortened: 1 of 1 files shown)


JacobBarthelmeh:
@miyazakh can I have some background on this, what was the commands used to test it? Did you test that the soft enable of sha1 still works as expected after the change?

ejohnstown:
I checked that sha-1 based algorithms were listed in the KEX initial message. I also disabled the ECDSA and Ed25519 algorithms so only ssh-rsa was available.

miyazakh:
Hi @JacobBarthelmeh 
Sorry for being the late input. This becomes failure when accessing OpenSSH when its version is less than 8.x under using public key client authentication.

Regarding the soft enable of sha1, I am going to check it again.

ejohnstown (approved):
I approve this. I also have a branch where I fixed this the same exact way and forgot to follow up with a PR.

JacobBarthelmeh (approved):

PR #703, opened 2024-05-30
improvements for ipv6

anhu:
Fixes ZD18053

### DIFF
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -991,13 +991,8 @@ static THREAD_RETURN WOLFSSH_THREAD wolfSSH_Client(void* args)
         err_sys("Couldn't set the username.");
 
     build_addr(&clientAddr, config.hostname, config.port);
-    tcp_socket(&sockFd, 1,
-#ifdef TEST_IPV6
-               clientAddr.sin6_family
-#else
-               clientAddr.sin_family
-#endif
-              );
+    tcp_socket(&sockFd, ((struct sockaddr_in *)&clientAddr)->sin_family);
+
     ret = connect(sockFd, (const struct sockaddr *)&clientAddr, clientAddrSz);
     if (ret != 0)
         err_sys("Couldn't connect to server.");
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -991,7 +991,13 @@ static THREAD_RETURN WOLFSSH_THREAD wolfSSH_Client(void* args)
         err_sys("Couldn't set the username.");
 
     build_addr(&clientAddr, config.hostname, config.port);
-    tcp_socket(&sockFd);
+    tcp_socket(&sockFd, 1,
+#ifdef TEST_IPV6
+               clientAddr.sin6_family
+#else
+               clientAddr.sin_family
+#endif
+              );
     ret = connect(sockFd, (const struct sockaddr *)&clientAddr, clientAddrSz);
     if (ret != 0)
         err_sys("Couldn't connect to server.");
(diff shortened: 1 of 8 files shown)


anhu:
Have confirmation from ZD Ticket that this change is working as expected. 
Fixups squashed and now ready for merge.

dgarske (approved):
Over to @ejohnstown

apps/wolfsshd/wolfsshd.c
```diff
@@ -2396,13 +2396,13 @@ static int StartSSHD(int argc, char** argv)
                 conn->fd = (int)accept(listenFd, (struct sockaddr*)&clientAddr,
                     &clientAddrSz);
                 if (conn->fd >= 0) {
-                    inet_ntop(AF_INET,
 #ifdef TEST_IPV6
-                              &clientAddr.sin6_addr,
+                    inet_ntop(AF_INET, &clientAddr.sin6_addr, conn->ip,
```
ejohnstown, line 2400: This doesn't actually work. inet_ntop() is going to take the first 32-bits of sin6_addr and turn it into a dotted-quad string.

wolfssh/test.h
```diff
@@ -546,7 +546,7 @@ static INLINE void build_addr(SOCKADDR_IN_T* addr, const char* peer,
 #endif
 
 
-static INLINE void tcp_socket(WS_SOCKET_T* sockFd)
+static INLINE void tcp_socket(WS_SOCKET_T* sockFd, int specifyProtocol, int targetProtocol)
```
ejohnstown, line 549: Can we make this a little smarter? Maybe have tcp_socket() and tcp6_socket() that call a utility function?

apps/wolfssh/wolfssh.c
```diff
@@ -991,7 +991,13 @@ static THREAD_RETURN WOLFSSH_THREAD wolfSSH_Client(void* args)
         err_sys("Couldn't set the username.");
 
     build_addr(&clientAddr, config.hostname, config.port);
-    tcp_socket(&sockFd);
+    tcp_socket(&sockFd, 1,
+#ifdef TEST_IPV6
```
ejohnstown, line 995: These checks are really awkward. TEST_IPV6 is meant to force the example tools to use IPV6 for testing. It was a quick-and-dirty way to support IPv6. In general, we shouldn't need checks like this. Information about the IP address type are kept with the IP address. (clientAddr.sin_family)

ejohnstown (changes requested):

apps/wolfssh/wolfssh.c
anhu: I'm not sure how to make it less awkward.  Since we have the typedef for `SOCKADDR_IN_T` , the check of the member will need this macro wrap. (clientAddr.sin_family vs. clientAddr.sin6_family).

ejohnstown (approved):

apps/wolfsshd/wolfsshd.c
```diff
@@ -2374,8 +2374,8 @@ static int StartSSHD(int argc, char** argv)
 #ifdef WOLFSSL_NUCLEUS
             struct addr_struct clientAddr;
 #else
-            SOCKADDR_IN_T clientAddr;
-            socklen_t     clientAddrSz = sizeof(clientAddr);
+            struct sockaddr_in6 clientAddr;
```
dgarske, line 2377: Is there a reason we don't have all this wrapped with `TEST_IPV6` and use `SOCKADDR_IN_T` in the non IPv6 case? I noticed a bit of a mess between the examples for the IPv4 and IPv6 support.

dgarske (changes requested):

apps/wolfsshd/wolfsshd.c
anhu: Already discussed internally.  No need for changes.  Back to @dgarske.

dgarske: The ssd will always support IPv6

dgarske (approved):

PR #705, opened 2024-06-01
Using Decode Raw RSA Private Key

ejohnstown:
1. Add a check for wc_RsaPrivateKeyDecodeRaw() to configure.
2. If wc_RsaPrivateKeyDecodeRaw() is available (from PR https://github.com/wolfSSL/wolfssl/pull/7608), use it to load the private key from GetOpenSshKeyRsa(). If unavailable, process the key the original way.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -33,6 +33,7 @@
 #include <wolfssh/ssh.h>
 #include <wolfssh/internal.h>
 #include <wolfssh/log.h>
+#include <wolfssl/version.h>
 #include <wolfssl/wolfcrypt/asn.h>
 #ifndef WOLFSSH_NO_DH
     #include <wolfssl/wolfcrypt/dh.h>
@@ -45,7 +46,6 @@
 #include <wolfssl/wolfcrypt/rsa.h>
 #include <wolfssl/wolfcrypt/ecc.h>
 #include <wolfssl/wolfcrypt/hmac.h>
-#include <wolfssl/wolfcrypt/integer.h>
 #include <wolfssl/wolfcrypt/signature.h>
 
 #ifdef WOLFSSH_HAVE_LIBOQS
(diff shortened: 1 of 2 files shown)


dgarske (dismissed):

ejohnstown:
I need to I'm leaving out the right headers if that new function exists.

src/internal.c
```diff
@@ -1353,6 +1352,46 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap)
...
+
+    if (ret != WS_SUCCESS)
+        ret = WS_RSA_E;
+
+    return ret;
+}
+
+#else /* HAVE_WC_RSAPRIVATEKEYDECODERAW */
+
+#include <wolfssl/wolfcrypt/integer.h>
```
dgarske, line 1393: You can just include `<wolfssl/wolfcrypt/wolfmath.h>` now to handle all cases. Its possible integer.h (heap math) could be removed.

src/internal.c
```diff
@@ -1353,6 +1352,46 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap)
 
 
 #ifndef WOLFSSH_NO_RSA
+
+#ifdef HAVE_WC_RSAPRIVATEKEYDECODERAW
```
dgarske, line 1356: This is not a portable way to detect this... meaning someone not using configure won't detect or use this new API. But its probably still the best option for now.

src/internal.c
```diff
@@ -1353,6 +1352,46 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap)
...
+    if (ret != WS_SUCCESS)
+        ret = WS_RSA_E;
+
+    return ret;
+}
+
+#else /* HAVE_WC_RSAPRIVATEKEYDECODERAW */
+
+#include <wolfssl/wolfcrypt/integer.h>
+
```
dgarske, line 1394: In the `#else` case please add warning if `RSA_LOW_MEM` is defined about only being supported with `HAVE_WC_RSAPRIVATEKEYDECODERAW` and `wc_RsaPrivateKeyDecodeRaw` added in wolfSSL PR 7608.

dgarske (changes requested):

src/internal.c
ejohnstown: When RSA_LOW_MEM is set, skipping reading the `u` value of the key and not calculating dP and dQ.

src/internal.c
ejohnstown: Switched this to checking that wolfSSL is greater than v5.7.0. The new function isn't available in v5.7.0.

src/internal.c
ejohnstown: Good idea.

dgarske (approved):

PR #706, opened 2024-06-02
Handle signing failure when signing a hash with RSA

ljanyst:
This results with malloc failures if the host private key is too large


ejohnstown:
Can I ask why? Are you getting a typecast error for that sign result?

ejohnstown:
Thank you for the bug report. I have fixed this issue with PR #708.

PR #709, opened 2024-06-05
RSA Verify Refactor

ejohnstown:
# Description

Some implementations of SSH will remove all leading zeros from an encoded RSA signature, which is allowed per the RFCs. Our verify function requires the signature to be the same size as the key. This refactor will pad a signature with leading zeros if it is short. (ZD 18095)

# Testing

Run the echoserver and the PuTTY plink tools:

    ./examples/echoserver/echoserver -I putty:keys/putty_rsa.pub
    plink -batch -P 22222 -l putty -i keys/putty_rsa.ppk localhost
    while $(echo | plink -batch -P 22222 -l putty -i keys/putty_rsa.ppk localhost) ; do : ; done

The `while` version will repeatedly call the command. There's a 1/256 chance of the signature failing before the fix.


src/internal.c
```diff
@@ -6419,27 +6385,44 @@ static int DoUserAuthRequestRsaCert(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
...
 
+    /* Check that the pubkey's type matches the expected one. */
+    if (ret == WS_SUCCESS) {
+        ret = GetStringRef(&publicKeyTypeSz, &publicKeyType,
+                pk->publicKey, pk->publicKeySz, &i);
+    }
+
+    if (ret == WS_SUCCESS) {
+        if (publicKeyTypeSz != 14 &&
+            WMEMCMP(publicKeyType, "x509v3-ssh-rsa", 14) != 0) {
```
dgarske, line 6417: Would be nice to have the `"x509v3-ssh-rsa"` as a macro and also use XSTRLEN. Is this a new check? Do you expect any compatibility issues?

src/internal.c
```diff
@@ -10523,38 +10477,86 @@ static INLINE byte SigTypeForId(byte id)
...
-        ForceZero(check, digestSz);
-        WFREE(check, heap, DYNTYPE_TEMP);
     }
 
+#ifdef WOLFSSH_SMALL_STACK
+    if (checkDigest)
+        WFREE(checkDigest, heap, DYNTYPE_BUFFER);
+#endif
+    if (checkSig)
+        WFREE(checkSig, heap, DYNTYPE_BUFFER);
```
dgarske, line 10559: Should be `DYNTYPE_TEMP`

dgarske (changes requested):

JacobBarthelmeh (approved):

src/internal.c
ejohnstown: Double checking this one. I'm comparing to the ECDSA version of the function and something is fishy.

ejohnstown: I do want to double check the strings and use names for the strings in general, I think that's outside the scope of this PR.

PR #710, opened 2024-06-10
add check for channel closed error

JacobBarthelmeh:
ZD 18129

### DIFF
--- a/examples/scpclient/scpclient.c
+++ b/examples/scpclient/scpclient.c
@@ -313,7 +313,9 @@ THREAD_RETURN WOLFSSH_THREAD scp_client(void* args)
 
     ret = wolfSSH_shutdown(ssh);
     /* do not continue on with shutdown process if peer already disconnected */
-    if (ret != WS_SOCKET_ERROR_E && wolfSSH_get_error(ssh) != WS_SOCKET_ERROR_E) {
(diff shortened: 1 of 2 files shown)


examples/scpclient/scpclient.c
```diff
@@ -313,7 +313,9 @@ THREAD_RETURN WOLFSSH_THREAD scp_client(void* args)
 
     ret = wolfSSH_shutdown(ssh);
     /* do not continue on with shutdown process if peer already disconnected */
-    if (ret != WS_SOCKET_ERROR_E && wolfSSH_get_error(ssh) != WS_SOCKET_ERROR_E) {
+    if (ret != WS_CHANNEL_CLOSED && ret != WS_SOCKET_ERROR_E &&
```
JacobBarthelmeh, line 316: Should wolfSSH_shutdown be returning WS_CHANNEL_CLOSED instead of WS_SUCCESS when the channel is closed?

ejohnstown: Channels first report they have reached End Of File first. After both ends of the channel EOF, they close.

As far as what should wolfSSH_shutdown return? I'm not sure yet. I need to get that Changing Channels PR in. The shutdown process will make more sense with the callbacks.

JacobBarthelmeh: Updated to return WS_CHANNEL_CLOSED

ejohnstown (approved):

PR #718, opened 2024-07-03
wolfSSHd Connection Closure

ejohnstown:
# DESCRIPTION

Improvements to wolfSSHd to shutdown the connection cleaner, without disconnection errors. The server will keep processing the channel data and close its side of the connection. On socket shutdown, it will `shutdown` its side of the socket and read data from the socket until 0 return, then it closes the socket. Will also interrupt the child process if there are any socket errors.
(ZD 17722)

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -247,7 +247,7 @@ static byte* getBufferFromFile(const char* fileName, word32* bufSz, void* heap)
 
     if (WFOPEN(NULL, &file, fileName, "rb") != 0)
         return NULL;
-    WFSEEK(NULL, file, 0, XSEEK_END);
+    WFSEEK(NULL, file, 0, WSEEK_END);
     fileSz = (word32)WFTELL(NULL, file);
     WREWIND(NULL, file);
 
@@ -1158,7 +1158,14 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     byte channelBuffer[EXAMPLE_BUFFER_SZ];
     char* forcedCmd;
     int   windowFull = 0;
-    int   idle = 0;
+    int   peerConnected = 1;
+    int   stdoutEmpty = 0;
+
+    childFd = -1;
+    stdoutPipe[0] = -1;
+    stdoutPipe[1] = -1;
+    stderrPipe[0] = -1;
+    stderrPipe[1] = -1;
 
     forcedCmd = wolfSSHD_ConfigGetForcedCmd(usrConf);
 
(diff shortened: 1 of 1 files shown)


apps/wolfsshd/wolfsshd.c
```diff
@@ -1313,12 +1320,11 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         }
 
         /* default to /bin/sh if user shell is not set */
-        WMEMSET(cmd, 0, sizeof(cmd));
-        if (XSTRLEN(pPasswd->pw_shell) == 0) {
-            XSNPRINTF(cmd, sizeof(cmd), "%s", "/bin/sh");
+        if (pPasswd->pw_shell && XSTRLEN(pPasswd->pw_shell)) {
```
douzzer, line 1323: should be `WSTRLEN()`.

apps/wolfsshd/wolfsshd.c
```diff
@@ -1313,12 +1320,11 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
...
+            XSNPRINTF(cmd, sizeof(cmd), "%s", pPasswd->pw_shell);
```
douzzer, line 1324: should be `WSNPRINTF()`.

apps/wolfsshd/wolfsshd.c
```diff
@@ -1313,12 +1320,11 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
...
         }
         else {
-            XSNPRINTF(cmd, sizeof(cmd),"%s", pPasswd->pw_shell);
+            XSNPRINTF(cmd, sizeof(cmd), "%s", "/bin/sh");
```
douzzer, line 1327: ditto

douzzer (changes requested):
looks good.  a stray `XSTRLEN()` and a couple `XSNPRINTF()`s to fix.

douzzer (approved):

apps/wolfsshd/wolfsshd.c
```diff
@@ -1882,6 +1915,13 @@ static void* HandleConnection(void* arg)
     /* check if there is a response to the shutdown */
     wolfSSH_free(ssh);
     if (conn != NULL) {
+        byte sc[1024];
+        shutdown(conn->fd, 1);
+        /* Spin until socket closes. */
+        do {
+            ret = (int)recv(conn->fd, sc, 1024, 0);
```
douzzer, line 1922: Need to test for error (`ret` `-1`) here else this will get stuck in an infinite loop.

douzzer (changes requested):

douzzer (approved):

PR #720, opened 2024-07-08
SSHD Banners

ejohnstown:
# DESCRIPTION

Add support for loading banner files into wolfSSHd. (ZD 18156)

# TESTING

Add "Banner filename.txt" to the sshd_config file used by wolfSSHd.

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -103,7 +103,6 @@ static WFILE* logFile = NULL;
 
 /* catch interrupts and close down gracefully */
 static volatile byte quit = 0;
-static const char defaultBanner[] = "wolfSSHD\n";
 
 /* Initial connection information to pass on to threads/forks */
 typedef struct WOLFSSHD_CONNECTION {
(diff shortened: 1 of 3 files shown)


apps/wolfsshd/wolfsshd.c
```diff
@@ -273,13 +263,30 @@ static int UserAuthResult(byte result,
         WS_UserAuthData* authData, void* userAuthResultCtx);
 
 
+/* Frees up the WOLFSSH_CTX struct */
+static void CleanupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
+        byte** banner)
+{
+    if (banner != NULL && *banner != NULL) {
+#ifndef NO_FILESYSTEM
+        freeBufferFromFile(*banner, NULL);
```
JacobBarthelmeh, line 272: This should be in the config for the specific user I think. To handle something like this in the future:

```
Match User sshUser
    Banner /path/to/specific_banner
```

JacobBarthelmeh (changes requested):

apps/wolfsshd/wolfsshd.c
JacobBarthelmeh: Opps sorry, looks like it is setup that way. Just init/free different than I expected. Please double check the XFREE call though after CleanupCTX (with *banner set to NULL here).

douzzer (approved):

PR #739, opened 2024-09-23
Modes Update

ejohnstown:
Several of the TTY mode bits we are checking and setting aren't a part of POSIX, but are available on Linux. Adds guards to the non-POSIX value checks and updates.


src/internal.c
```diff
@@ -8253,10 +8267,14 @@ int wolfSSH_DoModes(const byte* modes, word32 modesSz, int fd)
...
                 break;
             case WOLFSSH_ECHOCTL:
-                TTY_SET_FLAG(term.c_lflag, ECHOCTL, arg);
+                #ifdef ECHOCTL
+                    TTY_SET_FLAG(term.c_lflag, ECHOCTL, arg);
+                #endif
                 break;
             case WOLFSSH_ECHOKE:
-                TTY_SET_FLAG(term.c_lflag, ECHOKE, arg);
+                #ifdef WCHOKE
```
JacobBarthelmeh, line 8275: Should this be `#ifdef ECHOKE`?

JacobBarthelmeh (approved):

src/internal.c
ejohnstown: Yup.

JacobBarthelmeh (approved):

PR #743, opened 2024-09-30
keep trailing delimiter with SFTP open

JacobBarthelmeh:
ZD 18566

Fix for SFTP cases with trailing delimiter.

OpenSSH client 9.6 (which does a SFTP command instead of SCP, -O restores using SCP protocol)
```
mkdir /tmp/subfolder
scp /tmp/test.txt jill@127.0.0.1:/tmp/subfolder/test/
```

Before this change wolfSSH would allow test.txt to be copied to /tmp/subfolder/test. After this change the command correctly fails.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -15720,16 +15720,22 @@ int wolfSSH_GetPath(const char* defaultPath, byte* in, word32 inSz,
     }
 
     if (out != NULL) {
-        WMEMCPY(out + curSz, in, inSz);
+        if (curSz + inSz < *outSz) {
+            WMEMCPY(out + curSz, in, inSz);
+        }
     }
     curSz += inSz;
     if (out != NULL) {
         if (!WOLFSSH_SFTP_IS_DELIM(out[0])) {
-            WMEMMOVE(out+1, out, curSz);
-            out[0] = '/';
+            if (curSz + 1 < *outSz) {
+                WMEMMOVE(out+1, out, curSz);
+                out[0] = '/';
+            }
             curSz++;
         }
-        out[curSz] = 0;
+        if (curSz < *outSz) {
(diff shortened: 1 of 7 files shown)


src/internal.c
```diff
@@ -15686,6 +15686,57 @@ int SendChannelSuccess(WOLFSSH* ssh, word32 channelId, int success)
...
+        WMEMCPY(out + curSz, in, inSz);
+    }
+    curSz += inSz;
+    if (out != NULL) {
+        if (!WOLFSSH_SFTP_IS_DELIM(out[0])) {
+            WMEMMOVE(out+1, out, curSz);
+            out[0] = '/';
+            curSz++;
+        }
+        out[curSz] = 0;
```
JacobBarthelmeh, line 15732: sanity checks needed on curSz before use (thanks @douzzer )

src/internal.c
```diff
@@ -15686,6 +15686,63 @@ int SendChannelSuccess(WOLFSSH* ssh, word32 channelId, int success)
...
+        if (!WOLFSSH_SFTP_IS_DELIM(out[0])) {
+            if (curSz + 1 < *outSz) {
+                WMEMMOVE(out+1, out, curSz);
+                out[0] = '/';
+            }
+            curSz++;
+        }
+        if (curSz < *outSz) {
+            out[curSz] = 0;
+        }
```
douzzer, line 15738: As discussed, when `*outSz` is to small to accommodate the shifted data with the terminating null, an error code should be returned, and checked for by callers.

douzzer (changes requested):

dgarske (approved):

douzzer (approved):

PR #751, opened 2024-11-13
SCP hang with interop fix

JacobBarthelmeh:
ZD18942

### DIFF
--- a/.github/workflows/scp-test.yml
+++ b/.github/workflows/scp-test.yml
@@ -95,6 +95,10 @@ jobs:
       - name: Setup test user
         run: sudo useradd -p password jak
 
+      # This is a regression test for a reported hang with SCP file transfer.
+      #     Verifying the error from a bad directory is propogated back, and
+      #     that the direcotry/file is not created. To account for potential
+      #     hanging of the operation, the timeout is set to 1 minute.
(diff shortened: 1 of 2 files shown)


.github/workflows/scp-test.yml
```diff
@@ -0,0 +1,107 @@
...
+          ./configure --enable-all LDFLAGS="-L${{ github.workspace }}/build-dir/lib" CPPFLAGS="-I${{ github.workspace }}/build-dir/include -DWOLFSSH_NO_FPKI -DWOLFSSH_NO_SFTP_TIMEOUT -DMAX_PATH_SZ=120"
+
+      - name: make
+        working-directory: ./wolfssh/
+        run: make
+
+      - name: Setup test user
+        run: sudo useradd -p password jak
+
+      - name: Run SCP example test
```
LinuxJedi, line 98: More a comment than anything else, but might be good to either explain what we are testing for / why we are testing this here, even if it is just a code comment in the YAML. Someone else retrospectively trying to find out why this test is here might be tough.

Apart from that, good catch! :)

LinuxJedi (approved):
Thanks for adding the comment.

douzzer (approved):

PR #758, opened 2025-01-15
Simplify autogen

ejohnstown:
The autogen script is doing too many things. The tests that are set up in the git hooks are already tested by GitHub with actions on branch push. Also they don't work if you install wolfSSL somewhere specific. All one needs to do is run `autoreconf -ivf` to make the configure script. This is all autogen does now, with the addition of warnings.


JacobBarthelmeh (approved):

gojimmypi:
Confirmed working on WSL.  Nice change to simplify things.

Thanks @ejohnstown for assistance regarding my error when using `~/` in the path.

```
 ./configure --with-wolfssl=~/wolfssl-install-dir
```

that resulted in:

```
[...]
checking for forkpty in -lutil... yes
checking for wolfSSL... configure: prefix NONE
configure: error: wolfSSL path error (~/wolfssl-install-dir/): missing lib and include
```

Using a fully-qualified path worked:

```
 ./configure --with-wolfssl=/home/gojimmypi/wolfssl-install-dir-with-ssh
```

PR #763, opened 2025-01-31
RFC-4256 Keyboard-Interactive authentication

LinuxJedi:
This implements Keyboard-Interactive authentication.

Adds an additional callback set by `wolfSSH_KeyboarAuthPrompts()` which will set a callback in the server to ask the application to provide the prompt details for the client.

PR #767, opened 2025-02-04
Add DH Group 16 and HMAC-SHA2-512

LinuxJedi:
This adds the `diffie-hellman-group16-sha512` key exchange and `hmac-sha2-512` mac support.

Echoserver can now take `-x` for key exchange and `-m` for mac setting, so that this can be used in the test suite.

New test added for this and it has been tested against OpenSSH.


LinuxJedi:
Note, this is expected to conflict with #763 around the test area, I can fix that when it happens.

LinuxJedi:
Moved to draft, there is a bug in HMAC-SHA2-512 which I am investigating.

LinuxJedi:
Problem found, should be good now.

ejohnstown (dismissed):

LinuxJedi:
Rebased to fix merge conflicts.

JacobBarthelmeh (approved):

Issue #772, opened 2025-02-10
piping data over wolfsshd stalls/fails


Forza-tng:
Hi,

I am having difficulties piping data from a openssh client to a  wolfsshd host.

For example, the following command never finishes.

```
cat test | ssh 192.168.0.1 -p 1234 'cat > xxxxxxxx'
```
ctrl-c aborts the connection. The test file is 48894 bytes, but 8188 bytes only makes it to the `xxxxxxx` file. The same happens using `mbuffer -i file ...` 

It seems to always be `8188` bytes, even if I cat a large ISO or some text files. However when doing `cat /dev/random` or `/dev/urandom`, the command finishes but no remote file is created at all.

I built wolfssh from a git checkout today using `./configure --prefix=/opt/wolfssh --enable-all --with-wolfssl=/opt/wolfssl`

I use the following `sshd_config`
```
Port 1234
PermitRootLogin yes
PasswordAuthentication yes
LoginGraceTime 30
HostKey /opt/wolfssh/etc/host_key_ed25519.pem
HostKey /opt/wolfssh/etc/host_key_ecdsa.pem
HostKey /opt/wolfssh/etc/host_key_rsa.pem
```
I start wolfsshd using ` /opt/wolfssh/bin/wolfsshd -d -E /tmp/wolfsshd.log  -f /opt/wolfssh/etc/sshd_config`

Logfile contains
```
...
[PID 67002]: [SSHD] Incoming TCP data found
[PID 81519]: [SSHD] Checking user name root
[PID 81519]: [SSHD] Unix check user
[PID 81519]: [SSHD] User ok.
[PID 81519]: [SSHD] Attempting to raise permissions level
[PID 81519]: [SSHD] Public key ok.
[PID 81519]: [SSHD] Lowering permissions level
[PID 81519]: [SSHD] Checking user name root
[PID 81519]: [SSHD] Unix check user
[PID 81519]: [SSHD] User ok.
[PID 81519]: [SSHD] Attempting to raise permissions level
[PID 81519]: [SSHD] Public key ok.
[PID 81519]: [SSHD] Lowering permissions level
[PID 81519]: [SSHD] grace time = 30 timeout = 0
[PID 81519]: [SSHD] Entering exec session [cat > /xxxxxxxx]
[PID 81519]: [SSHD] Attempting to raise permissions level
[PID 81519]: [SSHD] Attempting to close down connection
[PID 81519]: [SSHD] Return from closing connection = 0
```

[...]

PR #776, opened 2025-02-17
fix for FD_SET call on pipes and handling of channel EOF

JacobBarthelmeh:
https://github.com/wolfSSL/wolfssh/issues/772

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -57,6 +57,11 @@
     #define WOLFSSHD_TIMEOUT 1
 #endif
 
+#ifdef EXAMPLE_BUFFER_SZ
+    #warning use WOLFSSHD_SHELL_BUFFER_SZ instead of EXAMPLE_BUFFER_SZ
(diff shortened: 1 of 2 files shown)


apps/wolfsshd/wolfsshd.c
```diff
@@ -816,7 +816,8 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     BOOL ret;
     word32 shellChannelId = 0;
 #ifndef EXAMPLE_BUFFER_SZ
-#define EXAMPLE_BUFFER_SZ 4096
+    /* default to try and read max packet size */
+    #define EXAMPLE_BUFFER_SZ 32768
```
LinuxJedi, line 820: Long term, we should probably rename this for things in the `apps` directory.

LinuxJedi (approved):
Both good catches!

JacobBarthelmeh:
Added in a test a case and it looks like the issue is not completely resolved yet. Am investigating it more.

PR #777, opened 2025-02-18
sanity check and free temporary strings on failure

JacobBarthelmeh:


src/internal.c
```diff
@@ -7886,14 +7886,15 @@ static int DoUserAuthInfoRequest(WOLFSSH* ssh, byte* buf, word32 len,
...
+            WMEMSET(prompts, '\0', sizeof(char*) * promptSz);
             echo = (byte*)WMALLOC(sizeof(byte) * promptSz, ssh->ctx->heap,
                                   DYNTYPE_BUFFER);
         }
 
         if (!echo) {
             ret = WS_MEMORY_E;
         } else {
-            WMEMSET(prompts, '\0', sizeof(char*) * promptSz);
+            WMEMSET(echo, 0, sizeof(byte) * promptSz);
```
LinuxJedi, line 7897: I can't think of a case where these memsets would be needed, but I suspect a static analyzer triggered it.

src/internal.c
```diff
@@ -7923,6 +7924,11 @@ static int DoUserAuthInfoRequest(WOLFSSH* ssh, byte* buf, word32 len,
         }
         WFREE(prompts, ssh->ctx->heap, DYNTYPE_BUFFER);
         WFREE(echo, ssh->ctx->heap, DYNTYPE_BUFFER);
+
+        /* free strings in fail case */
+        WFREE(authName, ssh->ctx->heap,  DYNTYPE_STRING);
+        WFREE(authInstruction, ssh->ctx->heap,  DYNTYPE_STRING);
+        WFREE(language, ssh->ctx->heap,  DYNTYPE_STRING);
```
LinuxJedi, line 7931: Oh arse... Good call, thanks!

LinuxJedi (approved):
Good catches, many thanks!

douzzer (approved):

PR #784, opened 2025-02-24
add initial SFTP benchmark collection scripts

JacobBarthelmeh:


.github/workflows/sftp-benchmark.yml
```diff
@@ -0,0 +1,184 @@
...
+          lookup-only: true
+
+      - name: Checkout, build, and install wolfssl
+        if: steps.cache-wolfssl.outputs.cache-hit != 'true'
+        uses: wolfSSL/actions-build-autotools-project@v1
+        with:
+          repository: wolfssl/wolfssl
+          ref: ${{ matrix.wolfssl }}
+          path: wolfssl
+          configure: --enable-all
```
LinuxJedi, line 45: This is going to be slow as molasses without some kind of optimisation enabled, would not be a fair test against OpenSSH which does optimise. I recommend `--enable-intelasm`

PR #789, opened 2025-03-10
Curve25519 Compat with LibSSH

ejohnstown:
1. Add curve25519-sha256@libssh.org back to the list of key exchange algorithms.
2. Treat curve25519-sha256@libssh.org as an alias for curve25519-sha256.
3. Remove the reference to ID_ECDH_SHA2_ED25519 from the list of algo IDs. It isn't actually a thing.

(ZD19526)


anhu (approved):
LGTM!!

JacobBarthelmeh (approved):

PR #790, opened 2025-03-18
microchip example and filesystem port

JacobBarthelmeh:


src/wolfsftp.c
```diff
@@ -4846,7 +4922,80 @@ int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
...
+    /* check for read only */
+    if ((stats->fattrib & SYS_FS_ATTR_RDO) & SYS_FS_ATTR_MASK) {
+        atr->per |= 0x124; /* octal 444 */
+    }
+    else {
+        atr->per |= 0x1ED; /* octal 755 */
+    }
+
+    /* volume label */
+    //stats.fattrib == SYS_FS_ATTR_VOL
```
dgarske, line 4953: Are these TODO or should they be removed?

dgarske (changes requested):

src/wolfsftp.c
JacobBarthelmeh: Can be removed. I had gotten all of the file state struct values from documentation and commented out the ones was not using (with a note of SYS_FS_ATTR_VOL in case needing it).

dgarske (approved):

ide/mplabx/include.am
```diff
@@ -0,0 +1,8 @@
+# vim:ft=automake
+# All paths should be given relative to the root
+
+EXTRA_DIST+= ide/mplabx/README.md
```
dgarske, line 4: Is it okay that some of the files are missing here?

dgarske (changes requested):

ide/mplabx/include.am
JacobBarthelmeh: I was hesitant because of how large the auto generated MPLABX files are. Adding them though makes sense. I'll update it.

dgarske (approved):

PR #795, opened 2025-04-04
adjust highwater check location to avoid masking return value 

JacobBarthelmeh:
ZD19648


JacobBarthelmeh:
Working on a unit test for this.

SparkiDev (approved):

PR #800, opened 2025-04-29
Make Keyboard Interactive a compile time option

LinuxJedi:
Now an off-by-default compile-time option. This saves resources and confusion if you are not expecting it.

ZD #19704


LinuxJedi:
FATFS test failing due to GitHub now being blocked by the only FATFS mirror.

ejohnstown (approved):

PR #801, opened 2025-05-02
KeyboardInteractive Fix

ejohnstown:
1. Snip out some extraneous states from the server handshake tracking for accept.
2. Change sending the keyboard interactive info response to a reaction to a request.


LinuxJedi (approved):
Good call!

PR #804, opened 2025-05-16
Soft Disable AES-CBC

ejohnstown:
1. By default, soft disable AES-CBC. It isn't offered as a default encrypt algorithm, but may be set at runtime.
2. Add guard where AES-CBC can be added back as a default.
3. Add option to example client to run it with a custom encrypt algorithm list.
4. In the client, add macro to add items to the arg lists while checking the number of items in the list.
(ZD 19606)

### DIFF
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -126,7 +126,8 @@ static void ShowUsage(void)
     printf(" -X            Ignore IP checks on peer vs peer certificate\n");
 #endif
     printf(" -E            List all possible algos\n");
-    printf(" -k            set the list of key algos to use\n");
+    printf(" -k            set the list of key algos\n");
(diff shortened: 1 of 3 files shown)


douzzer (approved):

douzzer:
This PR appears to be unmergeable due to a bug in githib (see https://github.com/orgs/community/discussions/144455)

A trick that often works is to close and reopen a PR.  I've used that to clear a rebase conflict (the instant-dismissed-review syndrome).  Doesn't work here.

LinuxJedi:
No problem merging for me, once I resolved the open conversation I could see.

douzzer:
Thanks @LinuxJedi -- not sure if I was missing something or if something changed, but all's well that ends well.

PR #822, opened 2025-07-21
Zendesk Coverity changes

rlm2002:
Adds NULL checks to clear Coverity issues reported in ZD#20212.

- Dereference before NULL check
    - check `ssh->ctx != NULL` before dereferencing in `GetAllowedAuth()` (wolfSSH Project **572900**)
- Dereference after NULL check
    - Add additional NULL checks as `SendUserAuthKeyboardRequest()` does not return immediately if `ssh` or `authData` are NULL (wolfSSH Project **CI 572871** and **CI 572874**)
    - `AlgoListSz()` returns 0 if param is NULL (wolfSSH Project **572893**)

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -13321,7 +13321,8 @@ static int BuildUserAuthRequestKeyboard(WOLFSSH* ssh, byte* output, word32* idx,
             begin += LENGTH_SZ;
             WMEMCPY(output + begin, authData->sf.keyboard.promptName, slen);
             begin += slen;
-        } else {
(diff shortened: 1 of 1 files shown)


dgarske (approved):

ejohnstown (changes requested):
Issues 572871 and 572874 are still present. I think it found two issues each in the areas you fixed, but you only fixed one of the issues.

ejohnstown (approved):

PR #829, opened 2025-08-25
Fix handling of WOLFSSH_USERAUTH_REJECTED result from userAuthCb

tjko:
I noticed that if compile option **NO_FAILURE_ON_RJECTED** is not used,
then _WOLFSSH_USERAUTH_REJECTED_ response from user authentication 
callback (userAuthCb) do not appear to be handled properly (?)  As this seems to prevent SSH server
(userAuthCb) from rejecting client connections.

If **NO_FAILURE_ON_REJECTED** is defined, then everything works ok, as "DoUserAuth" functions (DoUserAuthInforResponse(), DouserAuthRequestPassword(), DoUserAuthRequestPublicKey()) 
end up returning correct return value (**WS_USER_AUTH_E**).

But when _NO_FAILURE_ON_REJECT_ is **not** defined, that causes _authFailure_ getting set:
```
            else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                WLOG(WS_LOG_DEBUG, "DUARKB: keyboard rejected");
                #ifndef NO_FAILURE_ON_REJECTED
                    authFailure = 1;
                #endif
                ret = WS_USER_AUTH_E;
            }
```

Which in turn results in the return value getting overridden little bit later by function call to _SendUserAuthFailure()_:

```
    if (authFailure || partialSuccess) {
        ret = SendUserAuthFailure(ssh, partialSuccess);
    }
```

I noticed that _DoUserAuthRequestNone()_ doesn't seem to have this problem as it handles this differently, so I updated the other "DoUserAuth..." functions to work same way, and now connections get rejected also when **NO_FAILURE_ON_REJECT** is not used.

This patch updates the other functions to follow same logic as  _DoUserAuthRequestNonte()_ already does. Not sure if this is necessary "correct" solution, but seems to solve my problem...

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -6517,9 +6517,12 @@ static int DoUserAuthInfoResponse(WOLFSSH* ssh,
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                 WLOG(WS_LOG_DEBUG, "DUARKB: keyboard rejected");
                 #ifndef NO_FAILURE_ON_REJECTED
-                    authFailure = 1;
-                #endif
+                ret = SendUserAuthFailure(ssh, 0);
+                if (ret == WS_SUCCESS)
+                    ret = WS_USER_AUTH_E;
+                #else
                 ret = WS_USER_AUTH_E;
+                #endif
             }
             else if (ret == WOLFSSH_USERAUTH_WOULD_BLOCK) {
                 WLOG(WS_LOG_DEBUG, "DUARKB: keyboard callback would block");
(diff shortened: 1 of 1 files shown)


dgarske:
> @dgarske, seems like there is/was some (macos-latest) workflow issue? As there seems to have been same failures on other PRs (like this one: https://github.com/wolfSSL/wolfssh/actions/runs/17214562015).
> 
> Btw, I don't seem to have permission to "Re-run jobs", so what would be best way to re-trigger workflows (once macos flows work again)?

@ejohnstown or @JacobBarthelmeh ?

tjko:
@JacobBarthelmeh, thanks. That worked great.

PR #832, opened 2025-09-16
do not treat shell as interactive untill pty-req recieved

JacobBarthelmeh:
ZD20321

WinSCP fails to connect with "SCP" and "Default" shell used. This is because WinSCP is opening a shell subsystem but not using pty-req and expecting the server to create a non-interactive shell. wolfSSHd previously would send interactive terminal characters which confused WinSCP leading to it hanging after the SSH handshake was completed.


JacobBarthelmeh:
Great feedback. I refactored the flag to be at the Channel level and adjust the checks on the flag in wolfsshd.c

PR #836, opened 2025-09-25
log: make WLOG macro a noop if logging disabled

AxlLind:
even if logging is disabled, the strings are still included in the binary. This is because the compiler does not see that wolfSSH_Log() is compiled as a noop.

By making WLOG a noop if wolfSSH_Log is we reduce the compiled so-file by ~30k (about 25%).

Compiling on my machine for linux x86_64 `libwolfssh.so.7.10.2` goes from `197k` to `144k`.

---

One downside is potential warnings about variables which are only used in WLOG invocations. This only happened in one place in the code base, so I would argue it's worth it.


AxlLind:
I see now that `log.h` is included in installations. Are users meant to use `WLOG` from their ssh clients/servers? If so they might not have the same defines when compiling.

PR #854, opened 2025-12-06
fix uninitialized authData

miyazakh:


divinity76:
Never understood why people bother initializing stuff like
```
WS_UserAuthData authData;
memset(authData, 0, sizeof(authData));
```
what's wrong with C's brace initializer
```
WS_UserAuthData authData = { 0 };
```
? why not use it? it was introduced in C89, it's been in the C standard for over 35 years, it's not like C compilers don't support it

PR #855, opened 2025-12-09
Out Of Order Messaging Checking

ejohnstown:
1. Improved the checking the appropriateness of messages depending on the state.
2. Added explicit checking for specific messages when appropriate.
3. Added a regression test that checks for specific failure cases.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -5989,8 +5989,9 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     }
 
     if (ret == WS_SUCCESS) {
-        ret = SendNewKeys(ssh);
         ssh->handshake->expectMsgId = MSGID_NEWKEYS;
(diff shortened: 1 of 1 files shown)


src/internal.c
```diff
@@ -12613,6 +12732,11 @@ int SendKexDhGexRequest(WOLFSSH* ssh)
     if (ret == WS_SUCCESS)
         ret = wolfSSH_SendPacket(ssh);
 
+    if (ret == WS_SUCCESS) {
```
JacobBarthelmeh: Should we handle WANT_WRITE cases (and other error cases) here and the other couple locations where expectMsgId is set only if wolfSSH_SendPacket returns success. This differs from the pattern after SendNewKeys where it is unconditionally set after the packet has been created and attempted to be sent.

JacobBarthelmeh (changes requested):

PR #900, opened 2026-03-25
Use of Windows certificate store for authentication

JacobBarthelmeh:


src/internal.c
```diff
@@ -16155,6 +16584,34 @@ static int PrepareUserAuthRequestRsaCert(WOLFSSH* ssh, word32* payloadSz,
                     authData->sf.publicKey.publicKeySz);
         else
         #endif /* WOLFSSH_AGENT */
+#ifdef WOLFSSH_WINDOWS_CERT_STORE
+        /* Note: already inside #ifdef WOLFSSH_CERTS */
+        if (authData->sf.publicKey.privateKey == NULL) {
+            /* Cert store: decode public key from the stored certificate */
+            word32 ki;
+            ret = WS_BAD_ARGUMENT;
+            for (ki = 0; ki < ssh->ctx->privateKeyCount; ki++) {
```
ejohnstown: The client userauth builders select the cert-store key by taking the first slot with useCertStore set, ignoring the algorithm of the auth attempt:

```
 for (ki = 0; ki < ssh->ctx->privateKeyCount; ki++) {
     if (ssh->ctx->privateKey[ki].useCertStore &&
         ssh->ctx->privateKey[ki].cert != NULL) {
         ... ExtractPubKeyDerFromCert(...) ...
         ret = wc_RsaPublicKeyDecode(...);
         break;
     }
 }

```

If a client configures both an RSA and an ECC cert-store key, this RSA path (and the ECC counterparts) can grab the wrong slot: wc_RsaPublicKeyDecode on an ECC cert fails, or the signer/key pair is mismatched. Single-key configs (the normal case) are unaffected, so real-world impact is low, but the selection should match the slot whose publicKeyFmt corresponds to the algorithm being built. The server side (SendKexGetSigningKey) already does this correctly via keyIdx. Same pattern at ~L16749, ~L17110, ~L17269 — a shared FindCertStoreKey(ctx, keyId) helper would fix all four.

src/internal.c
```diff
@@ -16725,24 +17254,89 @@ static int BuildUserAuthRequestEccCert(WOLFSSH* ssh,
...
+                    break;
+                }
+            }
+            if (pvtKey != NULL) {
+                ret = SignWithCertStoreKey(ssh, pvtKey,
+                        digest, digestSz, hashId, sig, &sigSz);
+                if (ret == WS_SUCCESS) {
+                    /* NCryptSignHash ECDSA output is raw r||s, each
+                     * component is half the total signature size. */
+                    word32 halfSz;
```
ejohnstown: Maintainability: this raw r||s split + leading-zero trim block is copy-pasted from the identical logic in SignHEcdsa. The bounds checks are correct in both copies, but duplicated code drifts. Suggest factoring the NCRYPT ECDSA r||s handling into a helper next to SignWithCertStoreKey (e.g. CertStoreEccSigToRs(...)) and calling it from both sites. Pairs naturally with the FindCertStoreKey helper suggested on the RSA path.

ejohnstown (changes requested):

ejohnstown (dismissed):

src/ssh.c
```diff
@@ -2676,6 +2687,334 @@ int wolfSSH_CTX_AddRootCert_buffer(WOLFSSH_CTX* ctx,
...
+        ctx->privateKey[keyIdx].key = NULL;
+        ctx->privateKey[keyIdx].keySz = 0;
+    }
+    if (ctx->privateKey[keyIdx].cert != NULL) {
+        WFREE(ctx->privateKey[keyIdx].cert, heap, DYNTYPE_CERT);
+        ctx->privateKey[keyIdx].cert = NULL;
+        ctx->privateKey[keyIdx].certSz = 0;
+    }
+
+    /* Set up the private key structure */
```
aidangarske: **🟠 [High] Store-backed host keys never negotiate RFC6187 certificates**
*🚫 BLOCK*  `bug`

The new store loader copies the certificate into `pvtKey->cert`, but registers the slot as the plain base key type with `ctx->privateKey[keyIdx].publicKeyFmt = keyId`. `RefreshPublicKeyAlgo()` therefore advertises plain `ssh-rsa`/`ecdsa-*` algorithms, and `SendKexDhReply()` only sends RFC6187 certificate K_S when `pubKeyFmtId` is one of the `ID_X509V3_*` values. The new wolfsshd `HostKeyStore` path does not allow a separate `HostCertificate`, and the workflow passes `-A` as if the host certificate is sent from the store, so store-backed host keys are exposed as raw public keys rather than the X.509 certificate from the store.

**Suggestion:**
```suggestion
For wolfsshd host keys loaded from the Windows certificate store, register an RFC6187 certificate key id, for example `CertTypeForId(keyId)`, or add an explicit API/flag to choose raw-key versus certificate presentation. Ensure the advertised host-key algorithm is `ID_X509V3_*` and that `BuildRFC6187Info()` is used for K_S when `HostKeyStore` is configured.
```

src/certman.c
```diff
@@ -85,6 +94,29 @@ struct WOLFSSH_CERTMAN {
 };
 
 
+/* used to import an external cert manager, frees and replaces existing manager
+ * returns WS_SUCCESS on success
+ */
+int wolfSSH_SetCertManager(WOLFSSH_CTX* ctx, WOLFSSL_CERT_MANAGER* cm)
```
aidangarske, line 132: **🟠 [High] Imported cert manager can dangle when up_ref fails**
*🚫 BLOCK*  `bug`

The new helper imports a `WOLFSSL_CERT_MANAGER*` from a temporary `WOLFSSL_CTX`, but it ignores the return value from `wolfSSL_CertManager_up_ref(cm)`. The wolfSSL API returns failure when reference-count locking fails. If that happens, this function still frees wolfSSH's existing manager and stores `cm`; the caller then frees the temporary `WOLFSSL_CTX`, which can free the same manager and leave `ctx->certMan->cm` dangling.

**Suggestion:**
```suggestion
if (wolfSSL_CertManager_up_ref(cm) != WOLFSSL_SUCCESS) {
    return WS_FATAL_ERROR;
}
if (ctx->certMan->cm != NULL) {
    wolfSSL_CertManagerFree(ctx->certMan->cm);
}
ctx->certMan->cm = cm;
```

apps/wolfsshd/wolfsshd.c
```diff
@@ -475,6 +562,72 @@ static int SetupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
 #endif /* WOLFSSH_OSSH_CERTS || WOLFSSH_CERTS */
 
 #ifdef WOLFSSH_CERTS
+    /* check if loading in system and/or user CA certs */
+    #ifdef WOLFSSL_SYS_CA_CERTS
+    if (ret == WS_SUCCESS && (wolfSSHD_ConfigGetSystemCA(conf)
```
aidangarske: **🟠 [High] System CA store can become a Windows SSH user-auth trust root without identity binding**
*💡 SUGGEST*

The PR adds `wolfSSH_TrustedSystemCAKeys yes`, which loads the OS system CA bundle into wolfSSH's certificate manager for SSH user-certificate authentication. Data Flow: sshd config `wolfSSH_TrustedSystemCAKeys yes` -> `wolfSSHD_ConfigSetSystemCA()` sets `useSystemCA` -> `SetupCTX()` calls `wolfSSL_CTX_load_system_CA_certs()` and installs that cert manager with `wolfSSH_SetCertManager()` -> a remote client cert chaining to any OS-trusted CA is accepted by wolfSSH's cert verification -> `RequestAuthentication()` reaches the Windows certificate-auth branch when no explicit `AuthorizedKeysFile` is set -> `SetupUserTokenWin()` creates an S4U token for the attacker-supplied SSH username. On Windows builds without `WOLFSSL_FPKI`, the unchanged username-to-UPN binding check is not compiled, and `SetupUserTokenWin()` does not inspect `pubKeyCtx` or `usrCaKeysFile`; it only logs on the requested username. This PR exposes that pre-existing Windows token path to a much broader trust source than an explicit `TrustedUserCAKeys` file. Attacker controls: SSH username and a client certificate/private key chaining to a public/system-trusted CA. Trigger: run wolfsshd on Windows with `wolfSSH_TrustedSystemCAKeys yes`, public key auth enabled, and no explicit authorized-keys file; connect as a local user with any accepted system-CA client cert. Impact: authentication as arbitrary local users that can be S4U logged on. Counter-argument considered: this requires a non-default admin configuration, and `WOLFSSL_FPKI` builds perform a UPN-vs-username check before this branch. The finding still survives for supported Windows/X.509 builds where FPKI is absent; the new directive makes a single config line trust the public OS root store for SSH user authentication.

**Recommendation:** Do not allow OS/root stores to be used as SSH user-auth CA roots unless the certificate identity is bound to the requested account. Fail configuration on Windows when store-backed user CAs are enabled without FPKI or an explicit per-user authorized key/cert mapping, and make `SetupUserTokenWin()` validate the certificate subject/UPN/SID before S4U logon. Prefer a dedicated SSH user-CA store/file over `wolfSSL_CTX_load_system_CA_certs()` for user authentication.

apps/wolfsshd/wolfsshd.c
```diff
@@ -475,6 +562,72 @@ static int SetupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
...
+                    "[SSHD] Issue copying over system CAs");
+                ret = WS_FATAL_ERROR;
+            }
+        }
+
+        if (sslCtx != NULL) {
+            wolfSSL_CTX_free(sslCtx);
+        }
+    }
+    #else
```
aidangarske, line 1276: **🟠 [High] CA-store config directives are accepted but ignored when unsupported**
*🚫 BLOCK*  `bug`

The PR adds `wolfSSH_TrustedSystemCAKeys` and `wolfSSH_TrustedUserCaStore` config directives, but when wolfSSL is built without `WOLFSSL_SYS_CA_CERTS`, `SetupCTX()` only logs an error and continues with `ret == WS_SUCCESS`. That lets wolfsshd start after accepting a trust-store directive that it did not apply. This is introduced by the new config path and makes deployments fail later during certificate authentication rather than rejecting the unsupported configuration at startup.

**Suggestion:**
```suggestion
if (ret == WS_SUCCESS && (wolfSSHD_ConfigGetSystemCA(conf)
                          || wolfSSHD_ConfigGetUserCAStore(conf))) {
    wolfSSH_Log(WS_LOG_ERROR,
        "[SSHD] wolfSSH_TrustedSystemCAKeys/wolfSSH_TrustedUserCaStore set "
        "but wolfSSL was built without WOLFSSL_SYS_CA_CERTS.");
    ret = WS_NOT_COMPILED;
}
```

src/certman.c
```diff
@@ -589,4 +621,101 @@ static int CheckProfile(DecodedCert* cert, int profile)
 }
 #endif /* WOLFSSH_NO_FPKI */
 
+
+#ifdef WOLFSSH_WINDOWS_CERT_STORE
+/* Parse a cert store spec string "store:subject:flags" into wide-string
+ * components.  Allocates wStoreName and wSubjectName via WMALLOC; caller
+ * must WFREE them.  dwFlags is set to the parsed flags value.
+ * Returns WS_SUCCESS on success. */
+int wolfSSH_ParseCertStoreSpec(const char* spec,
```
aidangarske, line 1071: **🟡 [Medium] Core cert-store parsing and state transitions lack unit coverage**
*💡 SUGGEST*  `test`

The PR adds parser behavior, key-slot replacement behavior, and cert-store signing paths, but the only new coverage is a Windows GitHub Actions workflow. `grep` found no tests under `tests/` for `wolfSSH_ParseCertStoreSpec`, `wolfSSH_CTX_UsePrivateKey_fromStore`, `wolfSSH_SetCertManager`, or the cert-store ECDSA/RSA conversion helpers. The workflow is useful integration coverage, but it will not exercise parser edge cases or failure paths such as missing fields, invalid flags, allocation failure, replacement of an existing slot, or `wolfSSL_CertManager_up_ref()` failure.

**Recommendation:** Add targeted unit tests for cert-store spec parsing, unsupported/invalid flags, replacing existing private-key slots, and failed cert-manager reference acquisition. Keep the Windows workflow as integration coverage.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Store-backed host keys never negotiate RFC6187 certificates** — `src/ssh.c:2896-3012`
- [High] **Imported cert manager can dangle when up_ref fails** — `src/certman.c:100-115`
- [High] **System CA store can become a Windows SSH user-auth trust root without identity binding** — `apps/wolfsshd/wolfsshd.c:567-607`
- [High] **CA-store config directives are accepted but ignored when unsupported** — `apps/wolfsshd/wolfsshd.c:618-627`
- [Medium] **Core cert-store parsing and state transitions lack unit coverage** — `src/certman.c:630-717`

*Review generated by Skoll.*

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 13 total — 13 posted, 0 skipped

### Posted findings
- [High] [review+review-security+bugs] **Unknown or undecodable ECC curves are silently registered as P-256** — `src/ssh.c:3106-3121`
- [High] [bugs+review] **Certificate-store keys can advertise algorithms excluded from the build** — `src/ssh.c:3075-3105`
- [High] [review+bugs] **Switching between TPM and certificate-store keys leaves both backends active** — `src/ssh.c:2981-3017`
- [High] [review+review-security+bugs] **Windows certificate-store libraries are missing from most Visual Studio configurations** — `ide/winvs/wolfsshd/wolfsshd.vcxproj:245-459`
- [High] [review-security] **Match blocks cannot restrict the new CA trust sources** — `apps/wolfsshd/configuration.c:1348-1372`
- [High] [review] **Certificate CN matching conflicts with case-insensitive Windows account names** — `apps/wolfsshd/auth.c:1792-1797`
- [Medium] [review+review-security] **Unsupported certificate-store provider is accepted and then silently ignored** — `apps/wolfsshd/wolfsshd.c:383-429`
- [Medium] [review-security] **Raw numeric location flags can trigger destructive certificate-store operations** — `src/certman.c:724-739`
- [Medium] [review] **Cert-store authentication leaves the shared client private-key pointer in an invalid reusable state** — `examples/client/common.c:1261-1263`
- [Medium] [review] **The workflow never exercises the X.509 host-key slot** — `.github/workflows/windows-cert-store-test.yml:522-545`
- [Medium] [bugs] **HostKeyStoreSubject cannot contain whitespace** — `apps/wolfsshd/configuration.c:1384-1389`
- [Low] [review-security] **CN lookup selects an arbitrary certificate when renewals share a subject** — `src/ssh.c:2867-2897`
- [Low] [review-security] **Configure accepts Cygwin even though the headers reject it** — `configure.ac:290-296`

wolfssh/test.h
```diff
@@ -1158,30 +1158,45 @@ static INLINE void build_addr_ipv6(struct sockaddr_in6* addr, const char* peer,
...
+    WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD,
+    WS_HEX_BAD, WS_HEX_BAD,
+    WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD, WS_HEX_BAD,
+    WS_HEX_BAD, WS_HEX_BAD,
+    WS_HEX_BAD, WS_HEX_BAD,  /* G - ` */
     10, 11, 12, 13, 14, 15   /* lower case a-f */
 };  /* A starts at 0x41 not 0x3A */
 
+#undef Base16_Decode
+#define Base16_Decode WS_Base16_Decode
```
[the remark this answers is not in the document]
JacobBarthelmeh: Treating this as a won't fix. This is for the test cases and not something included when linking with the main wolfssh library.

apps/wolfsshd/wolfsshd.c
```diff
@@ -359,6 +471,384 @@ static void CleanupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
...
+    }
+    if (!IsElevationProtectedHive(dwFlags)) {
+        wolfSSH_Log(WS_LOG_WARN,
+            "[SSHD] wolfSSH_WinUserDwFlags selects a store hive that its own "
+            "account can write to without elevation.");
+        wolfSSH_Log(WS_LOG_WARN,
+            "[SSHD] LOCAL_MACHINE is the safer location for trust anchors.");
+    }
+
+#ifdef WOLFSSH_NO_FPKI
```
aidangarske, line 697: **Default Windows builds accept certificates restricted to non-SSH purposes** · _Security_

BEFORE this PR, wolfsshd had no Windows user-CA-store trust-source path. AFTER, `LoadUserCACertsFromStore()` imports CA certificates from the configured store and explicitly permits authentication without FPKI, merely warning that TLS-server, S/MIME, or code-signing certificates may be accepted. An attacker-supplied chain reaches `wolfSSH_CERTMAN_VerifyCerts_buffer()`, while the `WOLFSSH_NO_FPKI` path omits application-purpose, EKU, and key-usage enforcement. The new Windows authentication branch then checks only that the leaf subject CN matches the requested username before creating the user's token. Consequently, anyone holding a valid certificate and private key issued by a configured CA can authenticate as the matching Windows account even when that certificate is explicitly restricted to serverAuth, S/MIME, code signing, or lacks digitalSignature usage. This affects the supplied Visual Studio configuration, which defines WOLFSSH_NO_FPKI. Existing tests use a certificate without an explicit incompatible EKU and therefore do not cover this bypass (RFC 6187 section 2.2 requires digitalSignature when KeyUsage is present and restricts EKU-bearing certificates to their indicated purposes).

**Fix:** Enforce SSH-appropriate leaf usage independently of FPKI: require digitalSignature whenever KeyUsage is present and, whenever EKU is present, require id-kp-secureShellClient or an explicitly configured equivalent. Retain FPKI as an additional profile check. Add Windows integration tests proving that valid matching-CN chains containing serverAuth-only or codeSigning-only leaves are rejected.

examples/echoserver/echoserver.c
```diff
@@ -3133,11 +3146,47 @@ static void ShowUsage(void)
...
+/* Detects whether argv or the environment requests a host key from the
+ * Windows certificate store, without doing the full option parse that
+ * echoserver_test() does later. Used to decide whether the root directory
+ * search for PEM key files should be skipped. */
+static int EchoserverUsingCertStore(int argc, char** argv)
+{
+    int i;
+    const char* spec;
+
+    for (i = 1; i < argc; i++) {
```
aidangarske: **Parse -W as an option before skipping the root search** · _bug_

The PR's preliminary scan treats every argument beginning with `-W` as the certificate-store option. This includes values consumed by preceding options and tokens following `--`. For example, a valid option value such as `-a -WolfCA.pem` makes `EchoserverUsingCertStore()` return true even though getopt will not process a `-W` option. `wolfSSH_Echoserver()` consequently skips `ChangeToWolfSshRoot()`, after which relative certificate and default host-key paths are resolved from an unexpected directory. The SFTP client avoids this ambiguity by using `mygetopt()` for its preliminary scan.

**Fix:** Use the same option string as the main parser, scan with `mygetopt()`, and restore `myoptind` afterward. Add coverage for an argument value beginning with `-W` and for `--`.

apps/wolfsshd/auth.c
```diff
@@ -2716,28 +2821,44 @@ static int RequestAuthentication(WS_UserAuthData* authData,
...
                         }
                         current = current->next;
                     }
-
-                    /* a UPN matched but no realm policy is set; warn per auth
-                     * attempt so the opt-in gap is visible, no shared state */
-                    if (upnRealmUnchecked) {
-                        wolfSSH_Log(WS_LOG_WARN, "[SSHD] AuthorizedUPNDomains "
-                            "not set; certificate UPN domain is not checked");
+                #else
```
aidangarske, line 2863: **Exercise the non-FPKI Windows CN-binding path** · _test_

The PR adds a distinct authentication path that binds certificates to Windows accounts through the subject CN when WOLFSSL_FPKI is absent. The new Windows workflow enables OPENSSL_ALL, which enables FPKI, and explicitly notes that only the UPN branch is exercised. Consequently, the newly introduced CN extraction, length comparison, and case-insensitive account matching can regress without any functional test detecting it.

**Fix:** Add a Windows build configuration without OPENSSL_ALL/WOLFSSL_FPKI and test a matching CN, a case-differing matching account name, and a mismatched CN.

tests/unit.c
```diff
@@ -17896,6 +18313,133 @@ static int test_ParseECCPubKey(void)
...
+        if (ret != WS_INVALID_PRIME_CURVE) {
+            printf("ParseECCPubKeyCert: unmapped id ret %d\n", ret);
+            result = -7;
+        }
+    }
+
+    /* matching curve accepted; a failure here means the chain itself did
+     * not verify under this build's profile, so skip the curve vectors
+     * rather than fail on unrelated policy */
+    if (result == 0) {
```
aidangarske, line 18620: **Do not turn arbitrary certificate-parser failures into skips** · _test_

The new control vector converts every failure from the matching P-256 ParseECCPubKeyCert() path into a positive skip result. The test harness then ignores all positive results. A regression in the newly changed certificate-chain parser, curve binding, or root verification can therefore make this test print SKIPPED while CI remains green, preventing the subsequent mismatch vector from running as well.

**Fix:** Skip only for an explicitly identified unsupported build profile or unavailable fixture. Treat unexpected parser and verification errors as test failures, or provide a fixture valid under every profile in which the test is compiled.

apps/wolfsshd/wolfsshd.c
```diff
@@ -359,6 +471,384 @@ static void CleanupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
...
+            wStoreName);
+    if (hStore == NULL) {
+        wolfSSH_Log(WS_LOG_ERROR,
+            "[SSHD] Unable to open user CA cert store '%.48s', error %lu",
+            storeNameStr, (unsigned long)GetLastError());
+        WFREE(wStoreName, heap, DYNTYPE_SSHD);
+        return WS_FATAL_ERROR;
+    }
+
+    /* Passing the previous context frees it and advances the enumeration. */
```
aidangarske: **Distinguish certificate-store enumeration failure from end of store** · _bug_

The PR's new loader treats any NULL result from CertEnumCertificatesInStore() as successful completion. Microsoft's API contract requires checking GetLastError() because NULL represents both normal exhaustion and enumeration errors. If an error occurs after at least one CA was loaded, loaded > 0 makes this function return success and report the partial trust-store load as complete, silently omitting the remaining configured authorities.

**Fix:** Capture GetLastError() immediately after enumeration returns NULL. Accept only the documented end-of-enumeration statuses such as CRYPT_E_NOT_FOUND and, where applicable, ERROR_NO_MORE_FILES; otherwise fail the load and avoid reporting partial success.

aidangarske (changes requested):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`

### Posted findings
- [High] [review-security] **Default Windows builds accept certificates restricted to non-SSH purposes** — `apps/wolfsshd/wolfsshd.c:696-702`
- [Medium] [review] **Parse -W as an option before skipping the root search** — `examples/echoserver/echoserver.c:3178-3182`
- [Medium] [review] **Exercise the non-FPKI Windows CN-binding path** — `apps/wolfsshd/auth.c:2836-2856`
- [Medium] [review] **Do not turn arbitrary certificate-parser failures into skips** — `tests/unit.c:18409-18416`
- [Medium] [review] **Distinguish certificate-store enumeration failure from end of store** — `apps/wolfsshd/wolfsshd.c:739-744`

apps/wolfsshd/wolfsshd.c
JacobBarthelmeh: I believe that this issue is pre-existing in wolfSSH master branch. The fix widens the scope some of this PR (will add more lines to it) but it is a good fix to have for all code paths making use of X509 certs without FPKI enabled.

examples/echoserver/echoserver.c
JacobBarthelmeh: Had seen this previously, was trying to avoid having the options set as a macro. It seemed a bit clunky in the examples to do it that way and changing the location of ChangeToWolfSshRoot() teases out a Mac OS / clang bug that leads to a segfault. I'll update the echoserver though to have the macro set for options.

apps/wolfsshd/auth.c
JacobBarthelmeh: Will expand the test coverage.

apps/wolfsshd/wolfsshd.c
JacobBarthelmeh: This wolfSSL PR (https://github.com/wolfSSL/wolfssl/pull/11361) adds some of the EKU logic needed for a cleaner bit check instead of the added manual parsing of the ASN1 syntax used for now. Later wolfSSH will be able to conditionally remove the parsing based on version of wolfSSL linked with.

apps/wolfsshd/wolfsshd.c
JacobBarthelmeh: Fixed to check GetLastError()

tests/unit.c
JacobBarthelmeh: Now only WS_CERT_PROFILE_E should skip.

JacobBarthelmeh:
Rebasing now for resolving the merge conflict with configure.ac. Current CI though is green, don't expect a configure.ac change to break that.

aidangarske (dismissed):

apps/wolfsshd/wolfsshd.c
```diff
@@ -3232,6 +4200,24 @@ static int StartSSHD(int argc, char** argv)
...
 
+    if (logFile == NULL) {
+        logFile = stderr;
+    }
+#ifdef _WIN32
+    /* The early -D detection (wide-string comparison of cmdArgs before
+     * conversion) may have set ServiceDebugCb even when -D was supplied.
+     * Now that mygetopt has been processed, restore the file-based
+     * callback in any case where output should go to logFile:
+     *   - isDaemon==0  → running interactively, logs to logFile (stderr)
```
ejohnstown: Replace `→` with `->` or a word.

apps/wolfsshd/wolfsshd.c
```diff
@@ -3232,6 +4200,24 @@ static int StartSSHD(int argc, char** argv)
...
+     *   - isDaemon==1 but -E was used → logs to the specified file
```
ejohnstown: Replace → with -> or a word.

.github/workflows/windows-cert-store-test.yml
```diff
@@ -0,0 +1,1716 @@
...
+        path: wolfssl
+
+    - uses: actions/checkout@v4
+      with:
+        path: wolfssh
+
+    - name: Add MSBuild to PATH
+      uses: microsoft/setup-msbuild@v1
+
+    - name: Restore wolfSSL NuGet packages
```
ejohnstown: Is NuGet actually used?

ejohnstown (changes requested):
These review items shouldn't affect the compile. Be careful using UTF-8 characters in your comments. Works well.

apps/wolfsshd/wolfsshd.c
JacobBarthelmeh: sorry about that, fixed both comment lines

.github/workflows/windows-cert-store-test.yml
JacobBarthelmeh: Removed. This was left over from an earlier iteration of the tests, it's no longer needed.

ejohnstown (approved):

PR #943, opened 2026-04-21
Add unit test for ChannelPutData

yosuke-wolfssl:
This PR adds a unit test that creates a channel with a small windowSz (e.g., 100 bytes), then calls ChannelPutData with a buffer larger than the window. Verify that WS_FATAL_ERROR is returned. Also verify that after consuming window with valid data, exceeding the remaining window is rejected.


yosuke-wolfssl:
Hello @ejohnstown ,
I solved the merge conflict now.
Could you review this please ?

ejohnstown (approved):

PR #952, opened 2026-04-24
Add independent ciper and MAC algorithms negotiation for each direction

yosuke-wolfssl:
This PR adds independent ciper and MAC algorithms negotiation for each direction to comply with RFC 4253 section 7.1.

- Add peerEncryptId, peerMacId, peerAeadMode, peerBlockSz, peerMacSz into HandshakeInfo.
- Fix DoKexInit() so that this matches each algorithm independently and store the results into each member in HandshakeInfo. 
- Fix DoNewKeys() so that this updates peer algorithm info like ssh->peerEncryptId with ssh->handshake->peerEncryptId and so on.

Also, new regress test for the case is added as well.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -570,13 +570,11 @@ static HandshakeInfo* HandshakeInfoNew(void* heap)
                                     heap, DYNTYPE_HS);
     if (newHs != NULL) {
         WMEMSET(newHs, 0, sizeof(HandshakeInfo));
-        newHs->expectMsgId = MSGID_NONE;
-        newHs->kexId = ID_NONE;
         newHs->kexHashId = WC_HASH_TYPE_NONE;
-        newHs->pubKeyId  = ID_NONE;
-        newHs->encryptId = ID_NONE;
-        newHs->macId = ID_NONE;
         newHs->blockSz = MIN_BLOCK_SZ;
+        newHs->peerBlockSz = MIN_BLOCK_SZ;
+        /* peerEncryptId, peerMacId, peerAeadMode, peerMacSz: left at 0
+         * (== ID_NONE / no-MAC) by the WMEMSET above. */
         newHs->eSz = (word32)sizeof(newHs->e);
         newHs->xSz = (word32)sizeof(newHs->x);
 #ifndef WOLFSSH_NO_DH_GEX_SHA256
@@ -2625,7 +2623,7 @@ static int GenerateKeys(WOLFSSH* ssh, byte hashId, byte doKeyPad)
     Keys* sK = NULL;
     int ret = WS_SUCCESS;
 
-    if (ssh == NULL)
+    if (ssh == NULL || ssh->handshake == NULL)
         ret = WS_BAD_ARGUMENT;
     else {
         if (ssh->ctx->side == WOLFSSH_ENDPOINT_SERVER) {
@@ -2658,19 +2656,17 @@ static int GenerateKeys(WOLFSSH* ssh, byte hashId, byte doKeyPad)
                           sK->encKey, sK->encKeySz,
                           ssh->k, ssh->kSz, ssh->h, ssh->hSz,
                           ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-    if (ret == WS_SUCCESS) {
-        if (!ssh->handshake->aeadMode) {
-            ret = GenerateKey(hashId, 'E',
-                              cK->macKey, cK->macKeySz,
-                              ssh->k, ssh->kSz, ssh->h, ssh->hSz,
-                              ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-            if (ret == WS_SUCCESS) {
-                ret = GenerateKey(hashId, 'F',
-                                  sK->macKey, sK->macKeySz,
-                                  ssh->k, ssh->kSz, ssh->h, ssh->hSz,
-                                  ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-            }
-        }
+    if (ret == WS_SUCCESS && cK->macKeySz > 0) {
+        ret = GenerateKey(hashId, 'E',
+                          cK->macKey, cK->macKeySz,
+                          ssh->k, ssh->kSz, ssh->h, ssh->hSz,
+                          ssh->sessionId, ssh->sessionIdSz, doKeyPad);
+    }
+    if (ret == WS_SUCCESS && sK->macKeySz > 0) {
+        ret = GenerateKey(hashId, 'F',
+                          sK->macKey, sK->macKeySz,
+                          ssh->k, ssh->kSz, ssh->h, ssh->hSz,
+                          ssh->sessionId, ssh->sessionIdSz, doKeyPad);
     }
 
 #ifdef SHOW_SECRETS
@@ -4243,6 +4239,52 @@ static word32 AlgoListSz(const char* algoList)
 }
 
 
+static void SetLocalAlgoIds(HandshakeInfo* hs,
+                             byte encId, byte aeadMode, byte macId)
+{
+    byte blkSz = BlockSzForId(encId);
+    hs->encryptId          = encId;
+    hs->aeadMode           = aeadMode;
+    hs->blockSz            = blkSz;
+    hs->keys.encKeySz      = KeySzForId(encId);
+    if (!aeadMode) {
+        hs->keys.ivSz      = blkSz;
+        hs->macId          = macId;
+        hs->macSz          = MacSzForId(macId);
+        hs->keys.macKeySz  = KeySzForId(macId);
+    }
+    else {
+        hs->keys.ivSz      = AEAD_NONCE_SZ;
+        hs->macId          = ID_NONE;
+        hs->macSz          = blkSz;
+        hs->keys.macKeySz  = 0;
+    }
+}
+
+
+static void SetPeerAlgoIds(HandshakeInfo* hs,
+                            byte encId, byte aeadMode, byte macId)
+{
+    byte blkSz = BlockSzForId(encId);
+    hs->peerEncryptId         = encId;
+    hs->peerAeadMode          = aeadMode;
+    hs->peerBlockSz           = blkSz;
+    hs->peerKeys.encKeySz     = KeySzForId(encId);
+    if (!aeadMode) {
+        hs->peerKeys.ivSz     = blkSz;
+        hs->peerMacId         = macId;
+        hs->peerMacSz         = MacSzForId(macId);
+        hs->peerKeys.macKeySz = KeySzForId(macId);
+    }
+    else {
+        hs->peerKeys.ivSz     = AEAD_NONCE_SZ;
+        hs->peerMacId         = ID_NONE;
+        hs->peerMacSz         = blkSz;
+        hs->peerKeys.macKeySz = 0;
+    }
+}
+
+
 static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 {
     int ret = WS_SUCCESS;
@@ -4258,6 +4300,9 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     word32 cannedAlgoNamesSz;
     word32 skipSz = 0;
     word32 begin;
+    byte c2sEncryptId = ID_NONE, s2cEncryptId = ID_NONE;
+    byte c2sMacId     = ID_NONE, s2cMacId     = ID_NONE;
+    byte c2sAeadMode  = 0,       s2cAeadMode  = 0;
 
     WLOG(WS_LOG_DEBUG, "Entering DoKexInit()");
 
@@ -4399,6 +4444,10 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
             ret = WS_MATCH_ENC_ALGO_E;
         }
     }
+    if (ret == WS_SUCCESS) {
+        c2sEncryptId = algoId;
+        c2sAeadMode  = AeadModeForId(algoId);
+    }
 
     /* Enc Algorithms - Server to Client */
     if (ret == WS_SUCCESS) {
@@ -4407,32 +4456,21 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ret = GetNameList(list, &listSz, buf, len, &begin);
     }
     if (ret == WS_SUCCESS) {
-        algoId = MatchIdLists(side, list, listSz, &algoId, 1);
+        cannedAlgoNamesSz = AlgoListSz(ssh->algoListCipher);
+        cannedListSz = (word32)sizeof(cannedList);
+        ret = GetNameListRaw(cannedList, &cannedListSz,
+                (const byte*)ssh->algoListCipher, cannedAlgoNamesSz);
+    }
+    if (ret == WS_SUCCESS) {
+        algoId = MatchIdLists(side, list, listSz, cannedList, cannedListSz);
         if (algoId == ID_UNKNOWN) {
             WLOG(WS_LOG_DEBUG, "Unable to negotiate Encryption Algo S2C");
             ret = WS_MATCH_ENC_ALGO_E;
         }
     }
     if (ret == WS_SUCCESS) {
-        ssh->handshake->encryptId = algoId;
-        ssh->handshake->aeadMode = AeadModeForId(algoId);
-        ssh->handshake->blockSz = BlockSzForId(algoId);
-        ssh->handshake->keys.encKeySz =
-            ssh->handshake->peerKeys.encKeySz =
-            KeySzForId(algoId);
-        if (!ssh->handshake->aeadMode) {
-            ssh->handshake->keys.ivSz =
-                ssh->handshake->peerKeys.ivSz =
-                ssh->handshake->blockSz;
-        }
-        else {
-#ifndef WOLFSSH_NO_AEAD
-            ssh->handshake->keys.ivSz =
-                ssh->handshake->peerKeys.ivSz =
-                AEAD_NONCE_SZ;
-            ssh->handshake->macSz = ssh->handshake->blockSz;
-#endif
-        }
+        s2cEncryptId = algoId;
+        s2cAeadMode  = AeadModeForId(algoId);
     }
 
     /* MAC Algorithms - Client to Server */
@@ -4441,7 +4479,7 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         listSz = (word32)sizeof(list);
         ret = GetNameList(list, &listSz, buf, len, &begin);
     }
-    if (ret == WS_SUCCESS && !ssh->handshake->aeadMode) {
+    if (ret == WS_SUCCESS && !c2sAeadMode) {
         cannedAlgoNamesSz = AlgoListSz(ssh->algoListMac);
         cannedListSz = (word32)sizeof(cannedList);
         ret = GetNameListRaw(cannedList, &cannedListSz,
@@ -4453,6 +4491,9 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
                 WLOG(WS_LOG_DEBUG, "Unable to negotiate MAC Algo C2S");
                 ret = WS_MATCH_MAC_ALGO_E;
             }
+            else {
+                c2sMacId = algoId;
+            }
         }
     }
 
@@ -4462,18 +4503,20 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         listSz = (word32)sizeof(list);
         ret = GetNameList(list, &listSz, buf, len, &begin);
     }
-    if (ret == WS_SUCCESS && !ssh->handshake->aeadMode) {
-        algoId = MatchIdLists(side, list, listSz, &algoId, 1);
-        if (algoId == ID_UNKNOWN) {
-            WLOG(WS_LOG_DEBUG, "Unable to negotiate MAC Algo S2C");
-            ret = WS_MATCH_MAC_ALGO_E;
-        }
-        else {
-            ssh->handshake->macId = algoId;
-            ssh->handshake->macSz = MacSzForId(algoId);
-            ssh->handshake->keys.macKeySz =
-                ssh->handshake->peerKeys.macKeySz =
-                KeySzForId(algoId);
+    if (ret == WS_SUCCESS && !s2cAeadMode) {
+        cannedAlgoNamesSz = AlgoListSz(ssh->algoListMac);
+        cannedListSz = (word32)sizeof(cannedList);
+        ret = GetNameListRaw(cannedList, &cannedListSz,
+                (const byte*)ssh->algoListMac, cannedAlgoNamesSz);
+        if (ret == WS_SUCCESS) {
+            algoId = MatchIdLists(side, list, listSz, cannedList, cannedListSz);
+            if (algoId == ID_UNKNOWN) {
+                WLOG(WS_LOG_DEBUG, "Unable to negotiate MAC Algo S2C");
+                ret = WS_MATCH_MAC_ALGO_E;
+            }
+            else {
+                s2cMacId = algoId;
+            }
         }
     }
 
@@ -4545,6 +4588,19 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ret = GetUint32(&skipSz, buf, len, &begin);
         if (ret == WS_SUCCESS)
             begin += skipSz;
+    }
+
+    if (ret == WS_SUCCESS) {
+        if (ssh->ctx->side == WOLFSSH_ENDPOINT_SERVER) {
+            /* s2c = local (server outgoing), c2s = peer (client incoming) */
+            SetLocalAlgoIds(ssh->handshake, s2cEncryptId, s2cAeadMode, s2cMacId);
+            SetPeerAlgoIds (ssh->handshake, c2sEncryptId, c2sAeadMode, c2sMacId);
+        }
+        else {
+            /* c2s = local (client outgoing), s2c = peer (server incoming) */
+            SetLocalAlgoIds(ssh->handshake, c2sEncryptId, c2sAeadMode, c2sMacId);
+            SetPeerAlgoIds (ssh->handshake, s2cEncryptId, s2cAeadMode, s2cMacId);
+        }
     }
 
     if (ret == WS_SUCCESS) {
@@ -6226,11 +6282,11 @@ static int DoNewKeys(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     }
 
     if (ret == WS_SUCCESS) {
-        ssh->peerEncryptId = ssh->handshake->encryptId;
-        ssh->peerMacId = ssh->handshake->macId;
-        ssh->peerBlockSz = ssh->handshake->blockSz;
-        ssh->peerMacSz = ssh->handshake->macSz;
-        ssh->peerAeadMode = ssh->handshake->aeadMode;
+        ssh->peerEncryptId = ssh->handshake->peerEncryptId;
+        ssh->peerMacId = ssh->handshake->peerMacId;
+        ssh->peerBlockSz = ssh->handshake->peerBlockSz;
+        ssh->peerMacSz = ssh->handshake->peerMacSz;
+        ssh->peerAeadMode = ssh->handshake->peerAeadMode;
         WMEMCPY(&ssh->peerKeys, &ssh->handshake->peerKeys, sizeof(Keys));
 
         switch (ssh->peerEncryptId) {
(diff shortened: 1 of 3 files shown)


src/internal.c
```diff
@@ -4422,6 +4422,21 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+        if (!ssh->handshake->peerAeadMode) {
+            ssh->handshake->peerKeys.ivSz = ssh->handshake->peerBlockSz;
+        }
+        else {
+#ifndef WOLFSSH_NO_AEAD
+            ssh->handshake->peerKeys.ivSz = AEAD_NONCE_SZ;
+            ssh->handshake->peerMacSz = ssh->handshake->peerBlockSz;
+#endif
+        }
+    }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

src/internal.c
```diff
@@ -4422,6 +4422,21 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+        ssh->handshake->peerKeys.encKeySz = KeySzForId(algoId);
+        if (!ssh->handshake->peerAeadMode) {
+            ssh->handshake->peerKeys.ivSz = ssh->handshake->peerBlockSz;
+        }
+        else {
+#ifndef WOLFSSH_NO_AEAD
+            ssh->handshake->peerKeys.ivSz = AEAD_NONCE_SZ;
+            ssh->handshake->peerMacSz = ssh->handshake->peerBlockSz;
+#endif
+        }
```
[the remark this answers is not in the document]
yosuke-wolfssl: internal.h forces WOLFSSH_NO_AEAD ↔ WOLFSSH_NO_AES_GCM to always agree (lines 283–310), so with AEAD disabled, GCM is never offered in negotiations and AeadModeForId always returns 0.

The #ifndef WOLFSSH_NO_AEAD guards that were inside the AEAD else branches were redundant dead code — entering those branches already proves WOLFSSH_NO_AEAD is unset. I've removed them

src/internal.c
```diff
@@ -4422,6 +4425,22 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
             ret = WS_MATCH_ENC_ALGO_E;
         }
     }
+    if (ret == WS_SUCCESS) {
```
aidangarske, line 4447: **🔴 [High] GenerateKeys skips both-direction MAC key derivation when only one direction is AEAD** · _Logic_

`GenerateKeys` derives MAC key 'E' (C2S) and 'F' (S2C) inside a single `if (!ssh->handshake->aeadMode)` gate. Before this PR, `DoKexInit` enforced that the S2C algorithms had to match the already-selected C2S algorithms (the old code passed `&algoId, 1` as the canned list when matching S2C enc and S2C MAC), so a single `aeadMode` flag correctly described both directions. After this PR, each direction is negotiated independently, and `handshake->aeadMode` now only reflects the S2C direction (`peerAeadMode` reflects C2S). When C2S is non-AEAD and S2C is AEAD — e.g. exactly the configuration exercised by the new `TestIndependentAlgoNegotiation` sub-test B (`aes128-cbc` C2S + `aes256-gcm@openssh.com` S2C) — `handshake->aeadMode == 1` causes the guard to be false and NEITHER 'E' nor 'F' is generated. The C2S/peer HMAC key (`peerKeys.macKey` on the server) therefore stays all-zero from `HandshakeInfoNew`'s `WMEMSET`, and the first MAC verification of an inbound C2S packet will use a zero key and fail, tearing down the session. The new regress test does not catch this because it only asserts `handshake` field values after `DoKexInit` and never drives `GenerateKeys` or the data path. The condition must gate each direction independently (e.g. generate 'E' when the C2S direction is non-AEAD, 'F' when the S2C direction is non-AEAD).

**Fix:** Split the gate so 'E' (C2S MAC key) is generated when the C2S direction is non-AEAD and 'F' (S2C MAC key) is generated when the S2C direction is non-AEAD. On the server: gate 'E' on `!peerAeadMode` (C2S) and 'F' on `!aeadMode` (S2C); on the client these roles are swapped. Then extend the regress test beyond `DoKexInit` to verify `GenerateKeys` derives the right keys (or a higher-level test that completes keying and exchanges data) to keep this regression from reappearing.

tests/regress.c
```diff
@@ -2107,6 +2137,79 @@ static void TestFirstPacketFollows(void)
     TestFirstPacketFollowsSkipped();
 }
 
+#if !defined(WOLFSSH_NO_AES_CBC) && !defined(WOLFSSH_NO_AES_CTR) \
+    && !defined(WOLFSSH_NO_HMAC_SHA1) && !defined(WOLFSSH_NO_HMAC_SHA2_256)
+static void TestIndependentAlgoNegotiation(void)
```
aidangarske, line 2172: **🔵 [Low] Test coverage asserts DoKexInit parse state only, not key derivation / data path for split-AEAD** · _Missing Tests_

`TestIndependentAlgoNegotiation` only drives `wolfSSH_TestDoKexInit` and asserts fields on `ssh->handshake` (e.g. `peerEncryptId`, `encryptId`, `aeadMode`). It does not drive `GenerateKeys`, `DoNewKeys`, or end-to-end encryption/MAC with the negotiated mixed ciphers. As a result, the GenerateKeys bug above (not generating a MAC key for the non-AEAD direction when the other direction is AEAD) is not caught even though sub-test B constructs exactly the offending configuration. This elevates the risk of future regressions in the split-AEAD area.

**Fix:** Extend the test (or add a companion test) that completes keying for a split-AEAD configuration and verifies that the correct MAC keys are derived (non-zero `macKey` with the expected `macKeySz` for the non-AEAD side, zero-sized MAC key on the AEAD side). A round-trip send/receive with HMAC verification would be ideal.

src/internal.c
```diff
@@ -578,6 +578,9 @@ static HandshakeInfo* HandshakeInfoNew(void* heap)
         newHs->encryptId = ID_NONE;
```
aidangarske: **⚪ [Info] HandshakeInfoNew skips peerMacSz/peerAeadMode initialization (safe due to WMEMSET but easy to miss)** · _Logic_

The new initializer block sets `peerEncryptId`, `peerMacId`, and `peerBlockSz`, mirroring the existing local-side initializers. It does not set `peerAeadMode` or `peerMacSz`. Functionally this is fine because `WMEMSET(newHs, 0, sizeof(HandshakeInfo))` zeros the struct first and the desired default for both fields is zero (non-AEAD, zero MAC size). This is purely a consistency/readability observation — future readers adding new peer fields may miss that `WMEMSET` already handles zero defaults, and the selective initialization becomes less symmetric with the local side. No action required for security.

**Fix:** Optional: for symmetry, omit the explicit `= ID_NONE` / `= 0` lines that merely restate the memset defaults, or add the missing `peerMacSz`/`peerAeadMode` / `macSz` / `aeadMode` lines to document intent.

---
> **Note:** Referenced line (`src/internal.c:572-590`) is outside the diff hunk. Comment anchored to nearest changed region.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review-security`

### Posted findings
- [High] **GenerateKeys skips both-direction MAC key derivation when only one direction is AEAD** — `src/internal.c:2696-2709`
- [Low] **Test coverage asserts DoKexInit parse state only, not key derivation / data path for split-AEAD** — `tests/regress.c:2142-2210`
- [Info] **HandshakeInfoNew skips peerMacSz/peerAeadMode initialization (safe due to WMEMSET but easy to miss)** — `src/internal.c:572-590`

src/internal.c
```diff
@@ -4440,21 +4465,14 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ssh->handshake->encryptId = algoId;
         ssh->handshake->aeadMode = AeadModeForId(algoId);
         ssh->handshake->blockSz = BlockSzForId(algoId);
-        ssh->handshake->keys.encKeySz =
-            ssh->handshake->peerKeys.encKeySz =
-            KeySzForId(algoId);
+        ssh->handshake->keys.encKeySz = KeySzForId(algoId);
         if (!ssh->handshake->aeadMode) {
```
yosuke-wolfssl: Fixed

src/internal.c
```diff
@@ -4422,6 +4425,22 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
             ret = WS_MATCH_ENC_ALGO_E;
         }
     }
+    if (ret == WS_SUCCESS) {
+        ssh->handshake->peerEncryptId = algoId;
```
yosuke-wolfssl: Fixed

src/internal.c
yosuke-wolfssl: Fixed

tests/regress.c
yosuke-wolfssl: Extended the tests

yosuke-wolfssl:
Hello @aidangarske ,

I fixed the issues you mentioned.
Thank you for your time.

src/internal.c
```diff
@@ -4292,6 +4296,35 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+            c2sMacSz     = &ssh->handshake->peerMacSz;
+            c2sKeys      =  &ssh->handshake->peerKeys;
+            s2cEncryptId = &ssh->handshake->encryptId;
+            s2cAeadMode  = &ssh->handshake->aeadMode;
+            s2cBlockSz   = &ssh->handshake->blockSz;
+            s2cMacId     = &ssh->handshake->macId;
+            s2cMacSz     = &ssh->handshake->macSz;
+            s2cKeys      =  &ssh->handshake->keys;
+        }
+        else {
```
aidangarske: **🟠 [Medium] DoKexInit client-side branch (WOLFSSH_ENDPOINT_CLIENT alias mapping) lacks a direct asymmetric-algo test** · _Test Coverage_

The PR introduces a new `if (side == WOLFSSH_ENDPOINT_SERVER) { ... } else { ... }` mapping that aliases the eight per-direction storage targets (`c2sEncryptId`/`s2cEncryptId`/etc. and `c2sKeys`/`s2cKeys`) differently for client vs server endpoints. The two new tests `TestIndependentAlgoNegotiation` and `TestGenerateKeysSplit` in `tests/regress.c` only construct `wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL)`, so only the SERVER branch (lines 4299-4312) is exercised with asymmetric per-direction cipher/MAC choices. A grep across `tests/`, `examples/`, and the rest of the repo confirms no caller of `wolfSSH_TestDoKexInit` uses a CLIENT-side context, so the CLIENT branch's mapping (where local=C2S/peer=S2C, the inverse of server) is never validated against a payload that would distinguish a correct mapping from a swapped one. Integration tests do run a real client, but with symmetric algo lists where any swap of `keys`/`peerKeys` for the client side would be invisible. A client-side mapping regression (e.g., accidentally writing C2S results into `peerKeys` on the client) would silently corrupt the wire, since `DoNewKeys` then propagates `handshake->peer*` into `ssh->peer*`.

**Fix:** Add a client-side analogue of `TestIndependentAlgoNegotiation` that builds a `WOLFSSH_ENDPOINT_CLIENT` context, feeds a synthetic peer KEX_INIT with distinct C2S vs S2C cipher/MAC choices (e.g., aes128-cbc/hmac-sha1 C2S, aes256-ctr/hmac-sha2-256 S2C), and asserts that on the client the C2S result lands in `handshake->keys` (local outgoing) and the S2C result lands in `handshake->peerKeys` (peer outgoing) — i.e., the mirror of the server-side asserts. This would be the test that distinguishes a correct client-side alias from an accidentally swapped one.

src/internal.c
```diff
@@ -4391,6 +4424,24 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
             ret = WS_MATCH_ENC_ALGO_E;
         }
     }
+    if (ret == WS_SUCCESS) {
```
aidangarske, line 4447: **🔵 [Low] Asymmetric AEAD reset between C2S and S2C branches in DoKexInit** · _Logic_

The newly added C2S AEAD branch (lines 4435-4443) explicitly resets `*c2sMacId = ID_NONE` and `c2sKeys->macKeySz = 0`, but the analogous S2C AEAD branch (lines 4473-4477) only sets `s2cKeys->ivSz` and `*s2cMacSz`. The two branches should be symmetric. Today this is benign because `HandshakeInfoNew` calls `WMEMSET(newHs, 0, sizeof(HandshakeInfo))` (zero == ID_NONE) and the handshake is freed and reallocated on every rekey (`HandshakeInfoFree` + `ssh->handshake = NULL` in DoNewKeys at internal.c:6331-6332), so `s2cMacId`/`s2cKeys->macKeySz` are guaranteed to start at 0. However, the implicit reliance on freshly-zeroed fields makes the S2C branch fragile: any future change that makes HandshakeInfo persist across rekeys, or that allows DoKexInit to be re-entered without freeing handshake, could leave a stale non-AEAD MacId/macKeySz in the S2C direction. Because GenerateKeys (internal.c:2664) now triggers MAC key 'F' derivation when `sK->macKeySz > 0`, a stale macKeySz would silently re-introduce a MAC key on an AEAD direction, and DoNewKeys (internal.c:6270) would assign a stale `peerMacId` for the AEAD direction. Both reviewers concur — severity Low / action NIT — but flag this as the same WMEMSET-reliance pattern as the implicit init removed in `HandshakeInfoNew`, just in a different location. Fix is trivial.

**Fix:** Mirror the C2S resets in the S2C AEAD branch — add `*s2cMacId = ID_NONE; s2cKeys->macKeySz = 0;` so both directions explicitly clear the stale MAC state when AEAD is selected. This removes the implicit dependency on the surrounding allocator behavior and costs only two lines.

src/internal.c
```diff
@@ -4250,6 +4244,16 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     word32 cannedAlgoNamesSz;
     word32 skipSz = 0;
     word32 begin;
+    /* handshake->keys/encryptId/... always represent the LOCAL endpoint's
+     * outgoing direction; peer* counterparts represent the peer's outgoing
+     * (= our incoming) direction.  Server: local=S2C, peer=C2S.
+     * Client: local=C2S, peer=S2C.  These aliases let the four enc/MAC
+     * parse sections store results without inline side checks, and keep the
+     * fields consistent with what GenerateKeys/SendNewKeys/DoNewKeys expect. */
+    byte *c2sEncryptId, *c2sAeadMode, *c2sBlockSz, *c2sMacId, *c2sMacSz;
```
aidangarske: **🔵 [Low] Pointer aliases used after first ret==WS_SUCCESS block may trip strict-uninit warnings** · _Logic_

The new local pointers `c2sEncryptId/s2cEncryptId/...` and `c2sKeys/s2cKeys` (declared at lines 4253-4256) are only assigned inside the `if (ret == WS_SUCCESS)` block at line 4295. They are subsequently dereferenced inside other `if (ret == WS_SUCCESS)` blocks (e.g., 4427, 4465, 4486, 4512). The control-flow invariant `ret == WS_SUCCESS at line 4427 implies the assignment block at 4295 ran` is correct (ret is only ever set to error or kept WS_SUCCESS — never reset to WS_SUCCESS), so the dereferences are safe at runtime. However, the connection between the assignment block and the use sites is via a chain of `ret == WS_SUCCESS` checks rather than direct dominance, which some static analyzers and `-Wmaybe-uninitialized` builds may flag. This is a code-quality observation, not an exploitable bug.

**Fix:** Either initialize the pointer aliases to NULL at declaration (clearer to readers and to static analyzers), or move the alias assignments out of the `if (ret == WS_SUCCESS)` block since they only depend on `ssh->ctx->side` which was already validated. Optionally suppress with explicit `(void)c2sEncryptId;` if a build hits `-Wmaybe-uninitialized`.

src/internal.c
```diff
@@ -571,13 +571,9 @@ static HandshakeInfo* HandshakeInfoNew(void* heap)
                                     heap, DYNTYPE_HS);
     if (newHs != NULL) {
```
aidangarske, line 571: **⚪ [Info] Removed redundant ID_NONE/MSGID_NONE initializations in HandshakeInfoNew rely on enum values being 0** · _Logic_

The PR removes the explicit assignments `expectMsgId = MSGID_NONE; kexId = ID_NONE; pubKeyId = ID_NONE; encryptId = ID_NONE; macId = ID_NONE;` from `HandshakeInfoNew`. This is safe today: `MSGID_NONE` is explicitly defined as `0` (wolfssh/internal.h:1221) and `ID_NONE` is the first value of the unnamed enum at wolfssh/internal.h:324, so it is also `0`. The preceding `WMEMSET(newHs, 0, sizeof(HandshakeInfo))` therefore covers all of these. No regression. Note for future maintainers: if either `ID_NONE` or `MSGID_NONE` is ever moved away from value `0`, all of these fields will silently become `ID_NONE != actual-zero` after `WMEMSET`, breaking everything that compares with the sentinel. Consider adding a `WC_STATIC_ASSERT(ID_NONE == 0)` (or equivalent) to lock the invariant in.

**Fix:** Either re-add the explicit assignments (defensive) or add a compile-time assertion that ties the implicit-zero behavior to the enum/macro values. This is informational only — the current change is correct given the existing values.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security + audit`

### Posted findings
- [Medium] [audit] **DoKexInit client-side branch (WOLFSSH_ENDPOINT_CLIENT alias mapping) lacks a direct asymmetric-algo test** — `src/internal.c:4313-4326`
- [Low] [review+review-security] **Asymmetric AEAD reset between C2S and S2C branches in DoKexInit** — `src/internal.c:4427-4478`
- [Low] [review-security] **Pointer aliases used after first ret==WS_SUCCESS block may trip strict-uninit warnings** — `src/internal.c:4253-4326,4427-4527`
- [Info] [review-security] **Removed redundant ID_NONE/MSGID_NONE initializations in HandshakeInfoNew rely on enum values being 0** — `src/internal.c:572-576`

yosuke-wolfssl:
Hi @aidangarske ,

Sorry to take your time.
Can you review again please ?

src/internal.c
```diff
@@ -4391,6 +4426,24 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+             * WOLFSSH_NO_AES_GCM to be unset, hence WOLFSSH_NO_AEAD unset
+             * (see internal.h). */
+            c2sKeys->ivSz  = AEAD_NONCE_SZ;
+            *c2sMacSz      = *c2sBlockSz;
+            *c2sMacId      = ID_NONE;
+            c2sKeys->macKeySz = 0;
+        }
+    }
 
     /* Enc Algorithms - Server to Client */
```
aidangarske, line 4452: **🟠 [Medium] Asymmetric S2C-only encryption-algo mismatch error path is untested** · _path_

Pre-PR, the S2C encryption algorithm was matched against the C2S-chosen algorithm only (`MatchIdLists(side, list, listSz, &algoId, 1)`). Post-PR, S2C is matched independently against `ssh->algoListCipher` via `cannedList`. This introduces a NEW failure mode: C2S successfully matches but S2C does not (or vice versa) — yielding `WS_MATCH_ENC_ALGO_E` from the S2C block. The new tests in `TestIndependentAlgoNegotiation`/`...Client` only cover the success paths for asymmetric cipher selection; no test exercises asymmetric mismatch where one direction agrees and the other fails. Grep confirms `WS_MATCH_ENC_ALGO_E` has zero references in `tests/`.

**Fix:** Add a regress.c test that builds a kex_init payload where the C2S cipher list contains an algorithm in the local `algoListCipher` but the S2C cipher list does not (e.g., local accepts `aes128-cbc,aes256-ctr`; payload offers C2S=`aes128-cbc`, S2C=`3des-cbc`). Verify `wolfSSH_TestDoKexInit` returns `WS_MATCH_ENC_ALGO_E`. Mirror the test for the inverse case (C2S fails, S2C succeeds).

src/internal.c
```diff
@@ -4454,18 +4513,22 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+                WLOG(WS_LOG_DEBUG, "Unable to negotiate MAC Algo S2C");
+                ret = WS_MATCH_MAC_ALGO_E;
+            }
+            else {
+                *s2cMacId          = algoId;
+                *s2cMacSz          = MacSzForId(algoId);
+                s2cKeys->macKeySz  = KeySzForId(algoId);
+            }
         }
     }
```
aidangarske, line 4521: **🟠 [Medium] Asymmetric S2C-only MAC-algo mismatch error path is untested** · _path_

The S2C MAC negotiation was rewritten from `MatchIdLists(side, list, listSz, &algoId, 1)` (which forced S2C to equal C2S) to an independent match against `ssh->algoListMac`. The new failure path `WS_MATCH_MAC_ALGO_E` can now trigger when C2S MAC matches but S2C MAC does not (and vice versa for the C2S branch at lines 4484-4508). Neither branch's `WS_MATCH_MAC_ALGO_E` return is exercised by any test. Grep across the repo confirms no test references `WS_MATCH_MAC_ALGO_E`.

**Fix:** Add a regress.c test that supplies asymmetric MAC lists where one direction shares no entries with `ssh->algoListMac` (e.g., local accepts `hmac-sha1,hmac-sha2-256`; payload offers C2S=`hmac-sha1`, S2C=`hmac-md5`). Assert `wolfSSH_TestDoKexInit` returns `WS_MATCH_MAC_ALGO_E`. Cover the inverse C2S-mismatch case as well.

tests/regress.c
```diff
@@ -2128,6 +2158,406 @@ static void TestFirstPacketFollows(void)
     TestFirstPacketFollowsSkipped();
 }
 
+#if !defined(WOLFSSH_NO_AES_CBC) && !defined(WOLFSSH_NO_AES_CTR) \
```
aidangarske, line 2170: **🟠 [Medium] TestIndependentAlgoNegotiation tests do not verify all ID/Sz fields are reset on each ssh instance** · _test_

The new tests only assert the *split* fields (`peerEncryptId`/`peerMacId`/`peerAeadMode`/`peerBlockSz`/`peerMacSz`) for the side they care about, but never assert `peerBlockSz`/`peerMacSz` directly. Given that the diff intentionally drops the explicit `ID_NONE`/`MSGID_NONE` initializers in `HandshakeInfoNew` (relying on the WMEMSET zero plus the assumption that `ID_NONE == 0`), an asymmetric subtle regression — e.g. `peerBlockSz` left at its `MIN_BLOCK_SZ` default in a code path that should overwrite it — would not be caught. A few additional `AssertIntEQ` lines for `peerBlockSz`, `peerMacSz`, `blockSz`, `macSz` after each negotiation would harden the tests against future refactors of `HandshakeInfoNew` or `DoKexInit`.

**Fix:** Add explicit assertions for the new `peerBlockSz`/`peerMacSz` fields (and for `blockSz`/`macSz` for symmetry) in both server and client variants.

src/internal.c
```diff
@@ -17896,6 +17959,13 @@ int wolfSSH_TestDoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     return DoKexInit(ssh, buf, len, idx);
 }
 
+int wolfSSH_TestGenerateKeys(WOLFSSH* ssh)
```
aidangarske: **🔵 [Low] wolfSSH_TestGenerateKeys NULL-argument validation is untested** · _test_

The newly added test helper validates `ssh == NULL || ssh->handshake == NULL` and returns `WS_BAD_ARGUMENT`, but no test in `tests/regress.c` (or anywhere else) calls it with NULL inputs to verify the guard. All four new test sites (`TestGenerateKeysSplit`, `TestGenerateKeysSplitClient` sub-tests A and B) invoke it only after `AssertNotNull(ssh->handshake)`. While the function is test-only infrastructure, a sibling invariant check is cheap to add.

**Fix:** Add a one-line assertion in regress.c (e.g., inside or adjacent to `TestGenerateKeysSplit`): `AssertIntEQ(wolfSSH_TestGenerateKeys(NULL), WS_BAD_ARGUMENT);` and a second call with a freshly-created `WOLFSSH*` whose handshake has been freed/nulled to verify the second branch.

tests/regress.c
```diff
@@ -2128,6 +2158,406 @@ static void TestFirstPacketFollows(void)
...
+    /* S2C AEAD: macKeySz==0 so key 'F' was skipped; macKey stays all-zero. */
+    AssertIntEQ(ssh->handshake->keys.macKeySz, 0);
+    AssertIntEQ(WMEMCMP(ssh->handshake->keys.macKey, zeros,
+                        WC_SHA_DIGEST_SIZE), 0);
+
+    wolfSSH_free(ssh);
+#endif /* !WOLFSSH_NO_AES_GCM */
+
+    wolfSSH_CTX_free(ctx);
+}
```
aidangarske, line 2710: **🔵 [Low] Missing blank line between TestGenerateKeysSplit and TestGenerateKeysSplitClient** · _style_

Every other top-level function in `regress.c` (including the other three new test functions added by this PR) is separated by a blank line. The closing brace of `TestGenerateKeysSplit` is followed immediately by the next function definition, which breaks consistency.

**Fix:** Add a blank line between the two functions to match the surrounding style.

tests/regress.c
```diff
@@ -2128,6 +2158,406 @@ static void TestFirstPacketFollows(void)
...
+            "hmac-sha1,hmac-sha2-256"), WS_SUCCESS);
+    idx = 0;
+    payloadSz = BuildKexInitPayloadFull(
+            FPF_KEX_GOOD, FPF_KEY_GOOD,
+            "aes128-cbc",    /* C2S enc */
+            "aes256-ctr",    /* S2C enc */
+            "hmac-sha1",     /* C2S MAC */
+            "hmac-sha2-256", /* S2C MAC */
+            0, payload, (word32)sizeof(payload));
+    (void)wolfSSH_TestDoKexInit(ssh, payload, payloadSz, &idx);
```
aidangarske, line 2202: **🔵 [Low] New tests silently ignore wolfSSH_TestDoKexInit return without explanation** · _test_

All four new tests call `(void)wolfSSH_TestDoKexInit(ssh, payload, payloadSz, &idx);` and then assert handshake state. The pre-existing `RunFirstPacketFollowsCase` (line 2088) has an explanatory comment about why the return is ignored (the tail of `DoKexInit` calls `SendKexInit` which fails for a stripped-down server with no keys). The four new test functions reuse this pattern but do not carry a similar explanation, so a future reader may mistakenly think the discarded return is a bug. For the client-side variants the rationale is even less obvious, since the privateKeyCount check does not apply to clients — the `SendKexInit` failure there comes from `wolfSSH_SendPacket` having no IO callback set up.

**Fix:** Add a brief comment in each new test (or factor a helper that wraps the call) explaining why the return value is intentionally discarded.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `audit + review + review-security`
**Findings:** 7 total — 7 posted, 0 skipped
**6 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] [audit] **Asymmetric S2C-only encryption-algo mismatch error path is untested** — `src/internal.c:4448-4466`
- [Medium] [audit] **Asymmetric S2C-only MAC-algo mismatch error path is untested** — `src/internal.c:4510-4533`
- [Medium] [review] **TestIndependentAlgoNegotiation tests do not verify all ID/Sz fields are reset on each ssh instance** — `tests/regress.c:2161-2332`
- [Low] [audit] **wolfSSH_TestGenerateKeys NULL-argument validation is untested** — `src/internal.c:17962-17967`
- [Low] [review] **Missing blank line between TestGenerateKeysSplit and TestGenerateKeysSplitClient** — `tests/regress.c:2446-2447`
- [Low] [review] **New tests silently ignore wolfSSH_TestDoKexInit return without explanation** — `tests/regress.c:2191,2227,2278,2313,2366,2420,2479,2532`
- [Low] [review] **DoKexInit reads side from ssh->ctx->side a second time after caching it in `side`** — `src/internal.c:4622,4642,4666,4673`

dgarske:
@yosuke-wolfssl please see the Fenrir feedback.

src/internal.c
yosuke-wolfssl: Added the test

tests/regress.c
yosuke-wolfssl: Added comments

yosuke-wolfssl:
Hello @aidangarske ,
I also fixed some issues with Skoll now.
Could you please check this ?

src/internal.c
```diff
@@ -4258,6 +4254,18 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     word32 cannedAlgoNamesSz;
     word32 skipSz = 0;
     word32 begin;
+    /* handshake->keys/encryptId/... always represent the LOCAL endpoint's
```
ejohnstown: I think this can get solved with adding just a few bytes and not having the pointers. We'd have byte c2sEncryptId, s2cEncryptId, s2cMacId, c2SMacId. Read those values. When we're done reading the message, we can check the side and do all the algorithm IDs or peer algorithm IDs depending on the side. All the other values that are based on the read-from-message IDs can be set as well. We'd also need s2cAeadMode and c2sAeadMode flags.

ejohnstown (changes requested):

ejohnstown (approved):

PR #953, opened 2026-04-24
Add service-name check and regress test

yosuke-wolfssl:
This PR adds the service-name check to comply with RFC 4252 section 5.
If service name is not equal to "ssh-connection", SSH_MSG_USERAUTH_FAILURE would be sent to the peer.
Also, new unit test is added to exercise service-name validation.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8345,7 +8345,8 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
 {
     word32 begin;
     int ret = WS_SUCCESS;
-    byte authNameId;
+    byte authNameId = ID_UNKNOWN;
+    int serviceValid = 1;
     WS_UserAuthData authData;
 
     WLOG(WS_LOG_DEBUG, "Entering DoUserAuthRequest()");
@@ -8356,37 +8357,32 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
     if (ret == WS_SUCCESS) {
         begin = *idx;
         WMEMSET(&authData, 0, sizeof(authData));
-        ret = GetSize(&authData.usernameSz, buf, len, &begin);
+        ret = GetStringRef(&authData.usernameSz, &authData.username,
+                buf, len, &begin);
     }
 
     if (ret == WS_SUCCESS) {
-        authData.username = buf + begin;
-        begin += authData.usernameSz;
-
-        ret = GetUint32(&authData.serviceNameSz, buf, len, &begin);
+        ret = GetStringRef(&authData.serviceNameSz, &authData.serviceName,
+                buf, len, &begin);
     }
 
     if (ret == WS_SUCCESS) {
-        ret = wolfSSH_SetUsernameRaw(ssh, authData.username, authData.usernameSz);
-    }
-
-    if (ret == WS_SUCCESS) {
-        if (authData.serviceNameSz > len - begin) {
-            ret = WS_BUFFER_E;
+        if (NameToId((const char*)authData.serviceName, authData.serviceNameSz)
+                != ID_SERVICE_CONNECTION) {
+            WLOG(WS_LOG_DEBUG, "DUAR: Invalid service name");
+            serviceValid = 0;
+            ret = SendUserAuthFailure(ssh, 0);
+            /* Consume all remaining data */
+            *idx = len;
+        }
+        else {
+            ret = GetStringRef(&authData.authNameSz, &authData.authName,
+                    buf, len, &begin);
         }
     }
 
-    if (ret == WS_SUCCESS) {
-        authData.serviceName = buf + begin;
-        begin += authData.serviceNameSz;
-
-        ret = GetSize(&authData.authNameSz, buf, len, &begin);
-    }
-
-    if (ret == WS_SUCCESS) {
-        authData.authName = buf + begin;
-        begin += authData.authNameSz;
-        authNameId = NameToId((char*)authData.authName, authData.authNameSz);
+    if (ret == WS_SUCCESS && serviceValid) {
+        authNameId = NameToId((const char*)authData.authName, authData.authNameSz);
         ssh->authId = authNameId;
 
         if (authNameId == ID_USERAUTH_PASSWORD)
@@ -8409,11 +8405,14 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
 #endif
         else {
             WLOG(WS_LOG_DEBUG,
-                 "invalid userauth type: %s", IdToName(authNameId));
+                 "DUAR: invalid userauth type: %s", IdToName(authNameId));
             ret = SendUserAuthFailure(ssh, 0);
+            /* Consume all remaining data */
+            begin = len;
         }
 
         if (ret == WS_SUCCESS) {
+            /* Set the username for valid service only */
             ret = wolfSSH_SetUsernameRaw(ssh,
                     authData.username, authData.usernameSz);
         }
@@ -17974,6 +17973,12 @@ int wolfSSH_TestChannelPutData(WOLFSSH_CHANNEL* channel, byte* data,
         word32 dataSz)
 {
     return ChannelPutData(channel, data, dataSz);
+}
+
+int wolfSSH_TestDoUserAuthRequest(WOLFSSH* ssh, byte* buf, word32 len,
+        word32* idx)
+{
+    return DoUserAuthRequest(ssh, buf, len, idx);
 }
 
 #ifndef WOLFSSH_NO_DH_GEX_SHA256
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -1181,13 +1181,25 @@ done:
     return result;
 }
 
+/* Plaintext SSH packet from IoSend (before encryption/MAC): LENGTH_SZ,
+ * PAD_LENGTH_SZ, then payload starting with the message ID (RFC 4253;
+ * wolfSSH PreparePacket/BundlePacket). Not for encrypted payloads or
+ * arbitrary truncated chunks. */
+static int CaptureMsgId(const byte* buf, word32 len)
+{
+    word32 off = LENGTH_SZ + PAD_LENGTH_SZ;
+
+    if (len <= off)
+        return -1;
+    return (int)buf[off];
+}
+
 /* Verify DoChannelRequest sends CHANNEL_SUCCESS for known types and
  * CHANNEL_FAILURE for unrecognized ones (RFC 4254 Section 5.4).
  *
  * A custom IoSend callback captures the outgoing packet in plaintext
- * (no cipher negotiated on a fresh session). The SSH packet layout is:
- *   [4-byte packet_length][1-byte padding_length][1-byte msg_id]...
- * so the message ID lives at byte offset 5. */
+ * (no cipher negotiated on a fresh session). Message ID is read via
+ * CaptureMsgId() using LENGTH_SZ + PAD_LENGTH_SZ. */
 static byte   s_chanReqCapture[256];
 static word32 s_chanReqCaptureSz = 0;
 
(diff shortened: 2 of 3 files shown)


tests/unit.c
```diff
@@ -918,6 +918,86 @@ static int test_DoChannelRequest(void)
...
+        /* service name */
+        buf[len++] = (byte)(snsz >> 24); buf[len++] = (byte)(snsz >> 16);
+        buf[len++] = (byte)(snsz >>  8); buf[len++] = (byte)snsz;
+        WMEMCPY(buf + len, cases[i].svcName, snsz); len += snsz;
+
+        /* auth method: "none" */
+        buf[len++] = 0; buf[len++] = 0; buf[len++] = 0; buf[len++] = 4;
+        WMEMCPY(buf + len, "none", 4); len += 4;
+
+        ret = wolfSSH_TestDoUserAuthRequest(ssh, buf, len, &idx);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

tests/unit.c
```diff
@@ -918,6 +918,86 @@ static int test_DoChannelRequest(void)
...
+
+        /* service name */
+        buf[len++] = (byte)(snsz >> 24); buf[len++] = (byte)(snsz >> 16);
+        buf[len++] = (byte)(snsz >>  8); buf[len++] = (byte)snsz;
+        WMEMCPY(buf + len, cases[i].svcName, snsz); len += snsz;
+
+        /* auth method: "none" */
+        buf[len++] = 0; buf[len++] = 0; buf[len++] = 0; buf[len++] = 4;
+        WMEMCPY(buf + len, "none", 4); len += 4;
+
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

src/internal.c
```diff
@@ -8401,10 +8402,19 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
         authData.serviceName = buf + begin;
         begin += authData.serviceNameSz;
 
-        ret = GetSize(&authData.authNameSz, buf, len, &begin);
+        if (NameToId((const char*)authData.serviceName, authData.serviceNameSz)
+                != ID_SERVICE_CONNECTION) {
+            WLOG(WS_LOG_DEBUG, "DUAR: Invalid service name");
```
padelsbach, line 8372: The other debug logs in this function do not have `DUAR` -- consider updating those to match this one, following convention from `DoUserAuthRequestNone`.

src/internal.c
```diff
@@ -8401,10 +8402,19 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
...
+            serviceValid = 0;
+            ret = SendUserAuthFailure(ssh, 0);
+            *idx = len;
```
padelsbach, line 8376: Other error cases in this function do not set `*idx`. Why is this needed for this specific error? Can you add a comment?

padelsbach:
@yosuke-wolfssl, can you please rebase to resolve conflicts?

src/internal.c
```diff
@@ -8401,10 +8402,19 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
         authData.serviceName = buf + begin;
         begin += authData.serviceNameSz;
 
-        ret = GetSize(&authData.authNameSz, buf, len, &begin);
+        if (NameToId((const char*)authData.serviceName, authData.serviceNameSz)
```
aidangarske, line 8370: **🔵 [Low] Inconsistent cast style in new NameToId call**

The new service-name check casts `authData.serviceName` as `(const char*)`, but the immediately adjacent NameToId call on line 8420 for `authData.authName` uses `(char*)`. Both compile fine because `NameToId` takes `const char*`, but the mixed styles inside the same function are distracting.

**Fix:** Pick one cast style for NameToId in this function (either both `(char*)` or both `(const char*)`) so the two call sites read identically.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`
**Overall recommendation:** `APPROVE`
**Findings:** 7 total — 7 posted, 0 skipped
**7 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Low] **Inconsistent cast style in new NameToId call** — `src/internal.c:8405-8406`
- [Low] **Error from SendUserAuthFailure overwrites ret, making invalid-service path return non-success** — `src/internal.c:8405-8411`
- [Low] **`*idx = len` is written even when SendUserAuthFailure returned an error** — `src/internal.c:8409-8411`
- [Low] **Test assumes output path succeeds on a fresh session without documentation** — `tests/unit.c:952-1041`
- [Low] **Test does not cover empty or oversize service names** — `tests/unit.c:964-973`
- [Info] **`serviceValid` could be replaced by a guard-style early return block for readability** — `src/internal.c:8370,8417`
- [Info] **`*idx = len` set before SendUserAuthFailure result is checked** — `src/internal.c:8409-8410`

src/internal.c
aidangarske, line 8370: **🔵 [Low] Error from SendUserAuthFailure overwrites ret, making invalid-service path return non-success**

When the service name is invalid the code calls `ret = SendUserAuthFailure(ssh, 0)`. If `SendUserAuthFailure()` fails (e.g., `WS_SOCKET_ERROR_E`, `WS_WANT_WRITE`), that error is returned to the outer dispatcher as if parsing itself failed. In the unknown-auth-method path below (line 8444) the same pattern is used, so this is internally consistent with the existing convention — but it is worth confirming that propagating a send error here is the intended behavior (e.g., `WS_WANT_WRITE` from a non-blocking socket could reach `DoReceive` here and bubble up differently than the caller expects). The unit test uses a capture-callback that always returns success, so this path is not exercised.

**Fix:** Confirm that it is acceptable to return the `SendUserAuthFailure()` result (including `WS_WANT_WRITE`) to the `DoReceive` loop when the only parsing failure was a bad service name. If the intent is "we handled this request, stay alive," consider saving the send result separately from `ret` or adding a test where the IO callback returns `WS_CBIO_ERR_WANT_WRITE`.

src/internal.c
```diff
@@ -8401,10 +8402,19 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
...
+                != ID_SERVICE_CONNECTION) {
+            WLOG(WS_LOG_DEBUG, "DUAR: Invalid service name");
+            serviceValid = 0;
+            ret = SendUserAuthFailure(ssh, 0);
```
aidangarske, line 8374: **🔵 [Low] `*idx = len` is written even when SendUserAuthFailure returned an error**

On the invalid-service path, `ret = SendUserAuthFailure(ssh, 0); *idx = len;` advances idx unconditionally. If `SendUserAuthFailure` returns a non-success value (e.g., a non-blocking `WS_WANT_WRITE` or an allocation failure inside `PreparePacket`), the function returns the error to `DoPacket` but idx is still fully consumed. For most error paths this is harmless because the caller tears the connection down, but it does mean the decoded-length cursor no longer matches what was actually processed. Other handlers in this file leave idx alone when ret is non-success.

**Fix:** Consider only advancing `*idx = len` when `SendUserAuthFailure` returns `WS_SUCCESS`, matching the idx-preservation convention used elsewhere in this function on error paths.

tests/unit.c
```diff
@@ -918,6 +918,128 @@ static int test_DoChannelRequest(void)
...
+ * check and still satisfy all three assertions — proving it never tries to
+ * parse the missing auth-method field.  If the short-circuit were absent,
+ * GetSize() for authNameSz would hit end-of-buffer and return WS_BUFFER_E,
+ * failing assertion 1.
+ *
+ * For the valid-service case, auth method "xyz-unknown" (always unsupported
+ * regardless of compile-time options) is included.  The function reaches
+ * auth-method dispatch, falls to the unknown-method else-branch, and sends
+ * USERAUTH_FAILURE via that normal path. */
+static int test_DoUserAuthRequest_serviceName(void)
```
aidangarske, line 1376: **🔵 [Low] Test assumes output path succeeds on a fresh session without documentation**

The test relies on no cipher being negotiated so the captured packet is plaintext with `MSGID_USERAUTH_FAILURE` at byte offset 5. That assumption is already documented in `test_DoChannelRequest` (lines 792-797) but not here. A one-line comment referencing the same packet-layout invariant would save the next reader from retracing the logic.

**Fix:** Add a short comment explaining why byte 5 holds the message ID, for parity with `test_DoChannelRequest`.

tests/unit.c
```diff
@@ -918,6 +918,128 @@ static int test_DoChannelRequest(void)
...
+    WOLFSSH* ssh = NULL;
+    int result = 0;
+    struct {
+        const char* svcName;
+        word32      svcNameSz;
+        const char* authMethod;   /* NULL = omit field (proves short-circuit) */
+        word32      authMethodSz;
+        int         expectRet;
+        const char* label;
+    } cases[] = {
```
aidangarske, line 1390: **🔵 [Low] Test does not cover empty or oversize service names**

The new test exercises three cases (`ssh-connection`, `ssh-agent`, `bad`). It does not cover the boundary cases `serviceNameSz == 0` (zero-length service name) or a service name equal to another known service ID (e.g., `ssh-userauth`). Both should also be rejected by the new `NameToId(...) != ID_SERVICE_CONNECTION` check and would provide stronger evidence that only `ssh-connection` is accepted. `NameToId("", 0)` returns `ID_UNKNOWN`, and `NameToId("ssh-userauth", 12)` returns `ID_SERVICE_USERAUTH`, both of which should short-circuit — but neither is currently verified.

**Fix:** Add a zero-length service-name case and an `ssh-userauth` service-name case to the `cases[]` table to verify the check is strict and not dependent on `NameToId` returning `ID_UNKNOWN` specifically.

src/internal.c
```diff
@@ -8367,6 +8367,7 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
     word32 begin;
     int ret = WS_SUCCESS;
     byte authNameId;
+    byte serviceValid = 1;
```
aidangarske: **⚪ [Info] `serviceValid` could be replaced by a guard-style early return block for readability**

A new local `byte serviceValid = 1;` is introduced solely to skip the auth-method dispatch block after the short-circuit. The function already uses the chained `if (ret == WS_SUCCESS)` pattern throughout. An alternative would be to leave `ret` as `WS_SUCCESS` (since the packet was handled) and skip the dispatch by leaving `authData.authName` zero-length and `authNameId` == `ID_UNKNOWN`, which already falls into the else-branch at line 8441. But that would still call `wolfSSH_SetUsernameRaw` a second time and `SendUserAuthFailure` a second time, which is undesirable. The `serviceValid` flag is fine as-is.

**Fix:** Keep as-is. Flag is clearer than refactoring around the existing chained-if pattern.

src/internal.c
aidangarske, line 8374: **⚪ [Info] `*idx = len` set before SendUserAuthFailure result is checked**

`*idx` is written unconditionally immediately after (and independent of) the `SendUserAuthFailure()` call. The surrounding convention in this function (and in `DoReceive`) is "advance the buffer index on success, leave untouched on failure." Here the index is advanced even when the send fails. The `DoReceive` dispatcher at src/internal.c:10178 will still advance `ssh->inputBuffer.idx` by `payloadIdx` even on error, so the behavior is not incorrect — but the asymmetry with the valid-path write `*idx = begin;` at line 8452 (guarded by `if (ret == WS_SUCCESS && serviceValid)`) reads slightly oddly.

**Fix:** Optional: move `*idx = len;` above the `SendUserAuthFailure` call to make it clear the assignment is unconditional and not contingent on send success, or leave as-is. Not a correctness issue.

src/internal.c
yosuke-wolfssl: Fixed

src/internal.c
yosuke-wolfssl: Fixed to cast both to const char*

src/internal.c
yosuke-wolfssl: Added a comment

src/internal.c
yosuke-wolfssl: This shall consume all of data to release the input buffer properly.
I fixed both of cases.

src/internal.c
yosuke-wolfssl: OK

src/internal.c
yosuke-wolfssl: OK

src/internal.c
yosuke-wolfssl: I'll keep current convention as-is.
I didn't find any problematic behaviour for now.

tests/unit.c
yosuke-wolfssl: Fixed

tests/unit.c
yosuke-wolfssl: Added two cases you mentioned.

yosuke-wolfssl:
Hello @aidangarske ,

It's ready for review again.
Could you please take a look at this ?

src/internal.c
```diff
@@ -8361,13 +8362,23 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
         authData.serviceName = buf + begin;
         begin += authData.serviceNameSz;
 
-        ret = GetSize(&authData.authNameSz, buf, len, &begin);
```
ejohnstown, line 8383: The function GetStringRef() was added for cases like this. It does all the appropriate length checking and it updates the idx. I just never got around to replacing all GetSize() and adjusting the idx myself with this function.

GetStringRef will give you a pointer into buf that points at the string, and it gives you the length. You don't need to free it, and the pointer is to const.

ejohnstown (changes requested):

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [Medium] **wolfSSH_SetUsernameRaw mutates session state before the service-name check** — `src/internal.c:8351-8376`
- [Low] **Unrelated blank-line removal between DoReceive and DoProtoId** — `src/internal.c:10719`
- [Low] **Test couples to internal packet layout (byte-5 message-id) without abstraction** — `tests/unit.c:1027-1035`

src/internal.c
```diff
@@ -8361,13 +8362,23 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
...
+            WLOG(WS_LOG_DEBUG, "DUAR: Invalid service name");
+            serviceValid = 0;
+            ret = SendUserAuthFailure(ssh, 0);
+            /* Consume all remaining data */
+            *idx = len;
+        }
+        else {
+            ret = GetSize(&authData.authNameSz, buf, len, &begin);
+        }
     }
```
aidangarske, line 8382: **🟠 [Medium] wolfSSH_SetUsernameRaw mutates session state before the service-name check**

The username is committed into `ssh->userName` at line 8351 *before* the new service-name validation runs. If the peer sends a request with a non-`ssh-connection` service name (or any other malformed thing later in parsing), the session keeps the attacker-supplied username while we send `USERAUTH_FAILURE`. The PR doesn't introduce this ordering, but by adding a new reject path between the username set and the auth-method dispatch it widens the window where this matters. A second retry will overwrite the field, so this is mostly cosmetic, but consider deferring `wolfSSH_SetUsernameRaw` until after we accept the request (or at least until after the service name is validated) so that rejected requests don't leave residual state. This would also let you drop the duplicate `wolfSSH_SetUsernameRaw` call at line 8410-8413.

**Fix:** Set the username only on a fully-validated request to avoid storing state from rejected attempts.

tests/unit.c
```diff
@@ -918,6 +918,136 @@ static int test_DoChannelRequest(void)
...
+
+        if (ret != cases[i].expectRet) {
+            printf("DoUserAuthRequest_svcName[%s]: ret=%d expected=%d\n",
+                   cases[i].label, ret, cases[i].expectRet);
+            result = -502 - i;
+            break;
+        }
+
+        /* MSGID_USERAUTH_FAILURE must be in the captured packet. */
+        if (s_authSvcCaptureSz <= 5 ||
```
aidangarske: **🔵 [Low] Test couples to internal packet layout (byte-5 message-id) without abstraction**

The assertion `s_authSvcCapture[5] != MSGID_USERAUTH_FAILURE` hard-codes the offset of the message id in the framed packet (4-byte length + 1-byte padding length + msg id). This is the same approach used by `test_DoChannelRequest`, so the inconsistency is repo-wide rather than PR-introduced, but every new copy of this pattern increases the cleanup cost if/when packet framing changes. Consider extracting a small helper like `static int CaptureMsgId(byte* buf, word32 len)` that the existing channel-request test could also adopt.

**Fix:** Optional cleanup; not a merge blocker.

src/internal.c
```diff
@@ -10716,7 +10729,6 @@ int DoReceive(WOLFSSH* ssh)
     return ret;
```
aidangarske, line 10734: **🔵 [Low] Unrelated blank-line removal between DoReceive and DoProtoId**

The diff removes a blank line at the top of the gap between `DoReceive` and `DoProtoId`. This change has nothing to do with the service-name check and just adds noise to the diff. Consider keeping unrelated whitespace edits out of feature PRs to keep `git blame` clean.

**Fix:** Revert the stray formatting hunk.

---
> **Note:** Referenced line (`src/internal.c:10719`) is outside the diff hunk. Comment anchored to nearest changed region.

src/internal.c
```diff
@@ -8390,8 +8401,10 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
 #endif
         else {
```
aidangarske, line 8406: **🔵 [Low] Unknown-auth-method `begin = len` consumption path lacks meaningful coverage** · _test-coverage_

The PR adds `begin = len;` (with comment `/* Consume all remaining data */`) in the unknown-auth-method else branch so that `DoUserAuthRequest()` defensively skips any trailing auth-method-specific payload when a request has a valid service name but an unsupported auth method. The new test `test_DoUserAuthRequest_serviceName` (tests/unit.c:953-1048) does exercise this branch via the `valid svc unknown auth` case (service `ssh-connection`, auth method `xyz-unknown`), but the test buffer ends exactly at the end of the auth-method name with no trailing bytes. As a result, `begin` already equals `len` before the new assignment is reached, and the test's `idx != len` assertion would still pass even if the new `begin = len;` line were removed. Because `DoReceive()` advances `inputBuffer.idx` from the handler-reported `payloadIdx` (src/internal.c:10146-10159), the PR still lacks a test proving that an unsupported auth method with extra trailing bytes is fully consumed and leaves packet decoding aligned. A regression that re-introduced an offset bug in this branch would not be caught.

**Fix:** Extend `test_DoUserAuthRequest_serviceName()` with a valid `"ssh-connection"` case that includes an unsupported auth method plus extra trailing bytes (for example, fake password-method fields), then verify `wolfSSH_TestDoUserAuthRequest()` still returns `WS_SUCCESS`, sends `MSGID_USERAUTH_FAILURE`, and advances `idx` to `len`. Without this scenario, the consumption-of-remaining-data behavior added by this PR is not under test.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `audit + review-security`

### Posted findings
- [Low] [audit] **Unknown-auth-method `begin = len` consumption path lacks meaningful coverage** — `src/internal.c:8402-8408`

dgarske:
@yosuke-wolfssl please see the feedback.

src/internal.c
yosuke-wolfssl: Fixed

yosuke-wolfssl:
Hello @aidangarske ,
I'm back with Skoll now and fixed some minor issues.
Could you review this again ?

ejohnstown (approved):

PR #959, opened 2026-04-28
DoProtoId: skip pre-version banner lines

ejohnstown:
1. Client skips non-SSH lines (RFC 4253 4.2); server rejects them
2. 255-byte per-line cap and 10-line cap (WOLFSSH_MAX_BANNER_LINES)
3. 28 test vectors plus scripted-IO mock for WANT_READ resumption

Issue: F-606

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3359,59 +3359,68 @@ retry:
 }
 
 
-static int GetInputText(WOLFSSH* ssh, byte** pEol)
+/* Reads one CRLF- or LF-terminated line into ssh->inputBuffer. On success,
+ * *pEol points to the start of the terminator ('\r' for CRLF, '\n' for LF
+ * only) so the caller can compute line length and terminator size.
+ *
+ * Per RFC 4253 Section 4.2, the SSH version line is at most 255 octets
+ * including the terminator. The same cap is applied to pre-version banner
+ * lines: a line longer than that returns WS_VERSION_E. */
+static int GetInputLine(WOLFSSH* ssh, byte** pEol)
 {
-    int gotLine = 0;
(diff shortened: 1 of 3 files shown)


src/internal.c
```diff
@@ -3359,59 +3359,69 @@ static int ReceiveData(WOLFSSH* ssh, byte* buf, word32 sz)
...
+            if (pEol)
+                *pEol = (byte*)lf;
+            return WS_SUCCESS;
         }
 
-        if (eol)
-            gotLine = 1;
+        /* 255-byte per-line cap: includes the terminator. If the buffer is
+         * already at the cap with no LF, the line is too long. */
+        if (ssh->inputBuffer.length >= 255)
```
padelsbach: Is there a `#define` to use instead of literal 255 to ensure this stays in sync with the surrounding logic?

src/internal.c
```diff
@@ -3359,59 +3359,69 @@ static int ReceiveData(WOLFSSH* ssh, byte* buf, word32 sz)
...
+            return WS_VERSION_E;
 
-    } while (!gotLine && inSz);
+        if (GrowBuffer(&ssh->inputBuffer,
```
padelsbach: Do we need to check if length <= 255 before doing this subtraction?

src/internal.c
ejohnstown: There is now.

src/internal.c
ejohnstown: We are checking for that on line 3400-ish. I rearranged this to calculate inSz first, then use it in GrowBuffer().

padelsbach (approved):

PR #960, opened 2026-04-30
Drop redundant mlkem.h include

ejohnstown:
Drop redundant mlkem.h include. wc_mlkem.h used to include it already. mlkem.h was merged with wc_mlkem.h as of today.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -57,8 +57,7 @@
 #endif
 
 #ifdef WOLFSSL_HAVE_MLKEM
-#include <wolfssl/wolfcrypt/mlkem.h>
-#include <wolfssl/wolfcrypt/wc_mlkem.h>
+    #include <wolfssl/wolfcrypt/wc_mlkem.h>
 #endif
 
 #ifdef NO_INLINE


src/internal.c
```diff
@@ -57,8 +57,10 @@
 #endif
 
 #ifdef WOLFSSL_HAVE_MLKEM
-#include <wolfssl/wolfcrypt/mlkem.h>
-#include <wolfssl/wolfcrypt/wc_mlkem.h>
+    #ifdef HAVE_WOLFSSL_WOLFCRYPT_MLKEM_H
+        #include <wolfssl/wolfcrypt/mlkem.h>
+    #endif
+    #include <wolfssl/wolfcrypt/wc_mlkem.h>
```
padelsbach, line 60: wc_mlkem.h includes wolfssl/wolfcrypt/mlkem.h, so I'm a bit confused about this logic

src/internal.c
```diff
@@ -57,8 +57,10 @@
 #endif
 
 #ifdef WOLFSSL_HAVE_MLKEM
-#include <wolfssl/wolfcrypt/mlkem.h>
-#include <wolfssl/wolfcrypt/wc_mlkem.h>
+    #ifdef HAVE_WOLFSSL_WOLFCRYPT_MLKEM_H
```
padelsbach: in what case(s) do we expect `WOLFSSL_HAVE_MLKEM` to be defined but mlkem.h to be missing?

padelsbach (commented):
Guessing the goal of this PR is to support an older wolfssl without mlkem -- is that right? Is there a specific version that I can try?

src/internal.c
padelsbach: Oh, I see now that this is removed in _newer_ wolfssl

src/internal.c
ejohnstown: wolfSSH v5.9.1. I'm pretty sure it was commit 7a2cf5b65 that broke it for wolfSSH. It got merged this morning.

src/internal.c
ejohnstown: As of this morning, any version going forward.

padelsbach (approved):

src/internal.c
ejohnstown: mlkem.h has been redundant since this was originally added.

src/internal.c
ejohnstown: mlkem.h has been redundant since this was originally added.

PR #962, opened 2026-05-04
Echoserver Channel Regrouping

ejohnstown:
Refactor of the echoserver example only; library is unchanged.

1. Collapse FwdStates and AGENT_STATE_* into a single WS_AppState (INIT/LISTEN/CONNECT/CONNECTED). FWD_STATE_DIRECT becomes an isDirect flag, cleared after connect.
2. Hoist the fields shared by the forwarding, agent, and shell paths into a common WS_AppCtx (privateData, listenFd, appFd, channelId, state, buffer). Shell only uses buffer.
3. Feature-specific remainder stays in WS_AgentCbActionCtx (name) and WS_FwdCbActionCtx (host/origin name+port, isDirect). Drop the unused pid field; inline getpid() at the call site.
4. wolfSSH_AGENT_DefaultActions and wolfSSH_FwdDefaultActions now take a WS_AppCtx* and reach feature-specific state via appCtx->privateData. echoserver_test() is updated to register the WS_AppCtx (not the old CbCtx) with wolfSSH_set_agent_cb_ctx and wolfSSH_SetFwdCbCtx.
5. Reindent ssh_worker()'s #ifdef blocks and add /* WOLFSSH_xxx */ markers.
6. fwd.test.expect: redirect echoserver and portfwd output to log files and sync on portfwd's -R ready file instead of matching "sampled" in the pty. Under --enable-debug the WLOG volume fills the pty and stalls portfwd in select() before channel-open.

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -137,38 +137,42 @@ static int passwdRetry = MAX_PASSWD_RETRY;
 #endif
 
 
+typedef enum WS_AppState {
+    APP_STATE_INIT,
+        /* Just started, not doing anything. */
+    APP_STATE_LISTEN,
+        /* Has listen socket open, waiting for client. */
+    APP_STATE_CONNECT,
+        /* SSH client peer is listening. */
+    APP_STATE_CONNECTED,
+        /* Client application connected, processing its data. */
+} WS_AppState;
+
+
+typedef struct WS_AppCtx {
+    void *privateData;
+    WS_SOCKET_T listenFd;
+    WS_SOCKET_T appFd;
+    word32 channelId;
+    WS_AppState state;
+    byte buffer[EXAMPLE_BUFFER_SZ];
+} WS_AppCtx;
+
+
 #ifdef WOLFSSH_AGENT
 typedef struct WS_AgentCbActionCtx {
     struct sockaddr_un name;
-    WS_SOCKET_T listenFd;
-    WS_SOCKET_T fd;
-    pid_t pid;
-    int state;
 } WS_AgentCbActionCtx;
 #endif
 
 
 #ifdef WOLFSSH_FWD
-enum FwdStates {
-    FWD_STATE_INIT,
-    FWD_STATE_LISTEN,
-    FWD_STATE_CONNECT,
-    FWD_STATE_CONNECTED,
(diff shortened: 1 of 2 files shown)


examples/echoserver/echoserver.c
```diff
@@ -495,36 +499,38 @@ static WS_SOCKET_T connect_addr(const char* name, word16 port)
 static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
         const char* name, word32 port)
 {
-    WS_FwdCbActionCtx* ctx = (WS_FwdCbActionCtx*)vCtx;
+    WS_AppCtx *appCtx = (WS_AppCtx *)vCtx;
+    WS_FwdCbActionCtx* fwdCbCtx = (WS_FwdCbActionCtx *)appCtx->privateData;
```
padelsbach, line 503: Nit: can we use a consistent style for pointer types? Looks like most of the file uses `type* var` rather than `type *var`. Line 503 has both.

scripts/fwd.test.expect
```diff
@@ -45,13 +55,15 @@ set nc_client_pid   ""
...
     global nc_client_pid wolfssh_clt_pid wolfssh_srv_pid nc_server_pid
+    global srv_log clt_log clt_ready
 
     puts "\n--- Cleaning up ---"
     foreach pid [list $nc_client_pid $wolfssh_clt_pid $wolfssh_srv_pid $nc_server_pid] {
         if {$pid ne ""} {
             catch {exec kill $pid}
         }
     }
+    file delete -force $srv_log $clt_log $clt_ready
```
padelsbach, line 66: I believe this deletes the logs unconditionally. Should we keep them in case of failure?

padelsbach (approved):
A couple minor suggestions, but looks good overall

JacobBarthelmeh (approved):

examples/echoserver/echoserver.c
ejohnstown: I agree, but I don't want to change it right now. There's more to change in the file anyway.

scripts/fwd.test.expect
ejohnstown: That script has a little more cleanup. I'll address that later.

PR #963, opened 2026-05-06
rekey: trigger highwater on per-key packet count

ejohnstown:
- Track txMsgCount/rxMsgCount per key epoch and reset on NEW_KEYS; seq/peerSeq still wrap freely per RFC 4253 Sec 6.4.
- Extend HighwaterCheck to fire highwaterCb when packet count crosses msgHighwaterMark (default 2^31, RFC 4344 Sec 3.1).
- Add wolfSSH_CTX_SetMsgHighwater / SetMsgHighwater / GetMsgHighwater.

Issue: F-246

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -188,6 +188,9 @@ Flags:
     Set the number of Miller-Rabin rounds used when the client checks the
     server's prime group when using GEX key exchange. The default is 8. More
     rounds are better, but also takes a lot longer.
+  WOLFSSH_DEFAULT_MSG_HIGHWATER_MARK
+    Set the default value for the number of messages to send or receive before
+    calling the highwater callback function. By default this forces a rekey.
 */
 
 static const char sshProtoIdStr[] = "SSH-2.0-wolfSSHv"
@@ -544,7 +547,10 @@ static int HashUpdate(wc_HashAlg* hash, enum wc_HashType type,
 static INLINE int HighwaterCheck(WOLFSSH* ssh, byte side)
 {
     int ret = WS_SUCCESS;
+    int fire = 0;
 
+    /* RFC 4253 Sec 9: bound bytes per key (txCount/rxCount reset on rekey)
+     * to limit cipher keystream/IV exhaustion under a single key. */
     if (!ssh->highwaterFlag && ssh->highwaterMark &&
         (ssh->txCount >= ssh->highwaterMark ||
          ssh->rxCount >= ssh->highwaterMark)) {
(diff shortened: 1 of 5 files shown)


src/internal.c
```diff
@@ -553,10 +559,26 @@ static INLINE int HighwaterCheck(WOLFSSH* ssh, byte side)
...
+             (side == WOLFSSH_HWSIDE_TRANSMIT) ? "Transmit" : "Receive");
 
-        if (ssh->ctx->highwaterCb)
-            ret = ssh->ctx->highwaterCb(side, ssh->highwaterCtx);
+        ssh->msgHighwaterFlag = 1;
+        fire = 1;
     }
+
+    if (fire && ssh->ctx->highwaterCb)
+        ret = ssh->ctx->highwaterCb(side, ssh->highwaterCtx);
```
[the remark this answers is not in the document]
ejohnstown: I'm OK with this.

wolfssh/internal.h
```diff
@@ -443,6 +443,12 @@ enum NameIdType {
 #ifndef DEFAULT_HIGHWATER_MARK
     #define DEFAULT_HIGHWATER_MARK ((1024 * 1024 * 1024) - (32 * 1024))
 #endif
+#ifndef WOLFSSH_DEFAULT_MSG_HIGHWATER_MARK
+    /* RFC 4344 Sec 3.1: bound packets per key epoch to limit cipher/IV
+     * exhaustion under a single key (not the absolute SSH sequence number,
+     * which is not reset by rekey); default to 2^31 packets per key. */
+    #define WOLFSSH_DEFAULT_MSG_HIGHWATER_MARK (1U << 31)
```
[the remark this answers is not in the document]
ejohnstown: Used 0x80000000 instead.

src/internal.c
```diff
@@ -553,10 +559,26 @@ static INLINE int HighwaterCheck(WOLFSSH* ssh, byte side)
...
+
+    /* RFC 4344 Sec 3.1: bound packets per key (txMsgCount/rxMsgCount reset on
+     * rekey) to limit cipher/IV exhaustion under a single key; this is not a
+     * guard for the absolute SSH sequence number, which never resets. */
+    if (!ssh->msgHighwaterFlag && ssh->msgHighwaterMark &&
+        (ssh->txMsgCount >= ssh->msgHighwaterMark ||
+         ssh->rxMsgCount >= ssh->msgHighwaterMark)) {
+
+        WLOG(WS_LOG_DEBUG, "%s over msg high water mark",
+             (side == WOLFSSH_HWSIDE_TRANSMIT) ? "Transmit" : "Receive");
```
padelsbach, line 574: `side` is only used for the log in this func. Seems the callback will be called even if `side` is not what the caller specified. Not sure if this is a problem

wolfssh/ssh.h
```diff
@@ -78,14 +78,19 @@ WOLFSSH_API int wolfSSH_SetFilesystemHandle(WOLFSSH*, void*);
...
-                                        WS_CallbackHighwater);
-WOLFSSH_API void wolfSSH_SetHighwaterCtx(WOLFSSH*, void*);
-WOLFSSH_API void* wolfSSH_GetHighwaterCtx(WOLFSSH*);
+WOLFSSH_API int wolfSSH_SetHighwater(WOLFSSH* ssh, word32 highwater);
+WOLFSSH_API word32 wolfSSH_GetHighwater(WOLFSSH* ssh);
+
+/* packet count high water mark functions (RFC 4344 Sec 3.1) */
+WOLFSSH_API int wolfSSH_CTX_SetMsgHighwater(WOLFSSH_CTX* ctx, word32 highwater);
+WOLFSSH_API int wolfSSH_SetMsgHighwater(WOLFSSH* ssh, word32 highwater);
+WOLFSSH_API word32 wolfSSH_GetMsgHighwater(WOLFSSH* ssh);
```
padelsbach: would it make sense to have a symmetric `wolfSSH_CTX_GetMsgHighwater`?

padelsbach:
can you add test cases for the new public api's?

src/internal.c
ejohnstown: It is also passed to the callback function. I don't think it is useful at this point. But, it is in the ABI now.

wolfssh/ssh.h
ejohnstown: I don't think so. In the years since I wrote it, I was beginning to this "It was your code that set it. Your code knows what it is."

padelsbach (approved):

PR #965, opened 2026-05-07
Static-analysis fixes

ejohnstown:
- src/internal.c (F-413): DoDisconnect/DoUnimplemented now use GetUint32() instead of raw ato32() so short payloads return WS_BUFFER_E instead of reading past the buffer. DoDisconnect also parses the description and language-identifier strings with GetSkip(), so a record truncated after the reason code returns WS_BUFFER_E instead of being treated as success.
- apps/wolfssh/wolfssh.c (F-214): config_print was printing keyFile twice; fixed to print pubKeyFile.
- src/port.c (F-50): WS_MoveFileA leaked unicodeOldName on two error paths; now freed before returning.
- apps/wolfsshd/auth.c (F-57): CheckPasswordUnix now ForceZeros storedHashCpy before freeing.

Issues: F-50, F-57, F-214, F-413

### DIFF
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -920,7 +920,7 @@ static int config_print(struct config* config)
         printf("port %u\n", config->port);
         printf("keyFile %s\n", config->keyFile ? config->keyFile : "none");
         printf("pubKeyFile %s\n",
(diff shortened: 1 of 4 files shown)


yosuke-wolfssl (commented):
everything looks good

padelsbach (approved):

PR #973, opened 2026-05-12
Fix wolfSSH_SFTPNAME_readdir with FATFS

J-Goldacker:
This PR fixes the following issues for wolfSSH_SFTPNAME_readdir with FATFS:

- f_readdir returns FR_OK with empty string in fname when there is no directory entry left
- wolfSSH_SFTPNAME_free would crash because of a double free in case wolfSSH_SFTPNAME_readdir returns an error and frees fName without setting it to NULL

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -3152,12 +3152,14 @@ static int wolfSSH_SFTPNAME_readdir(WOLFSSH* ssh, WDIR* dir, WS_SFTPNAME* out,
                 >= (int)sizeof(r)) {
             WLOG(WS_LOG_SFTP, "Path length too large");
             WFREE(out->fName, out->heap, DYNTYPE_SFTP);
+            out->fName = NULL;
             return WS_FATAL_ERROR;
         }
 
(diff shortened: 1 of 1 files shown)


dgarske:
Hi @J-Goldacker , no contributor agreement on file. Assigning to @ejohnstown to review and make recommendation.

J-Goldacker:
Hi @dgarske, thank you for assigning the PR.

Is this the correct way the give bugfixes we found back to you? (We have licensed your software for Hoppecke Systemtechnik GmbH.)

@ejohnstown One thing, I wanted to mention: 
The double free problem for fname can occur on other places too. We only fixed what we needed to. Another approach would be to not free fname on error and rely on the fact, that wolfSSH_SFTPNAME_free is called after an error is returned. (But we chose the safer approach.)

ejohnstown:
> Is this the correct way the give bugfixes we found back to you?

This is fine. In this case, since there were 14 other locations to change, telling us about the bug and then letting us fix it would be better, but that is mostly because of it only being one PR instead of two. Macht nichts. : ]

Thank you for the fix and the bug report.

ejohnstown (approved):

PR #974, opened 2026-05-12
Fix SFTP_GetAttributes with fatfs for root

J-Goldacker:
After updating Filezilla to 3.70.5 (from 3.69.1) we got an error message during reception of the directory content. Connecting to a home directory not being root works. (This took us a while to find out.)

We tracked it down to SFTP_GetAttributes where f_stat is called for "/". f_stat doesn't support this and returns FR_INVALID_NAME. Interestingly, the code handling this is mainly there. But the function returns too early.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -4861,9 +4861,6 @@ static int SFTP_GetAttributes(void* fs, const char* fileName,
     (void) noFollow;
     (void) heap;
 
-    ret = f_stat(fileName, &info);
-    if (ret != FR_OK)
-        return -1;
     WMEMSET(atr, 0, sizeof(WS_SFTP_FILEATRB));
     if (sz > 2 && fileName[sz - 2] == ':') {
         atr->flags |= WOLFSSH_FILEATRB_PERM;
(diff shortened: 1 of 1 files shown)


dgarske:
Hi @J-Goldacker , I don't see you on our approved contributor list. Can you tell us more about your project and use of wolfSSH? For such a small change we may treat this as a bug report. I've assigned this PR to @ejohnstown who is the project owner and can review and decide. Thanks

J-Goldacker:
Hi @dgarske, thank you for assigning the PR. 

Is this the correct way the give bugfixes we found back to you? (We have licensed your software for Hoppecke Systemtechnik GmbH.)

We are using WolfSSH in our BMS (HW: STM32F4, Ethernet, SD Card / SW: STM32CubeHAL, FreeRTOS, FATFS, ...).
Goal is to have secured SFTP access to our SD Card. This is what we use WolfSSH for. This PR is one of the changes, we always need to patch into WolfSSH during build in order to use it.

The problem seems to be specific to newer versions (3.70.x) of FileZilla (which we want to use). Other tools like WinSCP and PSFTP (and FileZilla 3.69.1) don't read the attributes for "/".

Do you need more information? Let me know.

ejohnstown (approved):

PR #976, opened 2026-05-12
Static analysis fixes

ejohnstown:
- Pin expectMsgId to MSGID_KEXDH_GEX_INIT after KEXDH_GEX_GROUP; null-check ssh->handshake (F-3668).
- Reject ParseRFC6187 name lengths that would overflow when added to sizeof(word32), preventing OOB reads (F-1277).
- ForceZero the lastBlock stack buffers on both remainder paths in GenerateKey (F-2494).
- ForceZero newKey before free in DoPemKey/DoOpenSshKey and the PEM read buffer in wolfSSH_ReadKey_file (F-2885, F-3211, F-3212).

### DIFF
--- a/apps/wolfssh/common.c
+++ b/apps/wolfssh/common.c
@@ -144,6 +144,9 @@ static int ParseRFC6187(const byte* in, word32 inSz, byte** leafOut,
 
     /* Skip the name */
     ato32(in, &l);
+    if (l > inSz - sizeof(word32))
+        return WS_BUFFER_E;
+
     m += l + sizeof(word32);
 
     /* Get the cert count */
--- a/src/internal.c
+++ b/src/internal.c
@@ -2584,6 +2584,7 @@ int GenerateKey(byte hashId, byte keyId,
                 ret = wc_HashFinal(&hash, enmhashId, lastBlock);
                 if (ret == WS_SUCCESS)
                     WMEMCPY(key, lastBlock, remainder);
+                ForceZero(lastBlock, sizeof(lastBlock));
             }
         }
         else {
@@ -2629,6 +2630,7 @@ int GenerateKey(byte hashId, byte keyId,
                     ret = wc_HashFinal(&hash, enmhashId, lastBlock);
                 if (ret == WS_SUCCESS)
                     WMEMCPY(key + runningKeySz, lastBlock, remainder);
+                ForceZero(lastBlock, sizeof(lastBlock));
             }
         }
     }
--- a/src/internal.c
+++ b/src/internal.c
@@ -13476,7 +13476,7 @@ int SendKexDhGexGroup(WOLFSSH* ssh)
     int ret = WS_SUCCESS;
 
     WLOG(WS_LOG_DEBUG, "Entering SendKexDhGexGroup()");
-    if (ssh == NULL)
+    if (ssh == NULL || ssh->handshake == NULL)
         ret = WS_BAD_ARGUMENT;
 
     if (ret == WS_SUCCESS) {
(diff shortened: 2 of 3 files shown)


apps/wolfssh/common.c
```diff
@@ -144,6 +144,9 @@ static int ParseRFC6187(const byte* in, word32 inSz, byte** leafOut,
 
     /* Skip the name */
     ato32(in, &l);
+    if (l > inSz - sizeof(word32))
```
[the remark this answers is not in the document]
ejohnstown: This is already checked for. The change in insolation could be mistaken for an issue.

src/internal.c
```diff
@@ -13521,8 +13523,11 @@ int SendKexDhGexGroup(WOLFSSH* ssh)
         ret = BundlePacket(ssh);
     }
 
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS) {
+        WLOG_EXPECT_MSGID(MSGID_KEXDH_GEX_INIT);
+        ssh->handshake->expectMsgId = MSGID_KEXDH_GEX_INIT;
         ret = wolfSSH_SendPacket(ssh);
```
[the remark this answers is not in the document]
ejohnstown: The pattern used throughout internal.c is to set the expected message ID, then send the current message. I'm not going to change this. If the send fails, the connection is ending anyway.

src/ssh.c
```diff
@@ -1868,7 +1868,7 @@ static int DoPemKey(const byte* in, word32 inSz, byte** out,
     WOLFSSH_UNUSED(heap);
 
     if (*out == NULL) {
-        newKey = (byte*)WMALLOC(inSz, heap, DYNTYPE_PRIVKEY);
+        newKey = (byte*)WMALLOC(newKeySz, heap, DYNTYPE_PRIVKEY);
```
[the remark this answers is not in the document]
ejohnstown: The variable newKeySz is initialized to inSz. The key data coming in is in PEM format, and PEM is always bigger than the associated DER format key. In the case where newKeySz is set to outSz, we aren't allocating a buffer.

src/ssh.c
```diff
@@ -1915,6 +1916,7 @@ static int DoPemKey(const byte* in, word32 inSz, byte** out,
     else {
         WLOG(WS_LOG_DEBUG, "Unable to identify PEM key");
         if (*out == NULL) {
+            ForceZero(newKey, newKeySz);
```
padelsbach: Looks like `newKeySz` might have changed (shrank?) since the original `WMALLOC`. Should we erase the original allocated `*inSz`?

ejohnstown: I've had some arguments with Copilot about this. There is a comment that the binary copy of the key will be smaller than the PEM copy. newKeySz is getting reset to the actual size of the binary copy of the certificate. Using inSz originally to allocate the buffer was intentional. It is guaranteed that the binary copy will be smaller than the text copy, but it isn't exactly half the size. We did something similar in wolfSSL reading certs. In the above fragment, we are only zeroing out the actual key in the buffer.

src/ssh.c
```diff
@@ -1879,6 +1879,7 @@ static int DoPemKey(const byte* in, word32 inSz, byte** out,
             return WS_BUFFER_E;
         }
         newKey = *out;
+        newKeySz = *outSz;
```
padelsbach, line 1882: This changes how `newKeySz` is set altogether. Seems more correct now, but still a change of the behavior beyond the zeroization. These are passed unmodified via a public API. Seems risky.

src/ssh.c
ejohnstown: The issue I think I see here is the ForceZero() should go outside the if clause. In either case, this function makes the buffer or the buffer is provided, if the Base64Decode fails, we should erase the partially filled buffer of possibly private key data. Which was the point of adding the ForceZero().

src/ssh.c
ejohnstown: We are either using the provided buffer or are making the buffer to return. This case is when the caller provides the output buffer to use. newKeySz is the size of the newKey's buffer in either case. Next thing it does is convert the PEM key file into DER, which will make it smaller. newKeySz is updated with the actual size of the key. We are sanity checking the size of the output buffer against the size of the incoming PEM encoded key.

I still believe the size used in the ForceZero() is correct.

padelsbach (dismissed):

padelsbach (approved):

Issue #975, opened 2026-05-12
SFTP Timestamp incorrect for FATFS


J-Goldacker:
https://github.com/wolfSSL/wolfssh/blob/6e6995980e7d7462bb0d081cddf2ade17a3a4f46/src/wolfsftp.c#L4958-L4959

The problem is here. FATFS time info has it's own format. The solution for NUCLEUS (function TimeTo32(d, t)) looks very similar. Maybe it is the same, but I could not test it yet.

SFTP_GetAttributes_Handle seems to have the same problem.

embhorn:
Hi @J-Goldacker 

Thanks for contacting wolfSSL Support. Could you tell us a bit about your project using wolfSSH? I've requested a review of this issue by our engineers.

Kind regards,
@embhorn - wolfSSL Support

ejohnstown:
Does the RTOS you are running on have the function mktime()?

ejohnstown:
I have a fix posted as a PR. I haven't tested the date conversion yet. That's why the PR is a draft. Please let me know if that works for you.

J-Goldacker:
Hi @ejohnstown, thank you very much. I've left two comments in your PR. 

Yes, we have mktime.

I'll try the changes on the target and give you feedback.

Edit: I've ran the code. With the change from my comment (add 80 to the year) it works
`#define WS_GETYEAR(d) ((((d) >> 9) & 0x007f) + 80)`

ejohnstown:
I have updated the PR. Thank you for the feedback.

ejohnstown:
PR #977 is merged. Thanks!

J-Goldacker:
Thank you for solving this so quickly.

PR #977, opened 2026-05-12
Convert FatFS date/time to Unix timestamp in SFTP

ejohnstown:
- Add SetAttrTime() helper to decode fdate/ftime into a Unix timestamp via mktime() and assigns it to atr->atime/mtime.
- Use it from both SFTP_GetAttributes() (was using raw fdate) and SFTP_GetAttributes_Handle() (was using raw ftime, losing the date).

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -4850,6 +4850,61 @@ static int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
 
 #elif defined(WOLFSSH_FATFS)
 
+/*
+ * fdate
+ * The date when the file was modified or the directory was created.
+ *     bit15:9
+ *         Year origin from 1980 (0..127)
+ *     bit8:5
+ *         Month (1..12)
+ *     bit4:0
+ *         Day (1..31)
+ *
+ * ftime
+ * The time when the file was modified or the directory was created.
+ *     bit15:11
+ *         Hour (0..23)
+ *     bit10:5
+ *         Minute (0..59)
+ *     bit4:0
+ *         Second / 2 (0..29)
+ *
+ * mktime() expects month from 0 to 11 and years starting at
+ * 1900. FatFS months are saved as 1 to 12 and the years
+ * start counting at 1980.
+ */
+#define FATFS_GETDAY(d) ((d) & 0x001f)
+#define FATFS_GETMON(d) ((((d) >> 5) & 0x000f) - 1)
+#define FATFS_GETYEAR(d) ((((d) >> 9) & 0x007f) + 80)
+#define FATFS_GETHOUR(t) (((t) >> 11) & 0x001f)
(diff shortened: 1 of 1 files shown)


ejohnstown:
This should hopefully fix issue #975.

src/wolfsftp.c
```diff
@@ -4850,6 +4850,54 @@ static int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
...
+ *     bit4:0
+ *         Second / 2 (0..29)
+ *
+ * mktime() expects month from 0 to 11. FatFS months
+ * are saved as 1 to 12. Hence 1 is being deducted to
+ * make it compatible with Unix time stamp.
+ */
+#define WS_GETDAY(d) ((d) & 0x001f)
+#define WS_GETMON(d) ((((d) >> 5) & 0x000f) - 1)
+#define WS_GETYEAR(d) (((d) >> 9) & 0x007f)
```
J-Goldacker: I think you have to add 80, because FATFS starts 1980 and tm_year starts 1900.

src/wolfsftp.c
J-Goldacker: I think TimeTo32_FatFS should be used here, too.

src/wolfsftp.c
```diff
@@ -4850,6 +4850,61 @@ static int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
...
+ *     bit10:5
+ *         Minute (0..59)
+ *     bit4:0
+ *         Second / 2 (0..29)
+ *
+ * mktime() expects month from 0 to 11 and years starting at
+ * 1900. FatFS months are saved as 1 to 12 and the years
+ * start counting at 1980.
+ */
+#define FATFS_GETDAY(d) ((d) & 0x001f)
```
padelsbach, line 4876: Consider using a bitfield instead of manual bit arithmetic. Much more readable imho

src/wolfsftp.c
```diff
@@ -4850,6 +4850,61 @@ static int SFTP_GetAttributes_Handle(WOLFSSH* ssh, byte* handle, int handleSz,
...
+#define FATFS_GETMON(d) ((((d) >> 5) & 0x000f) - 1)
+#define FATFS_GETYEAR(d) ((((d) >> 9) & 0x007f) + 80)
+#define FATFS_GETHOUR(t) (((t) >> 11) & 0x001f)
+#define FATFS_GETMIN(t)  (((t) >> 5 ) & 0x003f)
+#define FATFS_GETSEC(t)  (((t) << 1 ) & 0x003f)
+
+static void SetAttrTime(const FILINFO* info, WS_SFTP_FILEATRB* atr)
```
padelsbach, line 4883: Existing function `TimeTo32` is used for similar purposes it seems, but has a much different function declaration. Consider matching it

padelsbach (approved):

PR #989, opened 2026-05-22
X.509 Interop Test

ejohnstown:
- Add a GitHub Actions workflow that builds PKIX-SSH and runs wolfSSHd against the PKIX-SSH ssh/sftp clients to exercise x509 user-certificate auth; runs weekly and on workflow_dispatch.
- Gate wolfSSH/wolfSSL debug logging in wolfsshd behind the `-d` flag, enabling it on startup and disabling both facilities during cleanup.

### DIFF
--- a/.github/workflows/x509-interop.yml
+++ b/.github/workflows/x509-interop.yml
@@ -1,0 +1,253 @@
+name: wolfSSH x509 Interop Test
+
+on:
+  schedule:
+    # Weekly: Mondays at 06:00 UTC
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  WOLFSSL_REF: v5.9.1-stable
+  PKIXSSH_VERSION: 14.4
+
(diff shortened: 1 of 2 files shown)


.github/workflows/x509-interop.yml
```diff
@@ -0,0 +1,251 @@
...
+          ../build-dir/bin/ssh -o StrictHostKeyChecking=accept-new \
+            -o PreferredAuthentications=publickey \
+            -p 22222 -F ssh-pkixssh-config \
+            -i ./keys/fred-key.pem fred@127.0.0.1 exit
+
+      - name: Test PKIX-SSH client ls command
+        working-directory: ./wolfssh/
+        run: |
+          ../build-dir/bin/ssh -o StrictHostKeyChecking=accept-new \
+            -p 22222 -F ssh-pkixssh-config \
```
padelsbach, line 187: Should we use `-o PreferredAuthentications=publickey \` for the ls and sftp commands?

ejohnstown: Copypaste oversight. Didn't affect the test, but good to have it.

padelsbach (approved):

PR #1002, opened 2026-06-04
tidy clang-tidy

ejohnstown:
Clean up clang-tidy findings across wolfSSH. The checks set initially flagged 378 issues across six categories; this branch clears all of them. Build is clean (exit 0); no functional or wire-protocol changes.

- Add parameter names to function declarations -- name the parameters in public/internal-header function decls and the few forward decls in .c files; clears all 360 readability-named-parameter findings. Prototypes that exceeded 80 columns after the names were added are re-wrapped using the existing 8-space continuation style.
- Remove redundant preprocessor guard -- drop the inner #ifdef WOLFSSH_CERTS inside DoUserAuthRequestRsaCert (the surrounding function is already under the same guard); clears readability-redundant-preprocessor.
- Move unsafe call out of signal handler -- drop fprintf from wolfsshd's interruptCatch and log "Closing down wolfSSHD" from the main accept loop after quit is observed; annotate sftpclient's wolfSSH_SFTP_Interrupt call with NOLINT since it is a single byte store; clears both bugprone-signal-handler findings.
- Use reentrant time conversion in SFTP -- SFTP_CreateLongName switches from XGMTIME (gmtime) to WLOCALTIME, which expands to localtime_r/localtime_s per platform; clears concurrency-mt-unsafe. Server will provide localtimes for ls -l output, except for Zephyr which only provides UTC.
- Tighten SCP timestamp bounds check -- in ParseTimestamp, move ++idx out of the || conditions and tighten the bound from > bufSz to >= bufSz; clears both bugprone-inc-dec-in-conditions findings and fixes a one-past-the-end read.
- Bound sshd Include directive recursion -- track include depth on WOLFSSHD_CONFIG, reject loads past WOLFSSHD_MAX_INCLUDE_DEPTH (16), and annotate the four functions in the include cycle with NOLINTNEXTLINE pointing at the bound; clears all 8 misc-no-recursion findings. Adds a test for the recursion depth.


src/wolfsftp.c
```diff
@@ -2723,15 +2724,17 @@ static int SFTP_CreateLongName(WS_SFTPNAME* name)
...
-#if defined(XGMTIME) && defined(XSNPRINTF)
+#if defined(XSNPRINTF)
     atr = &name->atrb;
 
-    /* get date as calendar date */
-    localTime = XGMTIME((const time_t*)&atr->mtime, &localTime);
-    if (localTime == NULL) {
+    /* get date as calendar date; use the reentrant WLOCALTIME
+     * wrapper so the shared static buffer behind gmtime() is
+     * not used. */
```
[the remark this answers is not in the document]
ejohnstown: We explain already we are using the reentrant localtime.

apps/wolfsshd/configuration.c
padelsbach, line 930: AI tells me `wolfSSHD_ConfigCopy` does not copy the `includeDepth` param, and thus could be wiped out here, partially defeating the purpose of this commit. Should we copy that field?

src/wolfsftp.c
```diff
@@ -2777,15 +2778,17 @@ static int SFTP_CreateLongName(WS_SFTPNAME* name)
...
+    mtime = (time_t)atr->mtime;
+    if (!WLOCALTIME(&mtime, &localTime)) {
```
padelsbach, line 3014: This switches from UTC time to local time in all platforms except Zephyr which uses `gmtime_r` (in port.h). Is this ok?

apps/wolfsshd/configuration.c
ejohnstown: Moved the depth counter out of the structure. It is now a parameter to a few of the functions.

src/wolfsftp.c
ejohnstown: The times should be shown in localtime, not UTC. The exception is Zephyr as its localtime_r is a wrapper around gmtime_r.

wolfssh/wolfsftp.h
```diff
@@ -139,6 +139,7 @@ struct WS_SFTP_FILEATRB_EX {
 typedef struct WS_SFTP_FILEATRB {
     word32 flags;
     word32 sz[2]; /* sz[0] being the lower and sz[1] being the upper */
+    word32 nlink; /* number of hardlinks */
```
dgarske: **🟠 [Medium] New nlink member inserted mid-struct in public WS_SFTP_FILEATRB breaks ABI**

`nlink` is inserted between `sz[2]` and `uid` in the public `WS_SFTP_FILEATRB` struct, shifting the offsets of `uid`, `gid`, `per`, `atime`, `mtime`, `extCount`, and `exts`. This is a public struct, and the project convention (RULES.md C coding style) states: 'New members at the end (mid-struct insertion breaks ABI).' Pre-compiled consumers linking against an older layout would read the wrong offsets. Source-recompiled builds are unaffected, but the convention is violated. (The field is consumed in src/wolfsftp.c:2903, so this is directly tied to the functional change.)

**Fix:** Move `nlink` to the end of the public struct to preserve ABI per the project convention. Confirm no raw-struct serialization relies on the new offset.

src/wolfsftp.c
```diff
@@ -2899,9 +2898,9 @@ static int SFTP_CreateLongName(WS_SFTPNAME* name)
     name->lSz = totalSz;
     name->lName[totalSz] = '\0';
 
-#if defined(XGMTIME) && defined(XSNPRINTF)
-    WSNPRINTF(name->lName, totalSz, "%s %3d %8d %8d %s %s %s",
-            perm, linkCount, atr->uid, atr->gid, sizeStr, date, name->fName);
+#if defined(WSNPRINTF)
+    WSNPRINTF(name->lName, totalSz, "%s %3u %8u %8u %s %s %s",
+            perm, atr->nlink, atr->uid, atr->gid, sizeStr, date, name->fName);
```
dgarske: **🔵 [Low] nlink reserved width (3) can truncate the SFTP long name for high link counts**

The previous code formatted a constant `linkCount = 1` with `%3d`, which always produced exactly 3 characters, matching the `totalSz += 3 + 8 + 8` reservation. This PR replaces it with `atr->nlink` (a real `st_nlink` value via the new POSIX helper) formatted as `%3u`. `%3u` is a *minimum* width, so a directory with many sub-directories (st_nlink > 999, common on busy servers) emits 4+ digits and exceeds the 3 bytes reserved. The buffer only has 1 byte of slack (7 spaces reserved vs 6 emitted), and `uid`/`gid` (`%8u`, word32 up to 10 digits) can already consume that slack. Because the write is bounded by `WSNPRINTF(name->lName, totalSz, ...)` there is no overflow, but the trailing `name->fName` in the `ls -l` output can be silently truncated. This is a behavioral regression versus the old constant `linkCount`.

**Fix:** Compute the reserved size from the actual decimal widths of nlink/uid/gid (or bump the fixed reservation) so the filename is never truncated for large link/uid/gid values.

src/wolfsftp.c
```diff
@@ -2782,7 +2782,7 @@ int wolfSSH_SFTP_RecvOpenDir(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
 
 #define WS_DATE_SIZE 12
 
-#if defined(XGMTIME) && defined(XSNPRINTF)
+#if defined(WSNPRINTF)
```
dgarske, line 2953: **🔵 [Low] longname/localtime path now compiled on platforms that previously used the fallback**

The compile guard changed from `#if defined(XGMTIME) && defined(XSNPRINTF)` to `#if defined(WSNPRINTF)`, and the body now calls `WLOCALTIME` instead of `XGMTIME`. `WSNPRINTF` is defined far more broadly than `XGMTIME` was (it is defined in every non-WSTRING_USER branch of port.h), so this enables the full timestamp/longname path on platforms (e.g. Nucleus) that previously fell through to the simple `WMEMCPY(name->lName, name->fName, ...)` branch. `WLOCALTIME` has an unconditional `#else` fallback so it always compiles, but its per-platform definitions differ in argument order/semantics (e.g. the Nucleus `localtime_s((c),(r))` and Harmony double-`localtime` macros). The main targets (Linux/Windows/MQX/Zephyr) are correct; this is worth a sanity check on the newly-enabled embedded platforms.

**Fix:** Confirm the longname path behaves correctly on platforms newly brought into the `WSNPRINTF` guard (especially Nucleus/Harmony WLOCALTIME argument order), or keep a tighter guard if those platforms should retain the simple fallback.

wolfssh/ssh.h
```diff
@@ -206,32 +206,44 @@ typedef enum WS_FwdCbError {
...
-
-WOLFSSH_API int wolfSSH_ChannelFree(WOLFSSH_CHANNEL*);
-WOLFSSH_API int wolfSSH_ChannelGetId(WOLFSSH_CHANNEL*, word32*, byte);
-WOLFSSH_API WOLFSSH_CHANNEL* wolfSSH_ChannelFind(WOLFSSH*, word32, byte);
-WOLFSSH_API WOLFSSH_CHANNEL* wolfSSH_ChannelNext(WOLFSSH*, WOLFSSH_CHANNEL*);
-WOLFSSH_API int wolfSSH_ChannelRead(WOLFSSH_CHANNEL*, byte*, word32);
-WOLFSSH_API int wolfSSH_ChannelSend(WOLFSSH_CHANNEL*, const byte*, word32);
-WOLFSSH_API int wolfSSH_ChannelExit(WOLFSSH_CHANNEL*);
-WOLFSSH_API int wolfSSH_ChannelGetEof(WOLFSSH_CHANNEL*);
+typedef int (*WS_CallbackFwd)(WS_FwdCbAction action, void* fwdCbCtx,
```
dgarske, line 209: **⚪ [Info] WS_CallbackFwdIO parameter naming is inverted relative to the analogous WS_CallbackAgentIO**

The new parameter names for `WS_CallbackFwdIO` are `(action, void* fwdCbCtx, word32 bufSz, void* buf)` — i.e. 2nd arg = ctx, 4th arg = buffer. The analogous and actually-invoked `WS_CallbackAgentIO` (agent.h) is `(action, void* buf, word32 bufSz, void* agentCbCtx)` — 2nd arg = buffer, 4th arg = ctx. `fwdIoCb` is never invoked in the tree, so intent can't be confirmed from call sites, but the inverted naming will likely mislead anyone implementing the forward-IO callback. Names in a typedef are cosmetic (no behavioral effect).

**Fix:** Align the FwdIO parameter ordering/names with the AgentIO convention (data buffer early, ctx last) to avoid misleading callback implementers.

wolfssh/port.h
```diff
@@ -1487,8 +1495,10 @@ extern "C" {
 
     #define WOPEN(fs,f,m,p) open((f),(m),(p))
     #define WCLOSE(fs,fd) close((fd))
-    int wPwrite(WFD, unsigned char*, unsigned int, const unsigned int*);
-    int wPread(WFD, unsigned char*, unsigned int, const unsigned int*);
+    int wPwrite(WFD, unsigned char* buf, unsigned int sz,
```
dgarske: **⚪ [Info] wPwrite/wPread leave first param WFD unnamed in one platform block**

This PR names parameters in the wPwrite/wPread declarations across three platform blocks (`WFD fd, unsigned char* buf, ...`) but in one block the first parameter is left as bare `WFD` while the others gained names. This is an incomplete/inconsistent application of the readability-named-parameter cleanup the PR is intended to perform; the named-parameter check would still flag the unnamed `WFD`.

**Fix:** Name the first `WFD` parameter `WFD fd` to match the other three platform blocks and fully satisfy the named-parameter check.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **New nlink member inserted mid-struct in public WS_SFTP_FILEATRB breaks ABI** — `wolfssh/wolfsftp.h:142`
- [Low] **nlink reserved width (3) can truncate the SFTP long name for high link counts** — `src/wolfsftp.c:2886-2903`
- [Low] **longname/localtime path now compiled on platforms that previously used the fallback** — `src/wolfsftp.c:2785-2845`
- [Info] **WS_CallbackFwdIO parameter naming is inverted relative to the analogous WS_CallbackAgentIO** — `wolfssh/ssh.h:209-211`
- [Info] **wPwrite/wPread leave first param WFD unnamed in one platform block** — `wolfssh/port.h:1498-1500`

wolfssh/wolfsftp.h
ejohnstown: Fixed.

src/wolfsftp.c
ejohnstown: Removing the nlink value as a part of the longname construction. nlink isn't a part of the file attrs described in the SFTP draft RFC. OpenSSH just prints a ? for that value.

src/wolfsftp.c
ejohnstown: Nucleus and Harmony are already using WLOCALTIME in log.c. It is available to them. This isn't pioneering a new behavior, it is expanding to a couple other RTOSes we already support. The `#else` case in longname is there covering for snprintf not being available in some other unsupported RTOS. I'm doing something a little different here to cover where WLOCALTIME may be missing, and handling where WSNPRINTF is missing.

wolfssh/port.h
ejohnstown: Fixed.

wolfssh/ssh.h
ejohnstown: Fixed.

dgarske (approved):

PR #1003, opened 2026-06-05
wolfsshd: honor Match-block auth restrictions in RequestAuthentication

yosuke-wolfssl:
This fixes auth-bypass where Match-block authentication restrictions were silently ignored. RequestAuthentication and DoCheckUser read authentication settings from the global root config node (authCtx->conf), which is assigned once in wolfSSHD_AuthCreateUser and never refreshed per-user. As a result, a restriction such as:
Match User alice
    PasswordAuthentication no
was correctly advertised by DefaultUserAuthTypes (pubkey-only) but not enforced — the server still accepted alice's password attempt. The same stale-pointer flaw also bypassed per-user PermitRootLogin.

**Changes**

- RequestAuthentication: resolve the per-user config once via wolfSSHD_AuthGetUserConf and use it for all five settings that were reading the global node — PasswordAuthentication, PermitEmptyPasswords, AuthorizedKeysFile (set flag), and the two TrustedUserCAKeys reads. This mirrors the correct pattern already used in DefaultUserAuthTypes.
- DoCheckUser: the root-login check now resolves the per-user config so a Match block override of PermitRootLogin is honored instead of only consulting the global node. Scoped to the root branch.
- Both new lookups fail closed: if the per-user config cannot be resolved (e.g. the user's primary group is unresolvable and getgrgid fails inside wolfSSHD_AuthGetUserConf), authentication is denied rather than falling back to the permissive global node. This is safe because such a user cannot complete a session anyway — session setup already rejects an unresolvable user config with WS_FATAL_ERROR. Comments explain this reasoning for future maintainers.
- New tests to exercise the new code path

Addressed by f_4793 and f_4943

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -463,18 +463,14 @@ static int CheckPasswordUnix(const char* usr, const byte* pw, word32 pwSz, WOLFS
 
 
 static const char authKeysDefault[] = ".ssh/authorized_keys";
-static char authKeysPattern[MAX_PATH_SZ] = { 0 };
 
-void SetAuthKeysPattern(const char* pattern)
-{
-    if (pattern != NULL) {
-        WMEMSET(authKeysPattern, 0, sizeof(authKeysPattern));
-        WSTRNCPY(authKeysPattern, pattern, sizeof(authKeysPattern) - 1);
-    }
-}
-
-
-static int ResolveAuthKeysPath(const char* homeDir, char* resolved)
+/* Resolve the authorized keys file path for a user. The pattern is the user's
+ * configured AuthorizedKeysFile (resolved per request from the per-user config)
+ * and is passed in explicitly rather than read from shared state so concurrent
+ * authentications (e.g. Windows threaded mode) cannot race on it. A NULL or
+ * empty pattern falls back to the default authorized_keys location. */
+static int ResolveAuthKeysPath(const char* homeDir, const char* pattern,
+                               char* resolved)
 {
     int ret = WS_SUCCESS;
     char* idx;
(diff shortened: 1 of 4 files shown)


apps/wolfsshd/auth.c
```diff
@@ -1124,18 +1138,41 @@ static int RequestAuthentication(WS_UserAuthData* authData,
...
+     * wolfSSHD_AuthGetUserConf). DoCheckUser has already confirmed the user
+     * exists, so this is a rare edge. Fail closed rather than fall back to the
+     * permissive global node: such a user cannot complete a session anyway
+     * (session setup in wolfsshd.c rejects an unresolvable user config with
+     * WS_FATAL_ERROR), so denying auth here is safe and avoids evaluating
+     * password/public-key authorization against the wrong config node. */
+    if (ret == WOLFSSH_USERAUTH_SUCCESS) {
+        usrConf = wolfSSHD_AuthGetUserConf(authCtx, usr, NULL, NULL, NULL, NULL,
+                                           NULL);
+        if (usrConf == NULL) {
```
yosuke-wolfssl, line 1207: Fixed

apps/wolfsshd/auth.c
```diff
@@ -1238,13 +1276,13 @@ static int RequestAuthentication(WS_UserAuthData* authData,
         /* if this is a certificate and no specific authorized keys file has
             * been set then rely on CA to have verified the cert */
         if (authData->sf.publicKey.isCert &&
-                !wolfSSHD_ConfigGetAuthKeysFileSet(authCtx->conf)) {
+                !wolfSSHD_ConfigGetAuthKeysFileSet(usrConf)) {
```
yosuke-wolfssl, line 1329: Fixed

apps/wolfsshd/auth.c
```diff
@@ -1238,13 +1287,13 @@ static int RequestAuthentication(WS_UserAuthData* authData,
...
```
yosuke-wolfssl, line 1329: Fixed

ejohnstown (approved):

PR #1015, opened 2026-06-10
Reject symlinks in default SCP send callback

yosuke-wolfssl:
# Reject symlinks in the default SCP send callback

## Summary

Closes the symlink-follow path-traversal in the default SCP **send** path
(finding 5112). An authenticated SCP peer could request a file or directory that
is (or contains) a symbolic link; the server followed it with `WFOPEN`/`WSTAT`,
streaming the link target's contents and escaping the intended directory scope.
`wolfSSH_CleanPath` only normalizes the path string and never resolves links, so
nothing rejected the link before the file operation.

This PR rejects symlinks at every send sink, makes the file-content open
*atomic* against a swap race on POSIX, and adds regression coverage. The
receive-side guards landed separately and are already on master; this is the
send-side counterpart.

### Port layer (`wolfssh/port.h`, `src/port.c`)
- New `WOLFSSH_O_NOFOLLOW` flag in the existing `WOLFSSH_O_*` family —
  `O_NOFOLLOW` on POSIX, `0` where unavailable (Windows / link-less filesystems).
- New `WFDOPEN` `W*`-wrapper (`fdopen` / `_fdopen`) so the fd→`FILE*` bridge
  follows the same convention as the other file macros.
- New `wFopenNoFollow()` helper: opens a file for reading **without following a
  final-component symlink**. On POSIX it is atomic via `open(O_RDONLY |
  O_NOFOLLOW)` + `fdopen`, closing the check-then-open race; on Windows / where
  `O_NOFOLLOW` is absent it falls back to a best-effort `wIsSymlink` +
  `WFOPEN`. Gated on `WOLFSSH_HAVE_SYMLINK && (WOLFSSH_SFTP || WOLFSSH_SCP)`.

### SCP send path (`src/wolfscp.c`)
- **Single-file request** and **recursive file entries** now open via
  `wFopenNoFollow` instead of `WFOPEN`.
- **Recursive root** is leaf-checked with `wIsSymlink` before `ScpPushDir`
  (it is opened with `opendir`, which follows links, so `O_NOFOLLOW` does not
  apply there) — symmetric with the single-file guard.
- **Recursive per-entry** symlinks continue to be rejected as `ScpProcessEntry`
  traverses, so every component below the root is checked.
- Symlink rejection aborts the transfer (`WS_SCP_ABORT`), matching the existing
  receive-side convention; the generic "error getting file stats" log is
  suppressed when the abort already logged its own reason.
- A "Symlink handling" note in the `wsScpSendCallback` header documents the
  atomic-vs-best-effort boundary and the residual TOCTOU posture.

### Tests
- `scripts/scp.test`: end-to-end case proving the server refuses a symlinked
  single-file request (now exercising the `O_NOFOLLOW` open), with cleanup of
  planted artifacts.
- `tests/api.c` (`test_wolfSSH_SCP_SendSymlinkReject`): callback-level coverage
  for the recursive-root rejection (`WS_SCP_ABORT`) and for `wFopenNoFollow`
  (opens a real file, refuses a symlink). This path is unreachable from the
  example client, which never issues `scp -f -r`.

## Design decisions

- **No SCP "trusted base path".** Unlike SFTP (`sftpDefaultPath`), SCP has no
  library-level base to contain a resolved path against, and `wolfSSH_RealPath`
  does not resolve links — so a "stays under base" check isn't possible. The
  proportionate guard is leaf-level symlink rejection, matching the single-file
  path. Confinement remains the deployment's responsibility (wolfsshd uses OS
  chroot).
- **Leaf-only, no leading-path walk.** Below-root components are covered by the
  per-entry check; the leading path (server-filesystem territory) is not walked,
  because without a base it would reject ordinary system symlinks (e.g. macOS
  `/tmp -> /private/tmp`) and break legitimate transfers.
- **Abort vs skip.** A rejected symlink aborts the whole transfer rather than
  skipping the entry, consistent with the receive-side guards and the safe
  security default (there is no "skip entry" primitive; `WS_NEXT_ERROR` means
  end-of-directory).
- **Portability over `O_NOFOLLOW`-everywhere.** `O_NOFOLLOW` is POSIX-only, so
  Windows/embedded keep the documented best-effort `wIsSymlink` check.

## Behavioral / compatibility notes

- A request whose **final path component is a symlink** is now refused (both
  single-file and recursive root). Deployments that intentionally serve a tree
  via a top-level symlink should request the real path, or build with
  `WOLFSSH_NO_SYMLINK_CHECK` to disable the check (and its protection).
- No change for ordinary (non-symlink) files/directories, and the feature
  compiles out cleanly where `WOLFSSH_HAVE_SYMLINK` is not defined.

## Testing

- `./configure --enable-scp --enable-sftp` and `--disable-scp --disable-sftp`:
  clean builds (verified the guards compile out).
- `scripts/scp.test` and `tests/api.test`: pass.
- Negative controls: reverting the single-file guard trips the scp.test bypass
  check; flipping the api.c expectation trips the assertion — confirming neither
  test is vacuous.
- ASan + UBSan (`-fsanitize=address,undefined`): `scp.test` and `api.test` clean,
  no reports in `wolfscp.c`/`port.c`/`api.c`.

### DIFF
--- a/src/wolfscp.c
+++ b/src/wolfscp.c
@@ -646,6 +646,16 @@ int DoScpSource(WOLFSSH* ssh)
                     continue;
 
                 } else if (ssh->scpConfirm == WS_SCP_ABORT) {
+                #if !defined(NO_FILESYSTEM) && \
+                        !defined(WOLFSSH_SCP_USER_CALLBACKS)
+                    /* drain any partial recursive dir stack so a later exec on
+                     * this connection starts from a fresh root, not a stale
+                     * handle left by the aborted walk */
+                    ScpSendCtx* sendCtx =
+                        (ScpSendCtx*)wolfSSH_GetScpSendCtx(ssh);
+                    if (sendCtx != NULL)
+                        ScpSendCtxFreeDirs(ssh->fs, sendCtx, ssh->ctx->heap);
+                #endif
                     ssh->scpState = SCP_SEND_CONFIRMATION;
                     ssh->scpNextState = SCP_DONE;
                     continue;
@@ -2455,7 +2465,14 @@ static int GetFileStats(void *fs, ScpSendCtx* ctx, const char* fileName,
         (ctx->s.dwFileAttributes & FILE_ATTRIBUTE_READONLY ? 0 : 0200);
     *fileMode |= (ctx->s.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) ? 040000 : 0;
 #else
+    /* WLSTAT (lstat on POSIX) leaves a symlink unfollowed, so it classifies as
+     * neither dir nor file and is skipped; WOLFSSH_NO_SYMLINK_CHECK falls back
+     * to WSTAT so links are followed by design. */
+#ifdef WOLFSSH_HAVE_SYMLINK
+    if (WLSTAT(fs, fileName, &ctx->s) < 0) {
+#else
     if (WSTAT(fs, fileName, &ctx->s) < 0) {
+#endif
         ret = WS_BAD_FILE_E;
         #ifdef WOLFSSL_NUCLEUS
         if (WSTRLEN(fileName) < 4 && WSTRLEN(fileName) > 2 &&
@@ -2543,11 +2560,16 @@ static ScpDir* ScpNewDir(void *fs, const char* path, void* heap)
         }
     }
 #else
+    #ifdef WOLFSSH_HAVE_SYMLINK
+    /* refuse a symlinked directory leaf atomically (closes the descend race) */
+    if (wOpendirNoFollow(fs, &entry->dir, path) != 0) {
+    #else
     if (WOPENDIR(fs, heap, &entry->dir, path) != 0
         #if !defined(WOLFSSL_NUCLEUS) && !defined(WOLFSSH_ZEPHYR)
             || entry->dir == NULL
         #endif
             ) {
+    #endif
         WFREE(entry, heap, DYNTYPE_SCPDIR);
         WLOG(WS_LOG_ERROR, scpError, "opendir failed on directory",
              WS_INVALID_PATH_E);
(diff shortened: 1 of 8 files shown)


src/wolfscp.c
```diff
@@ -2616,6 +2616,17 @@ static int ScpProcessEntry(WOLFSSH* ssh, char* fileName, word64* mTime,
         }
     }
 
+#ifdef WOLFSSH_HAVE_SYMLINK
+    /* GetFileStats classifies via WSTAT, which follows symlinks; a planted link
+     * to a file outside the traversed tree would be sent transparently.  Reject
+     * the entry here, before descending into it or opening it, so neither the
+     * directory-descent nor the file-open sink below follows the link. */
+    if (ret == WS_SUCCESS && WS_IsSymlink(filePath)) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

src/wolfscp.c
```diff
@@ -2820,6 +2820,16 @@ static int ScpProcessEntry(WOLFSSH* ssh, char* fileName, word64* mTime,
             WSTRNCPY(fileName, sendCtx->entry->d_name,
                      DEFAULT_SCP_FILE_NAME_SZ);
         #endif
+        #ifdef WOLFSSH_HAVE_SYMLINK
+            /* filePath is fully built; reject a planted symlink before
+             * GetFileStats or any descend/open follows it. */
+            if (ret == WS_SUCCESS && wIsSymlink(filePath)) {
```
yosuke-wolfssl, line 2859: Added test

src/wolfscp.c
```diff
@@ -3037,9 +3094,40 @@ int wsScpSendCallback(WOLFSSH* ssh, int state, const char* peerRequest,
...
-                /* first request, keep track of request directory */
-                ret = ScpPushDir(ssh->fs, sendCtx, peerRequest, ssh->ctx->heap);
+                /* first request, keep track of request directory.  Reject a
+                 * symlink root here (a trailing separator can still follow);
+                 * see the symlink-handling note in this function's header. */
+                ret = WS_SUCCESS;
+            #ifdef WOLFSSH_HAVE_SYMLINK
+                /* lstat() follows the link when the path ends in a separator,
+                 * so check the root with any trailing separators removed */
+                rootLen = (word32)WSTRLEN(peerRequest);
```
yosuke-wolfssl: Fixed

ejohnstown (approved):

PR #1016, opened 2026-06-10
wolfsshd: reject unsafe ownership/symlink on trust-anchor file loads

yosuke-wolfssl:
## Description
SetupCTX() in apps/wolfsshd/wolfsshd.c loads the host key, host certificate, and user CA through getBufferFromFile(), which called WFOPEN(..., "rb") with no stat/lstat/fstat/O_NOFOLLOW guard. The most dangerous is UserCAKeysFile: its buffer is handed to wolfSSH_CTX_AddRootCert_buffer, installing a trusted root CA for user-auth certificate verification.
A local non-root attacker who can pre-create a symlink at the configured path (or drop a file via a writable ancestor directory) before the root-launched daemon starts gets an attacker-controlled CA into the trust store, then authenticates as any user (including root) with a cert signed by that CA — bypassing pubkey and password auth. Same missing-ownership root cause as the tracked HostKeyFile finding and the host-cert load.
Addressed by f_5901.
### Fix (`apps/wolfsshd/wolfsshd.c`)
Added an `int secure` parameter to `getBufferFromFile()`. When set (POSIX only), the file is validated before it is read:
- **`lstat()`** rejects a symlink or any non-regular leaf (FIFO / device / directory).
- **`open(O_RDONLY | O_NOFOLLOW | O_NONBLOCK)`** won't follow a symlink and can't block on a FIFO swapped in after the `lstat()`.
- **`fstat()`** on the opened fd (TOCTOU-free) requires a regular file, owned by root or the daemon's euid, and not group/world writable.
- **`O_NONBLOCK`** is cleared before the buffered reads.
Any failure logs and returns `NULL`; each call site already maps `NULL` to an error and skips the load, so the daemon **fails closed**.

Windows keeps the existing `WFOPEN` path (no comparable uid model; relies on filesystem ACLs), with a comment noting the intentional platform exemption.
This closes the `UserCAKeysFile` issue and also covers the `HostKeyFile` finding and the host-cert load with the same gate.
### New Test
Added new tests to exercise the new code path

### DIFF
--- a/.github/workflows/paramiko-sftp-test.yml
+++ b/.github/workflows/paramiko-sftp-test.yml
@@ -94,8 +94,12 @@ jobs:
           Subsystem sftp internal-sftp
           EOF
           
-          # Set proper permissions for keys
+          # wolfSSHd refuses to load a host key unless it is owned by root or
+          # the daemon's user and is not group or world writable. The daemon is
+          # launched with sudo (euid 0) while the checkout key is owned by the
+          # runner user, so make the key root-owned and 0600.
           chmod 600 ./keys/server-key.pem
+          sudo chown 0:0 ./keys/server-key.pem
           
           # Print debug info
           echo "Contents of sshd_config.txt:"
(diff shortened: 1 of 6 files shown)


ejohnstown (changes requested):
- Gap 1 — ancestor-directory validation. getBufferFromFile()'s secure branch checks only the leaf file, not its parent chain. Add a POSIX-only OpenSSH-style path walk (realpath then stat each ancestor up to /, rejecting any dir not owned by root/euid or group/world-writable), called before the open() at apps/wolfsshd/wolfsshd.c:291, failing closed on rejection.
- Gap 2 — O_NOFOLLOW-absent leaf TOCTOU. On platforms where O_NOFOLLOW compiles to 0 (apps/wolfsshd/wolfsshd.c:53-57), a narrow lstat→open window on the leaf lets a symlink be swapped in (still backstopped by the fstat uid check). Close it by comparing the lstat result against the opened fd's fstat (st_dev/st_ino match) and rejecting on mismatch; note the realpath walk from Gap 1 also covers the intermediate-symlink case here.

yosuke-wolfssl:
Close since this is covered by #1042

PR #1027, opened 2026-06-15
Treat combined Match User/Group blocks as a conjunction in wolfsshd

yosuke-wolfssl:
## Summary

`wolfSSHD_GetUserConf` evaluated a combined `Match User X Group Y` config block with **OR** semantics instead of **AND**. OpenSSH treats a `Match` line as a conjunction, but the lookup loop broke out on a username match *or* a group match independently. Since a combined `Match User X Group Y` line sets **both** `usrAppliesTo` and `groupAppliesTo` on the same config node, the block was returned for:

- any user named `alice` regardless of group, and
- any user whose primary group is `admins` regardless of name.

When such a block relaxes a global `ForceCommand` (e.g. from `internal-sftp` to a full shell) or changes `ChrootDir`, an authenticated user satisfying only one of the two selectors inherited the permissive policy and could escape SFTP-only confinement.

Addressed by f_5692.

## Fix

Each config node is now evaluated as a conjunction of its set selectors: a node applies only when **every non-NULL selector** on it matches; a NULL selector acts as a wildcard.

| Node has      | Match condition                |
|---------------|--------------------------------|
| user only     | user matches                   |
| group only    | group matches                  |
| user + group  | user matches **AND** group matches |
| neither       | wildcard (global/head default) |

This preserves the legitimate, common single-selector blocks (`Match User alice`, `Match Group admins`) while closing the combined-block bypass. When no node matches, it still falls back to the global head config. The NULL-group path (WIN32, or failed group resolution in `wolfSSHD_AuthGetUserConf`) now fails closed against a combined node rather than treating NULL as a wildcard.

## Changes

- **`apps/wolfsshd/configuration.c`** — rewrote the matching loop in `wolfSSHD_GetUserConf` to use per-node conjunction semantics.
- **`apps/wolfsshd/test/test_configuration.c`** — added `test_GetUserConfMatchGroupAnd` covering combined and single-selector blocks with users satisfying zero, one, and both selectors, plus the NULL-group fail-closed case.

## Testing

- New test `test_GetUserConfMatchGroupAnd` and existing `test_GetUserConfMatchOverride` pass.
- **Negative control:** with the fix reverted, the new test fails (`ret=-1001`), confirming it actually catches the OR bug.
- **Sanitizers:** built and ran the config test under `-fsanitize=address,undefined -fno-omit-frame-pointer` — clean, no ASan/UBSan reports.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -1084,10 +1084,10 @@ static int DoCheckUser(const char* usr, WOLFSSHD_AUTH* auth)
          * scoped to this branch.
          *
          * A NULL return means the root user's configuration could not be
-         * resolved (e.g. getgrgid() failure inside wolfSSHD_AuthGetUserConf).
-         * Fail closed and reject the login in that case rather than falling
-         * back to the global node, since denying root on an unresolvable
-         * configuration is the safe choice. */
+         * resolved (e.g. group set enumeration failed inside
+         * wolfSSHD_AuthGetUserConf). Fail closed and reject the login in that
+         * case rather than falling back to the global node, since denying root
+         * on an unresolvable configuration is the safe choice. */
         usrConf = wolfSSHD_AuthGetUserConf(auth, usr, NULL, NULL, NULL, NULL,
                                            NULL);
         if (usrConf == NULL || wolfSSHD_ConfigGetPermitRoot(usrConf) == 0) {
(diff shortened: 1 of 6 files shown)


ejohnstown (changes requested):
There is still the secondary group defect to fix. An example, I'm user "john" on my Linux box, and I'm in primary group "john". I'm also in secondary groups "vboxusers" and "wireshark". If I want to give people who use wireshark on my machine a specific banner, I'd put in my sshd_config:
```
Match Group wireshark
  Banner /etc/ssh/wireshark-banner.txt
```

yosuke-wolfssl:
Hello @ejohnstown ,
I fixed it to handle the group list accordingly.

ejohnstown (approved):

PR #1035, opened 2026-06-16
Bound OSC index before reads in wolfSSH_DoOSC

yosuke-wolfssl:
### Summary

`wolfSSH_DoOSC` read `buf[*idx]` without first checking `*idx < bufSz`. The
caller `wolfSSH_ConvertConsole` only guards `i + 1 < bufSz` before forwarding
`z = i + 2` as `*idx`, so `*idx` can equal `bufSz`. A peer-supplied
channel-data buffer whose last two bytes are `ESC ]` (`0x1B 0x5D`) drives
`wolfSSH_DoOSC` to dereference `buf[bufSz]` — one byte past the heap
allocation — at the very first conditional.

Two adjacent reads inside the OSC `'0'` (set icon/window title) handler shared
the same flaw: the `;` check after `*idx += 1`, and the post-loop BEL test that
read `buf[i]` when the scan reached `i == bufSz`.

This code is compiled only under `USE_WINDOWS_API` and is reached through the
documented `wolfSSH_ConvertConsole` API when an external (Windows) consumer
feeds received channel-data bytes into it.

### Changes

**`src/wolfterm.c`**
- Add an entry guard to `wolfSSH_DoOSC`: when `*idx >= bufSz`, drop the
  sequence (`return WS_SUCCESS`) before any `buf[*idx]` read.
- Guard the `;` read in the title handler after `*idx += 1`.
- Replace the post-loop `buf[i] != 0x07` test with `i >= bufSz`, and set
  `*idx = bufSz` so a truncated title is consumed/dropped rather than emitted
  as raw console output.
- Truncated OSC input is dropped consistently (`WS_SUCCESS`) instead of
  returning `WS_WANT_READ`. `wolfSSH_DoOSC` has no resume path — it never saves
  state to `escBuf` and `escState` is never set to `WS_ESC_OSC` — so a
  `WS_WANT_READ` would leave broken parser state (`escState == WS_ESC_START`,
  `escBufSz == 0`) and mis-parse the continuation buffer on re-entry.

**`tests/api.c`**
- Extend `test_wolfSSH_ConvertConsole` with OSC vectors covering each new
  bounds check — buffers ending at `ESC ]`, `ESC ] 0`, and `ESC ] 0 ; title`
  (no BEL) — plus a well-formed `ESC ] 0 ; hi BEL` positive control.

**`.github/workflows/windows-check.yml`**
- Add an `asan-tests` job that builds and runs `api-test` and `unit-test` under
  the MSVC AddressSanitizer. This is the only CI job that exercises the
  `USE_WINDOWS_API` console code (`wolfSSH_DoOSC`) under a sanitizer: the
  existing `sanitizer.yml` runs on Linux, where `wolfterm.c` is compiled out,
  so the OSC path previously had no sanitizer coverage.
  - Discovers an installed MSVC toolset that ships the ASan runtime libs (the
    pinned v142 does not) and builds the whole job with it.
  - Builds with `/p:EnableASAN=true` (instrumenting the referenced `wolfssh`
    library and `src/wolfterm.c`) and `/p:WholeProgramOptimization=false` so
    objects/libs link without a cross-version `/GL` code-generation conflict.
  - Emits explicit signals so a green run is meaningful: a positive-control
    canary proving ASan detects an overflow on the runner, a `dumpbin`
    assertion that the test binaries import the ASan runtime, and
    `ASAN_OPTIONS=print_stats=1` for in-process evidence.

### Testing

- Existing `tests/api.c` suite. The new OSC vectors run on `USE_WINDOWS_API`
  builds via the `api-test` binary.
- The fixed logic was extracted into a standalone harness and run under
  ASan + UBSan locally: clean on the PoC and all truncation/well-formed cases.
  A negative control with the guard removed reproduced the original
  `heap-buffer-overflow READ of size 1, 0 bytes after a 2-byte region`.
- New Windows `asan-tests` CI job is green: the positive control reports
  `AddressSanitizer: stack-buffer-overflow` then continues, the test binaries
  are confirmed linked against the ASan runtime, and `api-test`/`unit-test`
  run clean under ASan.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -2127,6 +2127,22 @@ static byte color_test[] = {
     0x1B, 0x5B, 0x34, 0x39, 0x6D, 0x64, 0x65, 0x66,
     0x61, 0x75, 0x6C, 0x74, 0x20, 0x62, 0x67, 0x0A,
 };
+
+/* OSC (ESC ]) sequences that end before the command is complete. These
+ * exercise the bounds checks in wolfSSH_DoOSC; each truncated sequence is
+ * dropped (WS_SUCCESS) without reading past the end of the buffer. */
+static byte osc_trunc_cmd[] = { /* ends right after "ESC ]" */
+    0x1B, 0x5D
+};
+static byte osc_trunc_arg[] = { /* ends after "ESC ] 0" */
+    0x1B, 0x5D, 0x30
+};
+static byte osc_trunc_str[] = { /* "ESC ] 0 ; title" with no BEL terminator */
+    0x1B, 0x5D, 0x30, 0x3B, 0x74, 0x69, 0x74, 0x6C, 0x65
+};
+static byte osc_full[] = { /* well formed "ESC ] 0 ; hi BEL" */
+    0x1B, 0x5D, 0x30, 0x3B, 0x68, 0x69, 0x07
+};
 #endif /* USE_WINDOWS_API */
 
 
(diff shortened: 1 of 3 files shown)


tests/api.c
```diff
@@ -2163,6 +2179,17 @@ static void test_wolfSSH_ConvertConsole(void)
     AssertIntEQ(wolfSSH_ConvertConsole(ssh, stdoutHandle, color_test, 1),
             WS_SUCCESS); /* should skip over unknown console code */
 
+    /* truncated OSC sequences must be dropped without an out of bounds read */
+    AssertIntEQ(wolfSSH_ConvertConsole(ssh, stdoutHandle, osc_trunc_cmd,
+                sizeof(osc_trunc_cmd)), WS_SUCCESS);
```
yosuke-wolfssl, line 2184: Extended the CI test

ejohnstown (approved):

PR #1038, opened 2026-06-16
Set keyAllocated=1 when initializing any key

padelsbach:
Fixes a possible memory leak if parsing fails. Only a problem with `--enable-heapmath` since normal spmath does not allocate buffers for the key.

Added unit tests and heapmath variant to CI.

Also, adding heapmath+asan uncovered a second unrelated leak when freeing the KEX key. Fix included as a second commit.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -622,6 +622,22 @@ static void HandshakeInfoFree(HandshakeInfo* hs, void* heap)
 #ifndef WOLFSSH_NO_DH
         WFREE(hs->primeGroup, heap, DYNTYPE_MPINT);
         WFREE(hs->generator, heap, DYNTYPE_MPINT);
+        if (hs->useDh) {
+            wc_FreeDhKey(&hs->privKey.dh);
+        }
+#endif
+#ifndef WOLFSSH_NO_ECDH
+        /* privKey is a union; the Curve25519+ML-KEM case also sets
+         * useEccMlKem but generates a curve25519 key, so free it below. */
+        if (hs->useEcc || (hs->useEccMlKem && !hs->useCurve25519MlKem)) {
+            wc_ecc_free(&hs->privKey.ecc);
+        }
+#endif
+#if !defined(WOLFSSH_NO_CURVE25519_SHA256) || \
+    !defined(WOLFSSH_NO_CURVE25519_MLKEM768_SHA256)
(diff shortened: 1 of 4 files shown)


ejohnstown (changes requested):
Ignore this comment.

src/internal.c
```diff
@@ -622,6 +622,22 @@ static void HandshakeInfoFree(HandshakeInfo* hs, void* heap)
 #ifndef WOLFSSH_NO_DH
         WFREE(hs->primeGroup, heap, DYNTYPE_MPINT);
         WFREE(hs->generator, heap, DYNTYPE_MPINT);
+#endif
+#ifndef WOLFSSH_NO_DH
+        if (hs->useDh) {
+            wc_FreeDhKey(&hs->privKey.dh);
+        }
+#endif
+#ifndef WOLFSSH_NO_ECDH
```
ejohnstown, line 629: There's a bug in SendKexDhInit() that should be fixed with this PR. When KEX is ID_CURVE25519_MLKEM768_SHA256, useEccMlKem and useCurve25519MlKem are set. This will cause a double free of hs->privKey, once as each of ECC and Curve25519.

ejohnstown (changes requested):

src/internal.c
```diff
@@ -622,6 +622,22 @@ static void HandshakeInfoFree(HandshakeInfo* hs, void* heap)
 #ifndef WOLFSSH_NO_DH
         WFREE(hs->primeGroup, heap, DYNTYPE_MPINT);
         WFREE(hs->generator, heap, DYNTYPE_MPINT);
+#endif
+#ifndef WOLFSSH_NO_DH
```
ejohnstown: This `#ifndef WOLFSSH_NO_DH` is adjacent to another one.

ejohnstown (changes requested):

compile_commands.json
ejohnstown, line 1: Please remove this file from the PR.

src/internal.c
ejohnstown: This was fine here. The bug is in SendKexDhInit(). The case where ID_CURVE25519_MLKEM768_SHA256 is handled, both useCurve25519MlKem (correct) and useEccMlKem (incorrect) are getting set true.

ejohnstown: Nevermind.

ejohnstown (approved):

Issue #1047, opened 2026-06-21
Several SSH (RFC 4252/4253/4254) Compliance Issues in wolfSSH


Zhaodl1:
Hi, while reviewing wolfSSH's(`e2bfa19`) SSH protocol implementation against RFC 4252, 4253, and 4254, we noticed several places where the current behavior appears to differ from the specification. Each item below cites the relevant RFC text alongside the corresponding source code for your reference. We hope this is helpful for improving RFC conformance.

### 1. Diffie-Hellman Exchange Value Range Not Verified for Modular Groups

**RFC Reference**: RFC 4253 Section 8 (lines 1199-1201)

> "Values of 'e' or 'f' that are not in the range [1, p-1] MUST NOT be sent or accepted by either side. If this condition is violated, the key exchange fails."

**Analysis**:
In `DoKexDhInit`, the client-supplied exchange value `e` is validated only by its encoded length (`eSz` bounds) before being copied into the handshake state and forwarded to `KeyAgreeDh_server`, where it is passed directly into `wc_DhAgree` to compute the shared secret. No check that the value lies in the range [1, p-1] is performed on this modular-exponentiation path (the curve25519 path separately validates the peer public value through `wc_curve25519_check_public`). As a result, an out-of-range `e` is currently accepted and a `KEXDH_REPLY` (containing `f` and the host-key signature) is produced, whereas the RFC states such values MUST NOT be accepted and the exchange must fail.

**Source Code Evidence** (`src/internal.c`):

```c
// src/internal.c:4964-4972
    if (ret == WS_SUCCESS) {
        /* Validate eSz */
        if (eSz > (word32)sizeof(ssh->handshake->e) || eSz == 0)
            ret = WS_PUBKEY_REJECTED_E;     // length-only validation
    }

    if (ret == WS_SUCCESS) {
        WMEMCPY(ssh->handshake->e, e, eSz);
        ssh->handshake->eSz = eSz;
```

```c
// src/internal.c:12180-12184  (KeyAgreeDh_server)
        if (ret == 0) {
            PRIVATE_KEY_UNLOCK();
            ret = wc_DhAgree(privKey, ssh->k, &ssh->kSz, y_ptr, ySz,
                    ssh->handshake->e, ssh->handshake->eSz);   // e used without range check
            PRIVATE_KEY_LOCK();
        }
```
---

### 2. "forwarded-tcpip" Channel Open Not Verified Against a Prior Remote-Forward Request

[...]

PR #1049, opened 2026-06-22
Harden KEX and userauth packet handling

ejohnstown:
Addresses parts of issue #1047 and finding F-5690.

- Validate peer KEX public keys — check the peer public value before deriving the shared secret: wc_DhCheckPubKey (DH), wc_ecc_check_key (ECDH and ECC+ML-KEM hybrid), client and server. DH rejects degenerate/small-order values per NIST SP 800-56A (stricter than RFC 4253's [1, p-1]); ECC rejects off-curve/bad-order points. Negative unit tests cover the plain DH and ECDH paths (degenerate DH values, off-curve ECDH points). The ECC+ML-KEM hybrid paths route the peer ECC point through the same EccCheckPeerKey helper as the plain ECDH path, so its off-curve rejection is exercised by the plain-path unit tests; the hybrid wrappers are not fed an off-curve point directly (that would require a valid ML-KEM ciphertext/public-key prefix) and are instead exercised end-to-end by kex.test. (F-5690, #1047(1))
- Simplify redundant KEX method flags — drop the redundant useCurve25519MlKem flag, deriving the hybrid from useCurve25519 && useEccMlKem, and merge the duplicate Curve25519 keygen branch.
- Reject packets with too-little padding — enforce the RFC 4253 6 minimum of 4 padding bytes on receive, returning WS_BUFFER_E below MIN_PAD_LENGTH, with a negative unit test. (#1047(5))
- Reject password-change auth requests — fail userauth when the password-change flag is set, never invoking the callback with the current password, while still fully parsing the new-password field; negative unit test. Per RFC 4252 8. (#1047(6))

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -11099,6 +11099,13 @@ static int DoPacket(WOLFSSH* ssh, byte* bufferConsumed)
     idx += UINT32_SZ;
     padSz = buf[idx++];
 
+    /* RFC 4253 section 6 requires at least four bytes of padding. */
+    if (padSz < MIN_PAD_LENGTH) {
+        WLOG(WS_LOG_DEBUG, "Packet padding length %u below minimum %u",
+             (word32)padSz, (word32)MIN_PAD_LENGTH);
+        return WS_BUFFER_E;
+    }
+
     /* check for underflow */
     if ((word32)(PAD_LENGTH_SZ + padSz + MSG_ID_SZ) > ssh->curSz) {
         return WS_OVERFLOW_E;
--- a/src/internal.c
+++ b/src/internal.c
@@ -642,15 +642,17 @@ static void HandshakeInfoFree(HandshakeInfo* hs, void* heap)
         }
 #endif
 #ifndef WOLFSSH_NO_ECDH
-        /* privKey is a union; Curve25519+ML-KEM sets useEccMlKem but uses
-         * a curve25519 key, freed in the curve25519 block below. */
-        if (hs->useEcc || (hs->useEccMlKem && !hs->useCurve25519MlKem)) {
+        /* privKey is a union; the Curve25519+ML-KEM hybrid sets both
+         * useMlKem and useCurve25519 but generates a curve25519 key, which is
+         * freed below. The NIST ECC hybrid sets useEcdh, so useEcdh alone
+         * covers every case that generates an ecc key. */
+        if (hs->useEcdh) {
             wc_ecc_free(&hs->privKey.ecc);
         }
 #endif
 #if !defined(WOLFSSH_NO_CURVE25519_SHA256) || \
     !defined(WOLFSSH_NO_CURVE25519_MLKEM768_SHA256)
-        if (hs->useCurve25519 || hs->useCurve25519MlKem) {
+        if (hs->useCurve25519) {
             wc_curve25519_free(&hs->privKey.curve25519);
         }
 #endif
@@ -6583,14 +6585,17 @@ static int KeyAgree_client(WOLFSSH* ssh, byte hashId, const byte* f, word32 fSz)
     if (ssh->handshake->useDh) {
         ret = KeyAgreeDh_client(ssh, hashId, f, fSz);
     }
-    else if (ssh->handshake->useEcc) {
+    /* Check useMlKem before the plain classical flags: a hybrid sets both
+     * useMlKem and its classical flag (useEcdh or useCurve25519) and must
+     * route to the hybrid agreement. */
+    else if (ssh->handshake->useMlKem) {
+        ret = KeyAgreeEcdhMlKem_client(ssh, hashId, f, fSz);
+    }
+    else if (ssh->handshake->useEcdh) {
         ret = KeyAgreeEcdh_client(ssh, hashId, f, fSz);
     }
     else if (ssh->handshake->useCurve25519) {
         ret = KeyAgreeCurve25519_client(ssh, hashId, f, fSz);
-    }
-    else if (ssh->handshake->useEccMlKem) {
-        ret = KeyAgreeEcdhMlKem_client(ssh, hashId, f, fSz);
     }
     else {
         ret = WS_INVALID_ALGO_ID;
@@ -6820,7 +6825,7 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 
         /* Hash in the shared secret K. */
         if (ret == WS_SUCCESS) {
-            if (!ssh->handshake->useEccMlKem) {
+            if (!ssh->handshake->useMlKem) {
                 ret = CreateMpint(ssh->k, &ssh->kSz, &kPad);
             }
         }
@@ -7000,8 +7005,8 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     }
 
     if (ret == WS_SUCCESS) {
-        /* If we aren't using ECC with ML-KEM, use padding. */
-        ret = GenerateKeys(ssh, hashId, !ssh->handshake->useEccMlKem);
+        /* If we aren't using an ML-KEM hybrid, use padding. */
+        ret = GenerateKeys(ssh, hashId, !ssh->handshake->useMlKem);
     }
 
     if (ret == WS_SUCCESS) {
@@ -14223,9 +14228,9 @@ int SendKexDhReply(WOLFSSH* ssh)
 #endif
     byte msgId = 0;
     byte useDh = 0;
-    byte useEcc = 0;
+    byte useEcdh = 0;
     byte useCurve25519 = 0;
-    byte useEccMlKem = 0;
+    byte useMlKem = 0;
 
     WLOG(WS_LOG_DEBUG, "Entering SendKexDhReply()");
 
@@ -14313,19 +14318,19 @@ int SendKexDhReply(WOLFSSH* ssh)
 #endif
 #ifndef WOLFSSH_NO_ECDH_SHA2_NISTP256
             case ID_ECDH_SHA2_NISTP256:
-                useEcc = 1;
+                useEcdh = 1;
                 msgId = MSGID_KEXDH_REPLY;
                 break;
 #endif
 #ifndef WOLFSSH_NO_ECDH_SHA2_NISTP384
             case ID_ECDH_SHA2_NISTP384:
-                useEcc = 1;
+                useEcdh = 1;
                 msgId = MSGID_KEXDH_REPLY;
                 break;
 #endif
 #ifndef WOLFSSH_NO_ECDH_SHA2_NISTP521
             case ID_ECDH_SHA2_NISTP521:
-                useEcc = 1;
+                useEcdh = 1;
                 msgId = MSGID_KEXDH_REPLY;
                 break;
 #endif
@@ -14338,19 +14343,22 @@ int SendKexDhReply(WOLFSSH* ssh)
 #endif
 #ifndef WOLFSSH_NO_NISTP256_MLKEM768_SHA256
             case ID_NISTP256_MLKEM768_SHA256:
-                useEccMlKem = 1;
+                useEcdh = 1;
+                useMlKem = 1;
                 msgId = MSGID_KEXKEM_REPLY;
                 break;
 #endif
 #ifndef WOLFSSH_NO_NISTP384_MLKEM1024_SHA384
             case ID_NISTP384_MLKEM1024_SHA384:
-                useEccMlKem = 1;
+                useEcdh = 1;
+                useMlKem = 1;
                 msgId = MSGID_KEXKEM_REPLY;
                 break;
 #endif
 #ifndef WOLFSSH_NO_CURVE25519_MLKEM768_SHA256
             case ID_CURVE25519_MLKEM768_SHA256:
-                useEccMlKem = 1;
+                useCurve25519 = 1;
+                useMlKem = 1;
                 msgId = MSGID_KEXKEM_REPLY;
                 break;
 #endif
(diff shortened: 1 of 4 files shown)


tests/unit.c
```diff
@@ -3888,6 +4166,104 @@ static int test_KeyAgreeDh_server_zeroesEphemeralPrivKey(void)
...
+    }
+    ssh->handshake->kexId = ID_ECDH_SHA2_NISTP256;
+
+    /* KeyAgreeEcdh_client frees ssh->handshake->privKey.ecc on the way out;
+     * give it an initialised key so that cleanup path is well defined. */
+    if (wc_ecc_init(&ssh->handshake->privKey.ecc) != 0) {
+        result = -772;
+        goto out;
+    }
+
```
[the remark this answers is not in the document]
ejohnstown: Good catch. The test now generates a real P-256 ephemeral key into ssh->handshake->privKey.ecc (wc_ecc_make_key_ex + wc_ecc_set_rng, mirroring SendKexDhInit) instead of only wc_ecc_init, so the only path to failure is the peer point validation rather than a missing private exponent.

tests/unit.c
```diff
@@ -3696,6 +3869,111 @@ static int test_KeyAgreeDh_client_zeroesEphemeralPrivKey(void)
...
+     * KEX path does before KeyAgreeDh_client runs. */
+    if (wolfSSH_TestSetDhKexKey(ssh) != 0) {
+        result = -752;
+        goto out;
+    }
+
+    /* Degenerate peer public key f = 1 (yields the shared secret 1). */
+    f[0] = 0x01;
+
+    if (wolfSSH_TestKeyAgreeDh_client(ssh, WC_HASH_TYPE_SHA256, f,
```
ejohnstown, line 5510: The proposed isolation isn't achievable against wolfCrypt as built: wc_DhAgree itself calls wc_DhCheckPubKey_ex unconditionally (dh.c:2055), and for the SSH DH groups (no q set) both that and our wc_DhCheckPubKey reduce to the identical 2 <= y <= p-2 range check. p-1 is rejected by wc_DhAgree for the same reason 1 is, so no peer value passes one and fails the other. The wolfSSH-level check is intentional defense-in-depth for builds/configs where wc_DhAgree may not validate; the test comment documents that it can't isolate the two against current wolfCrypt. Leaving the guard and the test as-is.

src/internal.c
```diff
@@ -13308,6 +13357,9 @@ static int KeyAgreeDh_server(WOLFSSH* ssh, byte hashId, byte* f, word32* fSz)
         if (ret == 0)
             ret = wc_DhSetKey(privKey, primeGroup, primeGroupSz,
                     generator, generatorSz);
+        if (ret == 0)
+            ret = wc_DhCheckPubKey(privKey, ssh->handshake->e,
```
aidangarske: why are we doing `wc_DhCheckPubKey` twice here?

src/internal.c
```diff
@@ -6135,7 +6136,8 @@ static int KeyAgreeDh_client(WOLFSSH* ssh, byte hashId,
      * validates the peer key inside wc_DhAgree. */
     ret = wc_DhCheckPubKey(&ssh->handshake->privKey.dh, f, fSz);
     if (ret != 0) {
-        WLOG(WS_LOG_ERROR, "Peer DH public value rejected, %d", ret);
+        WLOG(WS_LOG_ERROR,
```
aidangarske: Prefer `DEBUG` for all peer-driven rejects?

src/internal.c
```diff
@@ -6196,12 +6233,8 @@ static int KeyAgreeEcdh_client(WOLFSSH* ssh, byte hashId,
         key_ptr = &key_s;
     #endif /* WOLFSSH_SMALL_STACK */
     ret = wc_ecc_init(key_ptr);
-    #ifdef HAVE_WC_ECC_SET_RNG
-    if (ret == 0)
-        ret = wc_ecc_set_rng(key_ptr, ssh->rng);
-    #endif
     if (ret == 0)
-        ret = wc_ecc_import_x963(f, fSz, key_ptr);
+        ret = EccCheckPeerKey(key_ptr, f, fSz, ECC_CURVE_INVALID, ssh->rng);
```
aidangarske: Since `kexId` and `wcPrimeForId` are both already in scope, can we pass the real prime here instead of `ECC_CURVE_INVALID`? That closes the client/server asymmetry and matches the "validate peer KEX public keys" goal of the PR. If there's a reason the client must stay curve-agnostic, a one-line comment saying enforcement is intentionally deferred would be enough.

src/internal.c
```diff
@@ -6431,14 +6464,13 @@ static int KeyAgreeEcdhMlKem_client(WOLFSSH* ssh, byte hashId,
...
-    #endif
+        /* Import and validate the peer ECC point via the shared helper, the
+         * same path the plain KeyAgreeEcdh_client uses, so the off-curve
+         * rejection cannot diverge between the plain and hybrid code. The ECC
+         * point sits after the ML-KEM ciphertext in f. */
         if (ret == 0) {
-            ret = wc_ecc_import_x963(f + length_ciphertext,
-                    fSz - length_ciphertext, key_ptr);
+            ret = EccCheckPeerKey(key_ptr, f + length_ciphertext,
+                    fSz - length_ciphertext, ECC_CURVE_INVALID, ssh->rng);
```
aidangarske: Same as the plain path above: hybrid client passes ECC_CURVE_INVALID while the hybrid server enforces wcPrimeForId(ssh->handshake->kexId) (13755). If we tighten KeyAgreeEcdh_client, tighten this too so they stay symmetric.

src/internal.c
```diff
@@ -13308,6 +13357,9 @@ static int KeyAgreeDh_server(WOLFSSH* ssh, byte hashId, byte* f, word32* fSz)
         if (ret == 0)
             ret = wc_DhSetKey(privKey, primeGroup, primeGroupSz,
                     generator, generatorSz);
+        if (ret == 0)
```
aidangarske: **🟠 [Medium] Duplicate wc_DhCheckPubKey call in KeyAgreeDh_server (redundant peer-key range check)** · _Redundant code / Logic_

All three scan modes (review, review-security, bugs) independently flagged this and it is confirmed against the code at src/internal.c:13357-13370. The PR adds a wc_DhCheckPubKey(privKey, ssh->handshake->e, ssh->handshake->eSz) call at lines 13360-13362 (before wc_DhGenerateKeyPair), but an identical wc_DhCheckPubKey on the same peer value ssh->handshake->e already exists at lines 13368-13370 (the pre-existing master block with the '/* Reject a peer public value outside the safe range [2, p-2] ... */' comment). Both calls validate the same peer value against the same DH group; wc_DhGenerateKeyPair does not alter the prime p that wc_DhCheckPubKey inspects, so the second check is fully redundant and every DH server handshake performs the same bignum range check twice. There is NO correctness or security impact — a bad peer value is rejected either way — so severity views differ (review: Medium/SUGGEST; bugs: Low; review-security: Info), and per the stricter-view rule Medium is retained. This appears to be a rebase collision with master commit 90692e69 ('Validate peer DH public value before key agreement'). Note: the new test test_KeyAgreeDh_server_rejectsBadPeerPublic passes on the pre-existing check and cannot distinguish whether the PR's added line is present, so it does not actually exercise the added check.

**Fix:** Keep a single wc_DhCheckPubKey call. Prefer the earlier (pre-keygen) placement so a bad peer value is rejected before the server spends time generating its ephemeral key, and remove the now-redundant post-keygen block (or vice versa) to eliminate the duplicate. This is a rebase collision with master commit 90692e69.

tests/unit.c
```diff
@@ -5229,6 +5402,111 @@ static int test_KeyAgreeDh_client_zeroesEphemeralPrivKey(void)
...
+
+    ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL);
+    if (ctx == NULL)
+        return -730;
+    ssh = wolfSSH_new(ctx);
+    if (ssh == NULL || ssh->handshake == NULL) {
+        result = -731;
+        goto out;
+    }
+#ifndef WOLFSSH_NO_DH_GROUP14_SHA256
```
aidangarske, line 5427: **🔵 [Low] New DH tests skip group16-only configurations** · _Logic Errors_

The two new DH rejection tests select a KEX group with group14-sha256, then group14-sha1, then unconditionally group1-sha1, omitting ID_DH_GROUP16_SHA512 even though wolfSSH supports it and WOLFSSH_NO_DH stays unset when group16 is the only enabled fixed DH group. With WOLFSSH_NO_DH_GROUP14_SHA256 and WOLFSSH_NO_DH_GROUP14_SHA1 defined and WOLFSSH_NO_DH_GROUP16_SHA512 not defined, the client test falls through to ID_DH_GROUP1_SHA1; if group1 is disabled, wolfSSH_TestSetDhKexKey() returns WS_INVALID_ALGO_ID and the test fails before exercising the peer-key check (result -752). The server test has the same bad selection, but any nonzero return is treated as success (line 5439 only sets the failure result when the call returns 0), so it can silently pass without testing the PR's DH public-key validation. Verified against tests/unit.c:5427-5433 and 5478-5491: both #if ladders fall through to ID_DH_GROUP1_SHA1 with no group16 branch.

**Fix:** Include group16 in the fallback chain and guard group1 explicitly, matching the existing regress-test selection pattern:
#ifndef WOLFSSH_NO_DH_GROUP14_SHA256
    ssh->handshake->kexId = ID_DH_GROUP14_SHA256;
#elif !defined(WOLFSSH_NO_DH_GROUP16_SHA512)
    ssh->handshake->kexId = ID_DH_GROUP16_SHA512;
#elif !defined(WOLFSSH_NO_DH_GROUP14_SHA1)
    ssh->handshake->kexId = ID_DH_GROUP14_SHA1;
#elif !defined(WOLFSSH_NO_DH_GROUP1_SHA1)
    ssh->handshake->kexId = ID_DH_GROUP1_SHA1;
#else
    result = -752;
    goto out;
#endif

src/internal.c
```diff
@@ -6168,6 +6170,41 @@ static int KeyAgreeDh_client(WOLFSSH* ssh, byte hashId,
 #endif /* WOLFSSH_NO_DH */
 
 
+#if !defined(WOLFSSH_NO_ECDH) || \
+    !defined(WOLFSSH_NO_NISTP256_MLKEM768_SHA256) || \
+    !defined(WOLFSSH_NO_NISTP384_MLKEM1024_SHA384)
+/* Import a peer's X9.63 ECC point and validate it before it is used to derive
```
aidangarske: **🔵 [Low] Multi-line explanatory comment block longer than repo norm on EccCheckPeerKey** · _Style_

The new EccCheckPeerKey helper carries a 10-line prose header comment (lines 6176-6185) explaining history/rationale (matching the client's behavior, independent of the linked wolfSSL, defense-in-depth notes) at a length better suited to the commit message or PR description. This conflicts with the repo's one-line comment norm, which discourages multi-line function-header blocks. Purely cosmetic; no correctness impact.

**Fix:** Trim the header comment to one or two lines capturing the WHY, and move the extended rationale to the commit message / PR body.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 4 total — 3 posted, 1 skipped

### Posted findings
- [Medium] [review+review-security+bugs] **Duplicate wc_DhCheckPubKey call in KeyAgreeDh_server (redundant peer-key range check)** — `src/internal.c:13360-13370`
- [Low] [bugs] **New DH tests skip group16-only configurations** — `tests/unit.c:5427-5489`
- [Low] [review] **Multi-line explanatory comment block longer than repo norm on EccCheckPeerKey** — `src/internal.c:6176-6185`

### Skipped findings
- [Info] ~~ML-KEM hybrid ECC peer-key validation not directly unit-tested~~

src/internal.c
aidangarske: Already commented this before

src/internal.c
ejohnstown: Mistake from rushing. Fixed.

src/internal.c
```diff
@@ -13308,6 +13357,9 @@ static int KeyAgreeDh_server(WOLFSSH* ssh, byte hashId, byte* f, word32* fSz)
         if (ret == 0)
             ret = wc_DhSetKey(privKey, primeGroup, primeGroupSz,
                     generator, generatorSz);
+        if (ret == 0)
+            ret = wc_DhCheckPubKey(privKey, ssh->handshake->e,
+                    ssh->handshake->eSz);
```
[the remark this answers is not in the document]
ejohnstown: Fixed.

src/internal.c
ejohnstown: Fixed.

src/internal.c
ejohnstown: Fixed.

src/internal.c
ejohnstown: Trimmed.

tests/unit.c
ejohnstown: Done.

src/internal.c
ejohnstown: Good idea. Updated.

src/internal.c
ejohnstown: Good idea. Updated.

src/internal.c
ejohnstown: How did that get in there? Updated.

aidangarske (dismissed):

tests/unit.c
```diff
@@ -5544,6 +5848,129 @@ static int test_KeyAgreeDh_client_rejectsOutOfRangePeer(void)
...
+     * missing private key. KeyAgreeEcdh_client() frees privKey.ecc on the way
+     * out, so the cleanup path stays well defined. */
+    if (wc_ecc_init(&ssh->handshake->privKey.ecc) != 0) {
+        result = -772;
+        goto out;
+    }
+    ssh->handshake->useEcdh = 1;
+#ifdef HAVE_WC_ECC_SET_RNG
+    if (wc_ecc_set_rng(&ssh->handshake->privKey.ecc, ssh->rng) != 0) {
+        result = -772;
```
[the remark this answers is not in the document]
ejohnstown: Changed.

aidangarske (approved):

PR #1050, opened 2026-06-23
feat: add make sbom target

MarkAtwood:
## Summary

Adds `make sbom`, `make install-sbom`, and `make uninstall-sbom` targets to the autotools build.

- Stages a temporary install to extract the built `.so` for hashing
- Dumps preprocessor defines via `cc -dM -E` for the options snapshot
- Calls `gen-sbom` (from wolfssl `scripts/gen-sbom`) to produce CycloneDX and SPDX outputs
- Validates SPDX with `pyspdxtools`

## Usage

\`\`\`sh
./configure --with-wolfssl=/path/to/wolfssl/install
make
make sbom WOLFSSL_DIR=/path/to/wolfssl/source
\`\`\`

Outputs: `wolfssh-<version>.cdx.json`, `wolfssh-<version>.spdx.json`, `wolfssh-<version>.spdx`

## Requirements

- `WOLFSSL_DIR` must point to a wolfssl source tree containing `scripts/gen-sbom` (available on the `feat/sbom-embedded` branch of wolfssl)
- `python3` and `pip install spdx-tools` on the build host

## Test plan
- [ ] `./configure --with-wolfssl=... && make && make sbom WOLFSSL_DIR=...`
- [ ] Verify three output files are produced
- [ ] `pyspdxtools --infile wolfssh-*.spdx.json` passes validation

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ 'master', 'main', 'release/**' ]
   pull_request:
-    branches: [ '*' ]
+    branches: [ '**' ]
   workflow_dispatch:
     inputs:
       wolfssl_ref:
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -1,0 +1,255 @@
+name: SBOM Test
+
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+  workflow_dispatch:
+    inputs:
+      wolfssl_ref:
+        description: 'wolfssl git ref that provides scripts/gen-sbom'
+        default: 'master'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# This workflow only reads the repo and uploads artefacts; no API writes.
+permissions:
+  contents: read
+
+jobs:
+  sbom:
+    name: wolfSSH SBOM generation (linux)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout wolfssh
+        uses: actions/checkout@v4
+        with:
+          path: wolfssh
+
+      # wolfssl is checked out once and used for two things: built + installed
+      # so wolfssh has a library to link, and its source tree (scripts/gen-sbom
+      # + wolfssl/version.h) is passed to `make sbom` via WOLFSSL_DIR.  Default
+      # ref is master; until the wolfSSL SBOM change is on master, use the
+      # "Run workflow" button with wolfssl_ref set to the SBOM branch.
+      - name: Checkout wolfssl (gen-sbom + library source)
+        uses: actions/checkout@v4
+        with:
+          repository: wolfSSL/wolfssl
+          ref: ${{ github.event.inputs.wolfssl_ref || 'master' }}
+          path: wolfssl
+
+      - name: Install SBOM validator (pyspdxtools) and pcpp
+        run: |
+          # spdx-tools -> pyspdxtools (validation); pcpp -> the embedded
+          # (--user-settings) path's C preprocessor for walking user_settings.h.
+          python3 -m pip install --user 'spdx-tools==0.8.*' pcpp
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Build and install wolfssl
+        working-directory: wolfssl
+        run: |
+          autoreconf -ivf
+          ./configure --enable-all --prefix="$GITHUB_WORKSPACE/wolfssl-install"
+          make -j"$(nproc)"
+          make install
+
+      # gen-sbom lives in wolfssl and may not be on the checked-out ref yet
+      # (the wolfSSL SBOM change can land separately).  Gate on its presence so
+      # this workflow is safe to merge before that: it stays green and simply
+      # skips SBOM generation until a ref that carries the script is used.
+      - name: Detect gen-sbom availability and capabilities
+        id: gate
+        run: |
+          GS="$GITHUB_WORKSPACE/wolfssl/scripts/gen-sbom"
+          if [ ! -f "$GS" ]; then
+            echo "have=no" >> "$GITHUB_OUTPUT"
+            echo "::notice::wolfssl scripts/gen-sbom not present on this ref; skipping SBOM generation. Re-run via 'Run workflow' with wolfssl_ref set to a branch that has it until it merges to wolfssl master."
+            exit 0
+          fi
+          echo "have=yes" >> "$GITHUB_OUTPUT"
+          if python3 "$GS" --help 2>/dev/null | grep -q -- '--dep-wolfssl'; then
+            echo "dep_wolfssl=yes" >> "$GITHUB_OUTPUT"
+          else
+            echo "dep_wolfssl=no" >> "$GITHUB_OUTPUT"
+            echo "::notice::gen-sbom on this ref has no --dep-wolfssl; the wolfssl dependency + name-derived identity assertions will be skipped (wolfSSL SBOM change not merged yet)."
+          fi
+
+      - name: Configure and build wolfssh
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          autoreconf -ivf
+          ./configure --with-wolfssl="$GITHUB_WORKSPACE/wolfssl-install"
+          make -j"$(nproc)"
+
+      - name: Generate SBOM
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: make sbom WOLFSSL_DIR="$GITHUB_WORKSPACE/wolfssl"
+
+      - name: Outputs exist and SPDX validates
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          ls wolfssh-*.cdx.json wolfssh-*.spdx.json wolfssh-*.spdx
+          pyspdxtools --infile wolfssh-*.spdx.json
+
+      - name: CycloneDX is valid JSON with expected identity
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          python3 - <<'PY'
+          import glob, json
+          cdx = json.load(open(glob.glob('wolfssh-*.cdx.json')[0]))
+          assert cdx['bomFormat'] == 'CycloneDX', cdx.get('bomFormat')
+          assert cdx['specVersion'] == '1.6', cdx.get('specVersion')
+          m = cdx['metadata']['component']
+          assert m['name'] == 'wolfssh', m['name']
+          assert m['purl'].startswith('pkg:github/wolfSSL/wolfssh@'), m['purl']
+          props = {p['name'] for p in m.get('properties', [])}
+          # The AM_CPPFLAGS/config.h snapshot must capture real build config;
+          # PACKAGE_VERSION always lands from config.h, so its absence means
+          # the options snapshot regressed to empty.
+          assert any(n.startswith('wolfssl:build:') for n in props), \
+              'no wolfssl:build:* properties - options snapshot is empty'
+          print('CDX identity ok:', m['name'], m['purl'])
+          PY
+
+      - name: Reproducible across two runs (SOURCE_DATE_EPOCH)
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          rm -f wolfssh-*.cdx.json wolfssh-*.spdx.json wolfssh-*.spdx
+          SOURCE_DATE_EPOCH=1700000000 make sbom \
+              WOLFSSL_DIR="$GITHUB_WORKSPACE/wolfssl"
+          sha256sum wolfssh-*.cdx.json wolfssh-*.spdx.json > /tmp/a.sums
+          rm -f wolfssh-*.cdx.json wolfssh-*.spdx.json wolfssh-*.spdx
+          SOURCE_DATE_EPOCH=1700000000 make sbom \
+              WOLFSSL_DIR="$GITHUB_WORKSPACE/wolfssl"
+          sha256sum wolfssh-*.cdx.json wolfssh-*.spdx.json > /tmp/b.sums
+          diff /tmp/a.sums /tmp/b.sums
+
+      - name: wolfssl recorded as a dependency + wolfssh identity
+        if: steps.gate.outputs.have == 'yes' && steps.gate.outputs.dep_wolfssl == 'yes'
+        working-directory: wolfssh
+        run: |
+          python3 - <<'PY'
+          import glob, json
+          d = json.load(open(glob.glob('wolfssh-*.spdx.json')[0]))
+          pkgs = {p['name']: p for p in d['packages']}
+          assert 'wolfssl' in pkgs, list(pkgs)
+          main = pkgs['wolfssh']
+          assert main['SPDXID'] == 'SPDXRef-Package-wolfssh', main['SPDXID']
+          assert 'github.com/wolfSSL/wolfssh' in main['downloadLocation'], \
+              main['downloadLocation']
+          rels = [(r['spdxElementId'], r['relationshipType'],
+                   r['relatedSpdxElement']) for r in d['relationships']]
+          assert ('SPDXRef-Package-wolfssh', 'DEPENDS_ON',
+                  'SPDXRef-Package-wolfssl') in rels, rels
+          print('wolfssl dependency + wolfssh identity ok')
+          PY
+
+      # ---- Embedded / IDE path (no autotools) ----------------------------
+      # Firmware customers don't run ./configure: there's no options.h and no
+      # installed libwolfssh to hash.  gen-sbom reads config from user_settings.h
+      # (via pcpp) and hashes the wolfSSH source set instead.  This exercises
+      # that path the same way the docs tell customers to invoke it.
+      - name: Generate embedded SBOM (user_settings.h + source set)
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          VER=$(sed -n 's/^AC_INIT(\[wolfssh\],\[\([^]]*\)\].*/\1/p' configure.ac)
+          VER=${VER:-0.0.0}
+          mkdir -p sbom-embedded/cfg sbom-embedded-2
+          # A minimal embedded-style config; the assertions below prove these
+          # #defines survive pcpp and land as build properties in the SBOM.
+          {
+            echo '#ifndef USER_SETTINGS_H'
+            echo '#define USER_SETTINGS_H'
+            echo '#define WOLFSSH_TERM'
+            echo '#define WOLFSSH_SFTP'
+            echo '#define WOLFSSH_SCP'
+            echo '#endif'
+          } > sbom-embedded/cfg/user_settings.h
+          DEP=()
+          if [ "${{ steps.gate.outputs.dep_wolfssl }}" = "yes" ]; then
+            DEP+=(--dep-wolfssl yes)
+          fi
+          gen() {
+            SOURCE_DATE_EPOCH=1700000000 python3 \
+              "$GITHUB_WORKSPACE/wolfssl/scripts/gen-sbom" \
+              --name wolfssh --version "$VER" \
+              --license-file LICENSING \
+              --user-settings wolfssh/settings.h \
+              --user-settings-include . \
+              --user-settings-include "$GITHUB_WORKSPACE/wolfssl" \
+              --user-settings-include sbom-embedded/cfg \
+              --user-settings-define WOLFSSL_USER_SETTINGS \
+              --srcs src/*.c \
+              "${DEP[@]}" \
+              --cdx-out "$1/wolfssh-embedded.cdx.json" \
+              --spdx-out "$1/wolfssh-embedded.spdx.json"
+          }
+          gen sbom-embedded
+          gen sbom-embedded-2
+          # Reproducible: identical bytes across runs (SOURCE_DATE_EPOCH fixed,
+          # namespace is uuid5(name,version), source-set hash is path-independent).
+          a=$(sha256sum < sbom-embedded/wolfssh-embedded.cdx.json)
+          b=$(sha256sum < sbom-embedded-2/wolfssh-embedded.cdx.json)
+          test "$a" = "$b" || { echo "embedded SBOM not reproducible"; exit 1; }
+
+      - name: Embedded SBOM validates + reflects user_settings.h + source hash
+        if: steps.gate.outputs.have == 'yes'
+        working-directory: wolfssh
+        run: |
+          pyspdxtools --infile sbom-embedded/wolfssh-embedded.spdx.json
+          python3 - <<'PY'
+          import json
+          cdx = json.load(open('sbom-embedded/wolfssh-embedded.cdx.json'))
+          m = cdx['metadata']['component']
+          assert m['name'] == 'wolfssh', m['name']
+          assert m['purl'].startswith('pkg:github/wolfSSL/wolfssh@'), m['purl']
+          # Embedded identity is the source-set hash (no library artifact exists).
+          algs = {h['alg'] for h in m.get('hashes', [])}
+          assert algs, 'no component hash - source-set (--srcs) hash missing'
+          # Config must come from user_settings.h through pcpp, not be empty.
+          props = {p['name'] for p in m.get('properties', [])}
+          assert any(n.endswith('WOLFSSH_SFTP') for n in props), \
+              'user_settings.h option WOLFSSH_SFTP not captured: %r' % sorted(props)
+          print('embedded ok:', m['name'], m['purl'],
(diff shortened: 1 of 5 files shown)


ejohnstown (changes requested):
Tested make sbom end-to-end on macOS (--enable-all, gen-sbom from wolfssl feat/sbom-embedded, spdx-tools + pcpp installed): exits 0, all three outputs produced, the Mach-O .dylib discovery branch fires, the trap cleans up the staging tree and temp defines file, the configured feature macros (WOLFSSH_SFTP/SCP/CERTS/AGENT/SSHD, NO_INLINE, NDEBUG) now appear in the SBOM, and pyspdxtools re-validation passes clean. Two follow-ups below, none blocking.

Makefile.am
```diff
@@ -79,3 +79,145 @@ merge-clean:
...
+	else \
+	    echo "NOTE: this gen-sbom has no --dep-wolfssl; the SBOM will not list"; \
+	    echo "      wolfssl as a dependency component. Point WOLFSSL_DIR at a"; \
+	    echo "      wolfSSL tree whose scripts/gen-sbom supports it to enable it."; \
+	fi; \
+	$(PYTHON3) $(WOLFSSL_DIR)/scripts/gen-sbom \
+	    --name wolfssh \
+	    --version $(PACKAGE_VERSION) \
+	    --supplier "wolfSSL Inc." \
+	    --license-file $(srcdir)/LICENSING \
```
ejohnstown: gen-sbom emits NOASSERTION for the license even though LICENSING clearly states GPLv3. Root cause is in gen-sbom's detect_license: its regex only matches the long form gnu general public license version N, but wolfSSH's LICENSING uses the abbreviation GPLv3, so detection fails and falls back to NOASSERTION.

Makefile.am
```diff
@@ -79,3 +79,145 @@ merge-clean:
...
+	    wolfssl_ver=`sed -n 's/.*LIBWOLFSSL_VERSION_STRING[ \t]*"\([^"]*\)".*/\1/p' \
+	        "$(WOLFSSL_DIR)/wolfssl/version.h"`; \
+	fi; \
+	if test -z "$${SOURCE_DATE_EPOCH:-}" && test -n "$(GIT)" && \
+	   $(GIT) -C "$(srcdir)" rev-parse --git-dir >/dev/null 2>&1; then \
+	    sde=`$(GIT) -C "$(srcdir)" log -1 --format=%ct 2>/dev/null`; \
+	    if test -n "$$sde"; then SOURCE_DATE_EPOCH="$$sde"; export SOURCE_DATE_EPOCH; fi; \
+	fi; \
+	dep_args=""; \
+	if $(PYTHON3) $(WOLFSSL_DIR)/scripts/gen-sbom --help 2>/dev/null \
```
ejohnstown: `--dep-wolfssl` doesn't exist yet in the gen-sbom script. Maybe a better description of what to do since there is no wolfSSL tree whose scripts/gen-sbom supports the opotion.

ejohnstown (changes requested):

Makefile.am
```diff
@@ -79,3 +80,21 @@ merge-clean:
...
+SBOM_LICENSE_FILE = $(srcdir)/LICENSING
+SBOM_DEP_WOLFSSL  = yes
+
+# wolfSSH is GPLv3-or-commercial; its LICENSING uses the "GPLv3" abbreviation
+# that older gen-sbom cannot parse (falls back to NOASSERTION).  Default to the
+# correct SPDX id so the SBOM is right regardless of gen-sbom version; commercial
+# licensees can still override it (e.g. LicenseRef-wolfSSL-Commercial).
+# Redundant-but-harmless once the detect_license fix lands in gen-sbom
+# (wolfSSL/wolfssl#10343).
+SBOM_LICENSE_OVERRIDE ?= GPL-3.0-only
```
aidangarske: **🟠 [High] Default SBOM license contradicts source license headers**
*🚫 BLOCK*  `bug`

The PR hard-codes the default SBOM license override to `GPL-3.0-only`, but wolfSSH source headers state the code is licensed under GPL version 3 "or (at your option) any later version." That maps to `GPL-3.0-or-later`, not `GPL-3.0-only`. Because this override is passed to `gen-sbom`, every default `make sbom` run will emit materially incorrect license metadata.

**Suggestion:**
```suggestion
SBOM_LICENSE_OVERRIDE ?= GPL-3.0-or-later
```

aidangarske (dismissed):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Default SBOM license contradicts source license headers** — `Makefile.am:98`
- [Medium] **SBOM workflow can pass without exercising the new target** — `.github/workflows/sbom.yml:64-79`

*Review generated by Skoll.*

jackctj117 (commented):
## Skoll Multi-Scan Review

**Modes:** `review`**Overall recommendation:** `COMMENT`
**1 finding(s) not tied to a diff line** (full detail below)

> One or more scans did not complete — the results below are partial.

### Posted findings
- [Medium] [review] **CI workflow pins wolfSSL to an unmerged PR ref for all events, including every pull_request** — `.github/workflows/sbom.yml:3-13, 41-46`
- [Low] [review] **wolfssl version.h parse uses [ \t] which does not match a literal tab** — `scripts/sbom.am:144-145`
- [Low] [review] **sbomdir hardcodes $(datadir)/doc/$(PACKAGE) instead of using $(docdir)** — `scripts/sbom.am:55`
- [Low] [review] **uninstall-hook removes SBOM files that plain make install never installs** — `scripts/sbom.am:185-196`

### Findings not tied to a diff line

#### SBOM targets silently require GNU make; ?= and GNU-only make functions are undocumented
**File:** `scripts/sbom.am:46-63, Makefile.am:98`
**Function:** `sbom`
**Severity:** Medium
**Category:** question

The fragment relies on GNU make extensions: conditional assignment `?=` (`SBOM_ARTIFACT ?= lib`, `SBOM_LICENSE_OVERRIDE ?= GPL-3.0-or-later`) and the GNU-only functions `$(wildcard ...)`, `$(if ...)`, `$(firstword ...)`, and `$(addprefix ...)`. `?=` is not one of Automake's documented assignment operators (`=`, `+=`, `:=`); confirm `autoreconf` accepts it without error. Note CI only runs `autoreconf`/`configure` on the wolfSSH tree when `gen-sbom` is detected (the `steps.gate.outputs.have == 'yes'` gate), so a hard Automake failure would only surface when CI checks out a wolfssl ref carrying the script. Under a non-GNU make (e.g. BSD make), `SBOM_GEN`/`SBOM_LIB_GLOBS` evaluate to empty and `make sbom` fails with a misleading 'gen-sbom not found' / 'no artifact found' error rather than a clear 'GNU make required' message.

**Recommendation:** Confirm autoreconf accepts `?=` on the toolchain versions you support, and document in the README/comment that the SBOM targets require GNU make. Optionally guard with a clearer error when GNU-make features are unavailable.

_Referenced code: `scripts/sbom.am:46-63, Makefile.am:98-100` (3 lines)_

---

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,258 @@
+name: SBOM Test
+
+on:
```
jackctj117, line 3: **🟠 [Medium] CI workflow pins wolfSSL to an unmerged PR ref for all events, including every pull_request** · _question_

The workflow triggers on `push` (master/main/release/**) and `pull_request: branches: ['*']`, and the wolfssl checkout defaults/falls back to `refs/pull/10343/head` for every non-dispatch run (`${{ github.event.inputs.wolfssl_ref || 'refs/pull/10343/head' }}`). This means: (1) every PR to the repo builds a full `--enable-all` wolfssl plus wolfssh — a heavy job on unrelated PRs; and (2) CI is permanently pinned to an unmerged PR head and never exercises wolfssl master. The two TODOs acknowledge this, but there is no forcing function: after #10343 merges, CI will keep testing the stale PR ref until someone manually updates it, and the SBOM feature could silently regress against master.

**Fix:** Track the #10343 merge with an issue/reminder so the fallback flips back to 'master'. Consider narrowing the pull_request trigger (e.g. paths filter on Makefile.am/scripts/sbom.am/configure.ac) to avoid running the heavy build on unrelated PRs.

scripts/sbom.am
```diff
@@ -0,0 +1,196 @@
...
+	    if test -n "$$sde"; then SOURCE_DATE_EPOCH="$$sde"; export SOURCE_DATE_EPOCH; fi; \
+	fi; \
+	dep_args=""; \
+	if test "$(SBOM_DEP_WOLFSSL)" = yes; then \
+	    if $(PYTHON3) "$(SBOM_GEN)" --help 2>/dev/null \
+	        | $(GREP) -q -- '--dep-wolfssl'; then \
+	        dep_args="$$dep_args --dep-wolfssl yes"; \
+	        wv="$(SBOM_WOLFSSL_VERSION)"; \
+	        if test -z "$$wv" && test -f "$(WOLFSSL_DIR)/wolfssl/version.h"; then \
+	            wv=`sed -n 's/.*LIBWOLFSSL_VERSION_STRING[ \t]*"\([^"]*\)".*/\1/p' \
```
jackctj117: **🔵 [Low] wolfssl version.h parse uses [ \t] which does not match a literal tab** · _bug_

The sed program `s/.*LIBWOLFSSL_VERSION_STRING[ \t]*"\([^"]*\)".*/\1/p` is passed literally to sed (make does not interpret `\t`). Inside a bracket expression, `[ \t]` matches a space, a backslash, or the letter 't' — not a tab character. It works for the standard `#define LIBWOLFSSL_VERSION_STRING "x.y.z"` (space-separated), but if a wolfssl `version.h` ever used a tab after the macro name, auto-detection would silently yield an empty version. Low impact since wolfssl's header uses a space.

**Fix:** Use `[[:space:]]*` (or an actual tab) instead of `[ \t]*` so tab-separated defines are also matched.

scripts/sbom.am
```diff
@@ -0,0 +1,196 @@
...
+SBOM_ARTIFACT ?= lib
+SBOM_LIB_STEM ?= lib$(SBOM_PKGNAME)
+SBOM_BIN_NAME ?= $(SBOM_PKGNAME)
+SBOM_DEP_WOLFSSL ?= no
+SBOM_DEP_OPENSSL ?= no
+
+SBOM_CDX     = $(SBOM_PKGNAME)-$(PACKAGE_VERSION).cdx.json
+SBOM_SPDX    = $(SBOM_PKGNAME)-$(PACKAGE_VERSION).spdx.json
+SBOM_SPDX_TV = $(SBOM_PKGNAME)-$(PACKAGE_VERSION).spdx
+sbomdir      = $(datadir)/doc/$(PACKAGE)
```
jackctj117: **🔵 [Low] sbomdir hardcodes $(datadir)/doc/$(PACKAGE) instead of using $(docdir)** · _convention_

`sbomdir = $(datadir)/doc/$(PACKAGE)` reconstructs the doc path rather than using Automake's `$(docdir)`. This equals the default docdir but ignores a user's `--docdir=` override. The existing Makefile.am convention uses `$(docdir)` (`exampledir = $(docdir)/example`), so this is a minor inconsistency with the project's own layout. Note this file is described as a canonical copy shared across wolfSSL products, so a change here should be coordinated with the upstream copy.

**Fix:** Consider `$(docdir)` for consistency with `exampledir` and to honor `--docdir`; coordinate with the canonical wolfSSL copy if changed.

scripts/sbom.am
```diff
@@ -0,0 +1,196 @@
...
+	    --lib "$$sbom_art" \
+	    $$dep_args \
+	    $(if $(SBOM_LICENSE_OVERRIDE),--license-override '$(SBOM_LICENSE_OVERRIDE)') \
+	    $(if $(SBOM_LICENSE_TEXT),--license-text '$(SBOM_LICENSE_TEXT)') \
+	    --cdx-out $(abs_builddir)/$(SBOM_CDX) \
+	    --spdx-out $(abs_builddir)/$(SBOM_SPDX); \
+	$(PYSPDXTOOLS) --infile $(abs_builddir)/$(SBOM_SPDX) \
+	    --outfile $(abs_builddir)/$(SBOM_SPDX_TV)
+
+install-sbom: sbom
```
jackctj117, line 239: **🔵 [Low] uninstall-hook removes SBOM files that plain make install never installs** · _convention_

`uninstall-hook: uninstall-sbom` wires SBOM removal into the standard `make uninstall`, but there is no matching `install-data-hook`/`install-exec-hook` for `install-sbom` (SBOM install is opt-in via `make install-sbom`). The result is asymmetric: `make install` does not place SBOM files, yet `make uninstall` always tries to remove them. It is harmless because `uninstall-sbom` uses `-rm -f`, but the asymmetry can be surprising. (Correctly, `install-sbom` is intentionally NOT hooked into `install`, which avoids the `install-sbom -> sbom -> make install` recursion — good.)

**Fix:** Either document the intentional asymmetry in the comment, or drop the automatic `uninstall-hook` chaining so install/uninstall of the SBOM stay symmetric (both opt-in).

.github/workflows/sbom.yml
sameehj: Ack, keeping this

scripts/sbom.am
sameehj: Fixed

scripts/sbom.am
sameehj: Fixed

scripts/sbom.am
sameehj: Fixed

.github/workflows/sbom.yml
MarkAtwood: Resolving per Sameeh: deliberate and temporary. gen-sbom exists only on wolfSSL/wolfssl#10343 until it merges, the workflow gates on the script being present (skips cleanly otherwise), permissions are contents: read, and the workflow carries the switch-back-to-master TODO for when #10343 lands.

ejohnstown (changes requested):
I guess my only issue is this PR depends on wolfSSL's PR 10343 to be merged first. The GitHub Action works without the PR merged, so that's OK. But someone may see the note about the `make sbom` and it not work if wolfSSH is released with this without the wolfSSL merge.

Makefile.am
```diff
@@ -79,3 +79,74 @@ merge-clean:
...
+
+.PHONY: sbom install-sbom uninstall-sbom
+
+sbom:
+	@if test -z "$(PYTHON3)"; then \
+	    echo ""; \
+	    echo "ERROR: 'python3' not found in PATH. Cannot generate SBOM."; \
+	    echo ""; \
+	    exit 1; \
+	fi
```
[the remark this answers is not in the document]
MarkAtwood: `AC_PATH_PROG` with no third argument leaves the variable empty, it does not set it to `no`. configure.ac:315 passes no third argument, so `test -n "$(PYTHON3)"` catches the not-found case. No change needed.

Makefile.am
```diff
@@ -79,3 +79,74 @@ merge-clean:
...
+	@if test -z "$(PYSPDXTOOLS)"; then \
+	    echo ""; \
+	    echo "ERROR: 'pyspdxtools' not found in PATH. Cannot validate SBOM."; \
+	    echo "       Install: pip install spdx-tools"; \
+	    echo ""; \
+	    exit 1; \
+	fi
```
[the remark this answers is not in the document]
MarkAtwood: Same as above. configure.ac:316 has no third argument, so the variable is empty when pyspdxtools is missing and the `-n` guard fires correctly.

Makefile.am
```diff
@@ -79,3 +79,74 @@ merge-clean:
...
+	    echo ""; \
+	    echo "ERROR: $(WOLFSSL_DIR)/scripts/gen-sbom not found."; \
+	    echo "       Use a wolfSSL tree that includes SBOM support."; \
+	    echo ""; \
+	    exit 1; \
+	fi
+	rm -rf $(abs_builddir)/_sbom_staging $(abs_builddir)/_sbom_defines.h
+	$(MAKE) install DESTDIR=$(abs_builddir)/_sbom_staging
+	$(CC) -dM -E -I$(srcdir) $(CPPFLAGS) -x c /dev/null \
+	    > $(abs_builddir)/_sbom_defines.h
```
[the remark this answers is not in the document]
MarkAtwood: Correct, fixed. The recipe now lives in `scripts/sbom.am` and the capture is `$(CC) -dM -E $(DEFAULT_INCLUDES) $(AM_CPPFLAGS) $(CPPFLAGS)`. See sbom.am:156.

Makefile.am
MarkAtwood: Fixed. Now `SBOM_LICENSE_OVERRIDE ?= GPL-3.0-or-later`, with a comment above it recording why the LICENSING abbreviation misleads `detect_license`. Commercial licensees can still override.

Makefile.am
MarkAtwood: Fixed. Pinning `SBOM_LICENSE_OVERRIDE` to `GPL-3.0-or-later` means wolfSSH emits the header-accurate id regardless of what `detect_license` does with the `GPLv3` abbreviation.

Makefile.am
MarkAtwood: `sbom.am` probes `gen-sbom --help` before passing `--dep-wolfssl`, so it degrades cleanly against a wolfssl tree that predates the option. See sbom.am:168 and :188.

MarkAtwood:
Note on the review threads above: the recipe moved out of `Makefile.am` into `scripts/sbom.am`, so the line numbers in the older comments now point at the wrong file. `Makefile.am` keeps five declaration variables and an `include scripts/sbom.am`.

ejohnstown (commented):
Re-reviewed at `7e22d45a` (+1 commit since my last pass: "fix: unpin SBOM CI from merged wolfssl PR ref").

Built and ran everything on macOS/arm64 (gcc-15, GNU make 3.81): `--enable-all` and plain `./configure` against a wolfSSL 5.9.2 install, `WOLFSSL_DIR` pointing at a wolfssl checkout at `origin/master` (`b83f5d60f`), `--prefix=/opt/wolfssh-sbom`. The prefix is root-owned here, so the install was exercised through `DESTDIR` staging of the identical recipe.

## Fixed: CI/docs unpinning

Verified `master` everywhere -- workflow input default, checkout `ref`, and both skip-notice strings -- plus the README and the `sbom.am` NOTE. wolfSSL master's `gen-sbom` is what I tested against throughout, exit 0.

The new "not an install prefix" error is a real improvement and fires correctly:

```
$ make sbom WOLFSSL_DIR=/opt/wolfssl
ERROR: no gen-sbom at /opt/wolfssl/scripts/gen-sbom
       WOLFSSL_DIR must name a wolfSSL *source tree*, not an
       install prefix such as /usr/local or /opt/wolfssl.
make: *** [sbom] Error 1
```

## New: the vendored `scripts/sbom.am` has diverged from the canonical copy

That fix was made by editing `scripts/sbom.am` in this repo only. The file's own header says:

> This is the canonical copy (wolfSSL repository, scripts/sbom.am); product repositories vendor a copy of it and must be kept in sync with this file.

sha256 of `scripts/sbom.am`:

```
wolfssl/scripts/sbom.am   282fd7ca...
wolfssh, previous head    282fd7ca...   (identical)
wolfssh, this head        5118dc2e...   (diverged)
```

Two hunks are wolfSSH-only now: the `WOLFSSL_DIR`-aware error block and the reworded `--dep-wolfssl` NOTE. A wolfssl checkout at `origin/master` does not carry them, and I found no open wolfSSL PR syncing them. Both changes are product-agnostic and belong upstream: land them in wolfSSL first, re-vendor here, and the file goes back to byte-identical. A CI check that diffs the vendored copy against the canonical one would stop this drifting again.

## Still open

### 1. The recorded wolfSSL dependency version does not track the linked wolfSSL

This is the one I would hold on. `SBOM_WOLFSSL_VERSION` falls back to `$(WOLFSSL_DIR)/wolfssl/version.h`, but `WOLFSSL_DIR` is only there to locate `gen-sbom` and is a different input from `--with-wolfssl`. Built and linked against wolfSSL 5.9.2, with `WOLFSSL_DIR` pointing at a tree whose `version.h` says 9.9.9:

```
component: wolfssl 9.9.9 pkg:github/wolfSSL/wolfssl@v9.9.9
```

Worse, with `gen-sbom` present but no `wolfssl/version.h` under `WOLFSSL_DIR`, `make sbom` exits 0 and emits:

```
component: wolfssl None None
```

No version, no purl, no warning -- silently unscannable, which defeats the README's stated reason for the entry ("so vulnerability scanners can associate wolfSSL advisories with a wolfSSH deployment"). `gen-sbom`'s pkg-config fallback does not rescue it, because a non-default prefix is not on the default `PKG_CONFIG_PATH` and the recipe does not set one.

The new error text sharpens the case: it tells users `WOLFSSL_DIR` is a *source tree* and explicitly not an install prefix -- i.e. not the thing you linked. Those are two independent inputs and only the wrong one feeds the dependency version.

Fixable entirely on the wolfSSH side through the documented `SBOM_WOLFSSL_VERSION` hook, so the canonical `sbom.am` does not need to change. I prototyped and verified this:

```m4
# configure.ac, after AC_CHECK_LIB([wolfssl],[wolfCrypt_Init],...)
# Record the version of the wolfSSL actually being linked, for the SBOM's
# dependency entry.  WOLFSSL_DIR (used by `make sbom` only to locate gen-sbom)
# may be an unrelated source tree, so it must not be the version source.
AC_MSG_CHECKING([wolfSSL version for SBOM])
WOLFSSL_VERSION=
if test -f "${wcpath}/include/wolfssl/version.h"; then
  WOLFSSL_VERSION=`sed -n ['s/.*LIBWOLFSSL_VERSION_STRING[ 	]*"\([^"]*\)".*/\1/p'] "${wcpath}/include/wolfssl/version.h"`
fi
AS_IF([test -n "$WOLFSSL_VERSION"],
      [AC_MSG_RESULT([$WOLFSSL_VERSION])],
      [AC_MSG_RESULT([unknown])])
AC_SUBST([WOLFSSL_VERSION])
```

```make
# Makefile.am, after SBOM_DEP_WOLFSSL = yes
# Version of the linked wolfSSL, detected by configure from --with-wolfssl.
SBOM_WOLFSSL_VERSION = @WOLFSSL_VERSION@
```

With that applied it records 5.9.2 even when `WOLFSSL_DIR` claims 9.9.9, and an explicit `make sbom SBOM_WOLFSSL_VERSION=...` still overrides.

### 2. The CI regression guard does not guard

```python
assert any(n.startswith('wolfssl:build:') for n in props), \
    'no wolfssl:build:* properties - options snapshot is empty'
```

The comment says this catches the options snapshot regressing to empty -- the bug fixed by "scripts/sbom.am: capture AM_CFLAGS/CFLAGS and configurable config.h". It does not. Reconfiguring plainly (exactly as CI does) and reintroducing that regression, i.e. dropping `AM_CPPFLAGS` from the `-dM -E` capture:

```
BASELINE  wolfssl:build:* count = 50   WOLFSSH_TERM present = True
REGRESSED wolfssl:build:* count = 48   WOLFSSH_TERM present = False
CI assertion any(startswith 'wolfssl:build:') -> True
```

config.h alone always supplies ~50 properties, so the assertion passes while the compiler-side capture is silently broken. Since CI configures plainly, `WOLFSSH_TERM` is the only marker that comes via `AM_CPPFLAGS`; asserting on `wolfssl:build:WOLFSSH_TERM` would make the guard real. (The embedded job's `WOLFSSH_SFTP` assertion is the strong one.)

### 3. Scope: the installed apps are not covered

An `--enable-all` install stages `bin/wolfssh` and `bin/wolfsshd`, but `SBOM_ARTIFACT = lib` means the only components are `wolfssh` (`hash-source: lib`) and `wolfssl`. `sbom.am` is lib-or-bin, so covering both needs a canonical change. Fine to defer, but given the CRA framing it is worth stating in the README rather than leaving implied.

### 4. Minor (both gen-sbom side, not this PR)

The main component is `GPL-3.0-or-later` while the wolfssl dependency component is hardcoded `GPL-3.0-only`, even though both projects' source headers carry the same "or (at your option) any later version" text. Also, properties on a wolfSSH SBOM are namespaced `wolfssl:build:*`.

## Everything else verified green

- Reproducible across two runs (byte-identical); CDX timestamp matches `git log -1 --format=%ct`
- `--enable-all` captures all 12 `WOLFSSH_*` feature macros (63 properties vs 50 plain), and the library hash changes with configuration
- `CPPFLAGS=-DDEFAULT_WINDOW_SZ=16384` at configure time lands in the SBOM
- VPATH build, and `--disable-shared --enable-static` correctly hashes `libwolfssh.a`
- `install-sbom` installs only under `share/doc/wolfssh`; `uninstall-hook` chain removes them from a plain `make uninstall` (0 files left)
- `make clean` removes the outputs (`CLEANFILES` correct)
- `autogen.sh` with `WARNINGS=all,error` exits 0
- `make dist` ships `scripts/sbom.am`; full tarball unpack -> configure -> make -> `make sbom` exits 0
- `pyspdxtools` validates the SPDX JSON; tag-value output carries `DESCRIBES` and `DEPENDS_ON`
- The `GPL-3.0-or-later` override is correct: `src/ssh.c` says "version 3 ... or (at your option) any later version", while `LICENSING` says only "GPLv3", which `detect_license` maps to `GPL-3.0-only`

Net: the mechanics hold up well outside the CI happy path. Findings 1 and 2 above, plus re-syncing `scripts/sbom.am` upstream, are what I would want before merge.

ejohnstown (approved):

PR #1051, opened 2026-06-23
tests: exercise DoKexDhReply host-key signature verify with corrupted signatures

yosuke-wolfssl:
### Summary

`DoKexDhReply` (client side) authenticates the server by verifying its
signature over the exchange hash `ssh->h` with the negotiated host key — RSA
(`wc_SignatureVerify`), ECC (`wc_SignatureVerify`), and Ed25519
(`wc_ed25519_verify_msg`). None of these verify calls was reached by any
negative test:

- `RewriteKexDhReplySignatureName` rewrote only the signature **name** string,
  which is rejected earlier at the name-match check (`WS_PARSE_E`).
- `TestKexDhReplyRejectsNoPublicKeyCheck` / `…WhenCallbackRejects` fail at the
  public-key-check callback, also before the crypto verify.

As a result, mutation testing showed the verify could be deleted without any
test failing — i.e. an attacker who knows only the server's public key could
impersonate the server / MITM the connection, and CI would stay green.

This PR adds negative regression tests that corrupt the signature **data**
(not the name) so the client reaches the actual cryptographic verify and must
reject the handshake. **Test-only change — no production code is modified;**
the verify logic in `src/internal.c` was already correct.

Addressed by f_6272.

### What changed (`tests/regress.c`)

- Added a second mode to the existing duplex KEXDH_REPLY mutator
  (`REGRESS_MUTATE_SIG_DATA`) that keeps the signature name and flips one byte
  of the signature data. The name-rewrite and data-corrupt paths now share a
  single parametrized helper pair (`RewriteSingleKexDhReplyPacket` /
  `RewriteKexDhReplyPacket`) instead of duplicating the packet parse/locate
  logic.
- Parametrized the test harness with a host-key path so ECC and Ed25519 keys
  can be loaded (it previously hardcoded the RSA key).
- New tests, each paired with a positive control and guarded by the relevant
  feature macros:
  - `TestKexDhReplyRejectsRsaSha2_256CorruptSig` / `…512` → `WS_RSA_E`
  - `TestKexDhReplyRejectsEccCorruptSig` (nistp256) → `WS_ECC_E`
  - `TestKexDhReplyRejectsEd25519CorruptSig` → `WS_ED25519_E`

### Testing

- `tests/regress.test` passes.
- Confirmed each new test fails when its verify is temporarily neutered
  (deleted the RSA/ECC `wc_SignatureVerify` assignment; bypassed the Ed25519
  branch) — the negative controls prove the tests have teeth.

### DIFF
--- a/tests/regress.c
+++ b/tests/regress.c
@@ -340,8 +340,12 @@ static void FreeChannelOpenHarness(ChannelOpenHarness* harness)
         wolfSSH_CTX_free(harness->ctx);
 }
 
+/* Needs server, client, key files, and one covered host-key algorithm. */
 #if !defined(NO_WOLFSSH_SERVER) && !defined(NO_WOLFSSH_CLIENT) && \
-    !defined(WOLFSSH_NO_RSA) && !defined(NO_FILESYSTEM)
+    !defined(NO_FILESYSTEM) && \
+    (!defined(WOLFSSH_NO_RSA_SHA2_256) || \
+     !defined(WOLFSSH_NO_RSA_SHA2_512) || \
+     !defined(WOLFSSH_NO_ECDSA_SHA2_NISTP256) || !defined(WOLFSSH_NO_ED25519))
     #if !defined(WOLFSSH_NO_DH_GROUP14_SHA256)
         #define KEXDH_REPLY_REGRESS_KEX_ALGO "diffie-hellman-group14-sha256"
     #elif !defined(WOLFSSH_NO_DH_GROUP16_SHA512)
@@ -395,11 +399,32 @@ static word32 LoadFileBuffer(const char* path, byte* buf, word32 bufSz)
 #define REGRESS_DUPLEX_QUEUE_SZ 32768U
 #define REGRESS_MUTATION_SCRATCH_SZ 4096U
 #define REGRESS_SERVER_KEY_PATH "keys/server-key-rsa.der"
+#define REGRESS_SERVER_KEY_ECC_PATH "keys/server-key-ecc.der"
+#define REGRESS_SERVER_KEY_ED25519_PATH "keys/server-key-ed25519.der"
 #define REGRESS_USERNAME "jill"
 #define REGRESS_PASSWORD "upthehill"
 #define REGRESS_MAX_HANDSHAKE_STEPS 2048
 #define REGRESS_SSH_PROTO_PREFIX "SSH-"
 #define REGRESS_SSH_PROTO_PREFIX_SZ 4U
+
+/* Host key used by the tests that are not tied to one key algorithm. */
+#if !defined(WOLFSSH_NO_RSA_SHA2_256)
+    #define REGRESS_DEFAULT_KEY_ALGO "rsa-sha2-256"
+    #define REGRESS_DEFAULT_KEY_PATH REGRESS_SERVER_KEY_PATH
+#elif !defined(WOLFSSH_NO_RSA_SHA2_512)
+    #define REGRESS_DEFAULT_KEY_ALGO "rsa-sha2-512"
+    #define REGRESS_DEFAULT_KEY_PATH REGRESS_SERVER_KEY_PATH
+#elif !defined(WOLFSSH_NO_ECDSA_SHA2_NISTP256)
+    #define REGRESS_DEFAULT_KEY_ALGO "ecdsa-sha2-nistp256"
+    #define REGRESS_DEFAULT_KEY_PATH REGRESS_SERVER_KEY_ECC_PATH
+#else
+    #define REGRESS_DEFAULT_KEY_ALGO "ssh-ed25519"
+    #define REGRESS_DEFAULT_KEY_PATH REGRESS_SERVER_KEY_ED25519_PATH
+#endif
+
+/* KEXDH_REPLY mutation modes for the duplex mutator. */
+#define REGRESS_MUTATE_SIG_NAME 0
+#define REGRESS_MUTATE_SIG_DATA 1
 
 typedef struct {
     byte data[REGRESS_DUPLEX_QUEUE_SZ];
@@ -407,12 +432,14 @@ typedef struct {
 } DuplexQueue;
 
 typedef struct {
+    const char* replaceName;
     byte enabled;
     int parseError;
     word32 matchedPackets;
     word32 mutatedPackets;
     byte scratch[REGRESS_MUTATION_SCRATCH_SZ];
     word32 scratchSz;
+    byte mode;
 } KexReplyMutator;
 
 typedef struct DuplexEndpoint {
@@ -549,8 +576,10 @@ static int QueueAppend(DuplexQueue* queue, const byte* data, word32 dataSz)
     return WS_SUCCESS;
 }
 
+/* SIG_NAME replaces the signature name; SIG_DATA flips a signature byte. */
 static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
-        const char* replacement, byte* out, word32 outSz, word32* outLen)
+        byte mode, const char* replacement, byte* out, word32 outSz,
+        word32* outLen)
 {
     const byte* payload;
     const byte* pubKey;
@@ -568,7 +597,10 @@ static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
     byte payloadBuf[REGRESS_MUTATION_SCRATCH_SZ];
     byte innerSig[REGRESS_MUTATION_SCRATCH_SZ];
 
-    if (replacement == NULL || out == NULL || outLen == NULL) {
+    if (out == NULL || outLen == NULL) {
+        return WS_BAD_ARGUMENT;
+    }
+    if (mode == REGRESS_MUTATE_SIG_NAME && replacement == NULL) {
         return WS_BAD_ARGUMENT;
     }
 
@@ -618,14 +650,38 @@ static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
         return WS_PARSE_E;
     }
 
-    (void)sigName;
-    (void)sigNameSz;
-
     innerSigSz = 0;
-    innerSigSz = AppendString(innerSig, sizeof(innerSig), innerSigSz,
-            replacement);
-    innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
-            sigData, sigDataSz);
+    if (mode == REGRESS_MUTATE_SIG_DATA) {
+        word32 dataStart;
+        word32 flipIdx;
+
+        if (sigDataSz <= LENGTH_SZ) {
+            return WS_PARSE_E;
+        }
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigName, sigNameSz);
+        /* Signature data value bytes start after their length prefix. */
+        dataStart = innerSigSz + LENGTH_SZ;
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigData, sigDataSz);
+
+        if (sigNameSz == sizeof("ssh-ed25519") - 1
+                && WMEMCMP(sigName, "ssh-ed25519", sigNameSz) == 0) {
+            /* Raw R || S: flip a byte of S so the verify rejects it. */
+            flipIdx = dataStart + sigDataSz / 2;
+        }
+        else {
+            /* RSA raw signature or ECC mpint r: flip an interior data byte. */
+            flipIdx = dataStart + LENGTH_SZ;
+        }
+        innerSig[flipIdx] ^= 0xFF;
+    }
+    else {
+        innerSigSz = AppendString(innerSig, sizeof(innerSig), innerSigSz,
+                replacement);
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigData, sigDataSz);
+    }
 
     outerIdx = 0;
     outerIdx = AppendBlob(payloadBuf, sizeof(payloadBuf), outerIdx,
@@ -638,12 +694,13 @@ static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
     return 1;
 }
 
-static int RewriteKexDhReplySignatureName(const byte* packet, word32 packetSz,
-        const char* replacement, byte* out, word32 outSz, word32* outLen)
+static int RewriteKexDhReplyPacket(const byte* packet, word32 packetSz,
+        byte mode, const char* replacement, byte* out, word32 outSz,
+        word32* outLen)
 {
     word32 offset = 0;
 
-    if (packet == NULL || replacement == NULL || out == NULL || outLen == NULL) {
+    if (packet == NULL || out == NULL || outLen == NULL) {
         return WS_BAD_ARGUMENT;
     }
 
@@ -657,7 +714,7 @@ static int RewriteKexDhReplySignatureName(const byte* packet, word32 packetSz,
 
         if (packet[offset + UINT32_SZ + PAD_LENGTH_SZ] == MSGID_KEXDH_REPLY) {
             rewriteRet = RewriteSingleKexDhReplyPacket(packet + offset,
-                    curPacketSz, replacement, out, outSz, outLen);
+                    curPacketSz, mode, replacement, out, outSz, outLen);
             if (rewriteRet <= 0) {
                 return rewriteRet;
             }
@@ -732,7 +789,8 @@ static int DuplexSend(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
         word32 mutatedSz = 0;
         int mutateRet;
 
-        mutateRet = RewriteKexDhReplySignatureName(output, outputSz, "ssh-rsa",
+        mutateRet = RewriteKexDhReplyPacket(output, outputSz,
+                endpoint->mutator->mode, endpoint->mutator->replaceName,
                 endpoint->mutator->scratch,
                 (word32)sizeof(endpoint->mutator->scratch), &mutatedSz);
         if (mutateRet < 0) {
@@ -785,7 +843,8 @@ static void FreeKexReplyHarness(KexReplyHarness* harness)
 }
 
 static void InitKexReplyHarnessEx(KexReplyHarness* harness,
-        const char* keyAlgo, byte mutateReply, byte skipPublicKeyCheck)
+        const char* keyAlgo, const char* keyPath, byte mutateReply,
+        byte mutateMode, const char* replaceName, byte skipPublicKeyCheck)
 {
     byte keyBuf[2048];
     word32 keySz;
@@ -794,6 +853,8 @@ static void InitKexReplyHarnessEx(KexReplyHarness* harness,
 
     InitDuplexPair(&harness->clientIo, &harness->serverIo, &harness->mutator);
     harness->mutator.enabled = mutateReply;
+    harness->mutator.mode = mutateMode;
+    harness->mutator.replaceName = replaceName;
 
     harness->clientCtx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_CLIENT, NULL);
     AssertNotNull(harness->clientCtx);
@@ -820,7 +881,7 @@ static void InitKexReplyHarnessEx(KexReplyHarness* harness,
         wolfSSH_CTX_SetPublicKeyCheck(harness->clientCtx, AcceptAnyServerHostKey);
     }
 
-    keySz = LoadFileBuffer(REGRESS_SERVER_KEY_PATH, keyBuf, sizeof(keyBuf));
+    keySz = LoadFileBuffer(keyPath, keyBuf, sizeof(keyBuf));
     AssertTrue(keySz > 0);
     AssertIntEQ(wolfSSH_CTX_UsePrivateKey_buffer(harness->serverCtx, keyBuf,
             keySz, WOLFSSH_FORMAT_ASN1), WS_SUCCESS);
@@ -840,9 +901,11 @@ static void InitKexReplyHarnessEx(KexReplyHarness* harness,
 }
 
 static void InitKexReplyHarness(KexReplyHarness* harness,
-        const char* keyAlgo, byte mutateReply)
+        const char* keyAlgo, const char* keyPath, byte mutateReply,
+        const char* replaceName)
 {
-    InitKexReplyHarnessEx(harness, keyAlgo, mutateReply, 0);
+    InitKexReplyHarnessEx(harness, keyAlgo, keyPath,
+            mutateReply, REGRESS_MUTATE_SIG_NAME, replaceName, 0);
 }
 
 static int IsHandshakeRetryable(int err)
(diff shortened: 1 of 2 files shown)


tests/regress.c
```diff
@@ -608,14 +622,39 @@ static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
...
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigName, sigNameSz);
+        /* Signature data value bytes start after their length prefix. */
+        dataStart = innerSigSz + LENGTH_SZ;
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigData, sigDataSz);
+
+        if (sigNameSz == sizeof("ssh-ed25519") - 1
+                && WMEMCMP(sigName, "ssh-ed25519", sigNameSz) == 0) {
+            /* Raw R || S: flip a low byte of S so R stays a valid point and S
```
ejohnstown: This comment is incorrect.

```
 /* Raw R || S: flip a low byte of S. A tampered signature makes
   * wc_ed25519_verify_msg reject it -- on this wolfSSL via ret ==
   * SIG_VERIFY_E; on builds that return ret == 0 the res != 1 guard
   * catches it. Either way DoKexDhReply maps it to WS_ED25519_E. */
```

ejohnstown (changes requested):

tests/regress.c
yosuke-wolfssl: Thanks.
I fixed the comment

aidangarske (dismissed):
## 🐺 Skoll Code Review

**Findings:** 2 total — 2 posted, 2 skipped

### Posted findings
- [High] **Ed25519 key fixture is not linked for out-of-tree test runs** — `tests/regress.c:388,1101-1106`
- [Medium] **Non-RSA host-key tests remain hidden behind the RSA build gate** — `tests/regress.c:332-333,1090-1106`

<details><summary>Skipped findings</summary>

- [High] Ed25519 key fixture is not linked for out-of-tree test runs
- [Medium] Non-RSA host-key tests remain hidden behind the RSA build gate

</details>

*Review generated by Skoll.*

tests/regress.c
```diff
@@ -384,12 +384,18 @@ static word32 LoadFileBuffer(const char* path, byte* buf, word32 bufSz)
 #define REGRESS_DUPLEX_QUEUE_SZ 32768U
 #define REGRESS_MUTATION_SCRATCH_SZ 4096U
 #define REGRESS_SERVER_KEY_PATH "keys/server-key-rsa.der"
+#define REGRESS_SERVER_KEY_ECC_PATH "keys/server-key-ecc.der"
+#define REGRESS_SERVER_KEY_ED25519_PATH "keys/server-key-ed25519.der"
```
aidangarske, line 403: **🟠 [High] Ed25519 key fixture is not linked for out-of-tree test runs**
*🚫 BLOCK*  `bug`

The PR adds `REGRESS_SERVER_KEY_ED25519_PATH` and the Ed25519 corrupted-signature test loads `keys/server-key-ed25519.der` via `LoadFileBuffer()`. Existing RSA and ECC fixtures are listed in `configure.ac` `AC_CONFIG_LINKS`, but the Ed25519 DER fixture is not. In an out-of-tree autotools build, `tests/regress.test` runs from the build directory, so this relative path is missing and `AssertTrue(keySz > 0)` aborts before the new test exercises the signature verifier.

**Recommendation:** Update the test fixture linking or load the key via a source-directory-aware path before merging. Add the new fixture to the build-tree links, for example in `configure.ac`: keys/server-key-ed25519.der:keys/server-key-ed25519.der

tests/regress.c
```diff
@@ -384,12 +384,18 @@ static word32 LoadFileBuffer(const char* path, byte* buf, word32 bufSz)
 #define REGRESS_DUPLEX_QUEUE_SZ 32768U
```
aidangarske, line 399: **🟡 [Medium] Non-RSA host-key tests remain hidden behind the RSA build gate**
*💡 SUGGEST*  `test`

The PR adds ECC and Ed25519 host-key verification regressions, but the whole KEXDH reply harness is still compiled only when `!defined(WOLFSSH_NO_RSA)`. That gate made sense when the harness only covered RSA, but now ECC-only or Ed25519-only builds skip the new coverage even when the relevant algorithms are enabled.

**Recommendation:** Broaden the compile guard so the new non-RSA regression tests run in non-RSA configurations. Relax the outer harness guard to require server/client/filesystem plus an enabled KEX algorithm and at least one host-key algorithm covered by the guarded tests, such as RSA SHA2, ECDSA P-256, or Ed25519.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Ed25519 key fixture is not linked for out-of-tree test runs** — `tests/regress.c:388,1101-1106`
- [Medium] **Non-RSA host-key tests remain hidden behind the RSA build gate** — `tests/regress.c:332-333,1090-1106`

*Review generated by Skoll.*

tests/regress.c
yosuke-wolfssl: Confirmed and fixed

tests/regress.c
yosuke-wolfssl: Fixed

yosuke-wolfssl:
Hello @aidangarske , @ejohnstown ,
I fixed the issues you mentioned. Could you please review this again ?

aidangarske (commented):
over to @ejohnstown

tests/regress.c
```diff
@@ -728,7 +789,8 @@ static int DuplexSend(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
         word32 mutatedSz = 0;
         int mutateRet;
 
-        mutateRet = RewriteKexDhReplySignatureName(output, outputSz, "ssh-rsa",
+        mutateRet = RewriteKexDhReplyPacket(output, outputSz,
```
aidangarske, line 792: **🔵 [Low] Signature-name WMEMCMP branch still unexercised: replacement name "ssh-rsa" only ever trips the length check** · _Missing Tests_

Verified independently. The PR's stated purpose is closing mutation-testing gaps in `DoKexDhReply`. It closes the crypto-verify gap but leaves an adjacent one open, and the two new name-downgrade tests do not close it. `DuplexSend` hardcodes `"ssh-rsa"` (7 bytes) as the replacement signature name for every algorithm (regress.c:792-795; confirmed via git diff that this literal is introduced by this PR). `DoKexDhReply` checks the name in two steps at src/internal.c:7031-7040: first `scratch != expectedSigNameSz` (length), then `WMEMCMP(sig + begin, expectedSigName, scratch) != 0` (content), joined by `||`. Confirmed by reading src/internal.c:7032-7033 that the length comparison short-circuits. Because `"ssh-rsa"` is 7 bytes and every expected name is longer -- `rsa-sha2-256`/`rsa-sha2-512` = 12, `ecdsa-sha2-nistp256` = 19, `ssh-ed25519` = 11 -- the length check returns WS_PARSE_E on every name test and the `WMEMCMP` is never reached. The pre-existing RSA tests had this gap; the PR adds `TestKexDhReplyRejectsEccSigNameDowngrade` and `TestKexDhReplyRejectsEd25519SigNameDowngrade` (regress.c:1012-1028), which reuse the same hardcoded replacement and therefore extend the gap to two more algorithms rather than closing it. The `WMEMCMP` could be deleted and all four tests would stay green. Confirmed that the only other callers of the SIG_NAME mutation path (`AssertHandshakeRejectsWithNoPublicKeyCheck`, regress.c:1036) pass mutation-enabled = 0, so no test anywhere reaches the WMEMCMP. A same-length replacement would reach it -- e.g. `rsa-sha2-256` -> `rsa-sha2-512` (12 vs 12)…

**Fix:** Plumb the replacement name through the harness the same way `keyPath` and `mutateMode` were plumbed, and add at least one same-length substitution (e.g. `rsa-sha2-256` -> `rsa-sha2-512`) so the `WMEMCMP` at src/internal.c:7033 is actually reached and mutation-covered.

tests/regress.c
```diff
@@ -929,25 +996,45 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo)
...
 {
-    AssertHandshakeSucceeds("rsa-sha2-512");
-    AssertHandshakeRejectsMutatedReply("rsa-sha2-512");
+    AssertHandshakeSucceeds("rsa-sha2-512", REGRESS_SERVER_KEY_PATH);
+    AssertHandshakeRejectsMutatedReply("rsa-sha2-512", REGRESS_SERVER_KEY_PATH);
 }
 #endif
 
-static void AssertHandshakeRejectsWithNoPublicKeyCheck(const char* keyAlgo)
+#ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
```
aidangarske, line 1018: **🔵 [Low] New ECC/Ed25519 name-downgrade tests do not pin the rejection error code** · _Missing Tests_

Verified independently. The PR adds `AssertHandshakeRejectsCorruptedSig` (regress.c:1091-1112) which pins the exact error with `AssertIntEQ(result.clientErr, expectedErr)` at line 1109 -- that is what gives the new corrupt-sig tests their teeth. The two new name-downgrade tests added in the same PR route through `AssertHandshakeRejectsMutatedReply` (regress.c:976-994), which only asserts `clientRet == WS_FATAL_ERROR` (line 990) and that the error is not `WS_WANT_READ`/`WS_WANT_WRITE` (line 991). Any rejection reason passes. Confirmed via git diff that this PR modifies that helper's signature (adding the `keyPath` parameter) without taking the opportunity to add an `expectedErr` parameter, while two sibling helpers in the same file (`AssertHandshakeRejectsWithNoPublicKeyCheck` at 1043 and `AssertHandshakeRejectsWhenCallbackRejects` at 1072) already do pin their error with `AssertIntEQ(result.clientErr, WS_PUBKEY_REJECTED_E)`. If a future refactor made the client fail earlier for an unrelated reason (a parse bug returning `WS_BUFFER_E`, a key-load failure, an algo-negotiation change), these tests would still report green while no longer testing the name check at all. The paired `AssertHandshakeSucceeds` positive control does rule out the gross negotiation/key-load failures, which is why this is Low rather than Medium -- but the assertion is strictly weaker than the standard the PR itself establishes three functions below it, and weaker than two pre-existing helpers in the same file.

**Fix:** Add an `expectedErr` parameter to `AssertHandshakeRejectsMutatedReply` mirroring `AssertHandshakeRejectsCorruptedSig`, and assert `AssertIntEQ(result.clientErr, WS_PARSE_E)` for all four name-downgrade tests.

tests/regress.c
```diff
@@ -614,14 +648,40 @@ static int RewriteSingleKexDhReplyPacket(const byte* packet, word32 packetSz,
...
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigName, sigNameSz);
+        /* Signature data value bytes start after their length prefix. */
+        dataStart = innerSigSz + LENGTH_SZ;
+        innerSigSz = AppendBlob(innerSig, sizeof(innerSig), innerSigSz,
+                sigData, sigDataSz);
+
+        if (sigNameSz == sizeof("ssh-ed25519") - 1
+                && WMEMCMP(sigName, "ssh-ed25519", sigNameSz) == 0) {
+            /* Raw R || S: flip a byte of S. wc_ed25519_verify_msg rejects the
```
aidangarske: **⚪ [Info] Multi-line block comments violate repository comment policy** · _Logic_

Verified independently against the diff. The PR adds several multi-line function/block header comments; confirmed as newly added lines via git diff: regress.c:575-577 (3-line header on `RewriteSingleKexDhReplyPacket`), 666-670 (5-line Ed25519 return-shape explanation), 1088-1090 (3-line header on `AssertHandshakeRejectsCorruptedSig`), 1133-1134 (2-line header on `TestKexDhReplyRejectsEccCorruptSig`), plus a 2-line block at the top of the diff. Repo policy (CLAUDE.md / rules/comments.md) is one line max and explicitly bans multi-line function/helper/block header comments -- long rationale belongs in the commit message or PR description, which in this case already contains it. The most notable is the 5-line comment at 666-670 explaining wolfCrypt's two possible Ed25519 return shapes; that reasoning is correct but belongs in the PR body, not inline.

**Fix:** Collapse each to a single-line WHY comment, or drop entirely where the code is self-evident. The detailed wolfCrypt return-shape analysis is already in the PR description.

tests/regress.c
```diff
@@ -989,13 +1077,77 @@ static void AssertHandshakeRejectsWhenCallbackRejects(const char* keyAlgo)
...
+    AssertTrue(result.clientErr != WS_WANT_READ &&
+            result.clientErr != WS_WANT_WRITE);
+    AssertIntEQ(result.clientErr, expectedErr);
+
+    FreeKexReplyHarness(&harness);
+}
+
 #ifndef WOLFSSH_NO_RSA_SHA2_256
-    AssertHandshakeRejectsWhenCallbackRejects("rsa-sha2-256");
+static void TestKexDhReplyRejectsRsaSha2_256CorruptSig(void)
```
aidangarske, line 1120: **⚪ [Info] Positive-control handshake runs twice per algorithm, doubling test runtime** · _Logic_

Verified independently. Each new `*CorruptSig` test opens with `AssertHandshakeSucceeds(algo, path)` (regress.c:1117, 1126, 1137, 1146), but the corresponding `*SigNameDowngrade` test already ran the identical positive control with the identical algo and key path (regress.c:999, 1007, 1015, 1024). Confirmed that both sets are actually invoked in `main()` -- the four SigNameDowngrade tests at regress.c:4845-4854 and the four CorruptSig tests at regress.c:4859-4868 -- so the duplication is real at runtime, not gated away. That is four extra full DH-group14 handshakes (two of them RSA-2048), each with a complete key exchange plus userauth, for no additional signal. Purely a runtime cost -- no correctness impact.

**Fix:** Drop the duplicate `AssertHandshakeSucceeds` from the `*CorruptSig` tests, or merge each SigNameDowngrade/CorruptSig pair into one test per algorithm that runs the positive control once.

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Signature-name WMEMCMP branch still unexercised: replacement name "ssh-rsa" only ever trips the length check** — `tests/regress.c:792-793`
- [Low] **New ECC/Ed25519 name-downgrade tests do not pin the rejection error code** — `tests/regress.c:1012-1028`
- [Info] **Multi-line block comments violate repository comment policy** — `tests/regress.c:575-577,666-670,1088-1090,1133-1134`
- [Info] **Positive-control handshake runs twice per algorithm, doubling test runtime** — `tests/regress.c:1117,1126,1137,1146`

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Multi-line block comments contradict the repository comment policy** — `tests/regress.c:339-340,575-577,666-670,1088-1090,1133-1134`
- [Info] **New Ed25519 tests are gated behind a non-default wolfSSL build option** — `tests/regress.c:1021-1028,1143-1150`
- [Info] **Positive-control handshake is now run twice per host-key algorithm** — `tests/regress.c:997-1028,1114-1150`

tests/regress.c
```diff
@@ -545,8 +572,12 @@ static int QueueAppend(DuplexQueue* queue, const byte* data, word32 dataSz)
     return WS_SUCCESS;
 }
 
+/* Rebuild a single KEXDH_REPLY packet: REGRESS_MUTATE_SIG_NAME replaces the
```
aidangarske: **🔵 [Low] Multi-line block comments contradict the repository comment policy** · _Quality_

Verified against the diff: `git diff HEAD~1 -- tests/regress.c` confirms every flagged block is added by this PR, not pre-existing. The PR introduces multi-line block comments including multi-line function-header comments on `RewriteSingleKexDhReplyPacket` (575-577) and `AssertHandshakeRejectsCorruptedSig` (1088-1090), a 5-line explanatory block inside the Ed25519 branch (666-670), a 2-line ECC mpint-length header on `TestKexDhReplyRejectsEccCorruptSig` (1133-1134), and a 2-line gate rationale at 339-340. The project comment policy is one-line max, reserved for a WHY that would surprise a reader, and explicitly forbids multi-line function/helper header blocks — extended rationale belongs in the commit message or PR description (where it already is). The single-line comment at line 661 (`/* Signature data value bytes start after their length prefix. */`) is the correct pattern, as are the added one-liners at the equivalents of diff lines 32, 47, 109 and 122. Style-only; the code itself is correct.

**Fix:** Collapse each block to a single line or drop it — e.g. `/* mode: SIG_NAME rewrites the name; SIG_DATA flips a signature byte. */`. The Ed25519 error-mapping rationale (666-670) and the ECC mpint-length rationale (1133-1134) are already captured in the PR description and can be removed from the source.

tests/regress.c
```diff
@@ -929,25 +996,45 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo)
 #ifndef WOLFSSH_NO_RSA_SHA2_256
 static void TestKexDhReplyRejectsRsaSha2_256SigNameDowngrade(void)
```
aidangarske, line 1000: **⚪ [Info] Positive-control handshake is now run twice per host-key algorithm** · _Quality_

Verified by reading both test blocks. Each new `...CorruptSig` test opens with `AssertHandshakeSucceeds(algo, path)`, and the corresponding pre-existing `...SigNameDowngrade` test already calls the identical `AssertHandshakeSucceeds(algo, path)` with the same arguments. Confirmed for all four algorithms (rsa-sha2-256, rsa-sha2-512, ecdsa-sha2-nistp256, ssh-ed25519). For every algorithm this executes a full unmutated handshake — a complete DH KEX plus signature verify — twice with identical arguments and identical assertions. Purely a runtime cost in the test suite; no correctness impact, and the positive control is genuinely valuable as a negative-control anchor for the mutation tests.

**Fix:** Optional: drop the duplicate `AssertHandshakeSucceeds` from the `...CorruptSig` tests, or fold the name-downgrade and corrupt-sig assertions into a single per-algorithm test that runs the positive control once.

tests/regress.c
```diff
@@ -929,25 +996,45 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo)
...
+#ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
+static void TestKexDhReplyRejectsEccSigNameDowngrade(void)
+{
+    AssertHandshakeSucceeds("ecdsa-sha2-nistp256", REGRESS_SERVER_KEY_ECC_PATH);
+    AssertHandshakeRejectsMutatedReply("ecdsa-sha2-nistp256",
+            REGRESS_SERVER_KEY_ECC_PATH);
+}
+#endif
+
+#ifndef WOLFSSH_NO_ED25519
```
aidangarske, line 1027: **⚪ [Info] New Ed25519 tests are gated behind a non-default wolfSSL build option** · _Missing Tests_

Verified independently. `wolfssh/internal.h:123-130` defines `WOLFSSH_NO_ED25519` unless wolfSSL supplies ALL of `HAVE_ED25519`, `WOLFSSL_ED25519_STREAMING_VERIFY`, `HAVE_ED25519_KEY_IMPORT`, and `HAVE_ED25519_KEY_EXPORT`. The wolfSSL install on this machine (`/usr/local/include/wolfssl/options.h`) has `HAVE_ED25519` but zero occurrences of `WOLFSSL_ED25519_STREAMING_VERIFY`, so `WOLFSSH_NO_ED25519` is defined and both new Ed25519 tests — each wrapped in `#ifndef WOLFSSH_NO_ED25519` — compile out. No workflow under `.github/workflows/` references `WOLFSSL_ED25519_STREAMING_VERIFY`. Consequently the `wc_ed25519_verify_msg` call in `DoKexDhReply` (src/internal.c:7122) may remain unreached by any negative test in CI — the specific mutation gap the PR sets out to close could stay open on the Ed25519 path. This is a pre-existing repository gate, not something the PR introduces, and the PR follows the exact convention already used by `tests/kex.c:309` and `tests/unit.c:2786`. The RSA tests are unconditional in practice and do close the RSA verify gap. Noted only so the coverage claim is calibrated.

**Fix:** Confirm at least one CI job builds wolfSSL with `WOLFSSL_ED25519_STREAMING_VERIFY` (and ECC P-256) so the new Ed25519 and ECC negative tests actually execute. If none does, consider adding a matrix entry, otherwise the mutation-coverage improvement is limited to the RSA paths in practice.

---
> **Note:** Referenced line (`tests/regress.c:1021-1028,1143-1150`) is outside the diff hunk. Comment anchored to nearest changed region.

aidangarske, line 1027: can you confirm at least one CI job builds wolfSSL with `WOLFSSL_ED25519_STREAMING_VERIFY` (and ECC P-256) so the new Ed25519 and ECC negative tests actually execute

tests/regress.c
```diff
@@ -929,25 +996,45 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo)
 #ifndef WOLFSSH_NO_RSA_SHA2_256
 static void TestKexDhReplyRejectsRsaSha2_256SigNameDowngrade(void)
 {
-    AssertHandshakeSucceeds("rsa-sha2-256");
-    AssertHandshakeRejectsMutatedReply("rsa-sha2-256");
+    AssertHandshakeSucceeds("rsa-sha2-256", REGRESS_SERVER_KEY_PATH);
```
aidangarske, line 1002: Drop the duplicate `AssertHandshakeSucceeds` from the `...CorruptSig` tests, or fold the name-downgrade and corrupt-sig assertions into a single per-algorithm test that runs the positive control once?

tests/regress.c
yosuke-wolfssl: Fixed. The replacement signature name is now plumbed through the harness (new replaceName field on KexReplyMutator, passed via InitKexReplyHarnessEx / AssertHandshakeRejectsMutatedReply) instead of the hardcoded "ssh-rsa". The name-downgrade tests now use same-length wrong names so the WMEMCMP at internal.c:7033 is actually reached: rsa-sha2-256 -> rsa-sha2-512, rsa-sha2-512 -> rsa-sha2-256, and ecdsa-sha2-nistp256 -> ecdsa-sha2-nistp384. I verified with a negative control: deleting the WMEMCMP (leaving only the length check) now fails the RSA-256 downgrade test, which it did not before this change.

tests/regress.c
yosuke-wolfssl: Fixed. AssertHandshakeRejectsMutatedReply now takes an expectedErr parameter and asserts AssertIntEQ(result.clientErr, expectedErr), mirroring AssertHandshakeRejectsCorruptedSig. All four name-downgrade tests now pin WS_PARSE_E.

tests/regress.c
yosuke-wolfssl: Confirmed. wolfSSL's --enable-all turns these on: in configure.ac, enable_ed25519_stream=yes is defaulted inside the ENABLED_ALL_CRYPT block (alongside enable_ed25519, ECC, etc.), so --enable-all defines HAVE_ED25519, WOLFSSL_ED25519_STREAMING_VERIFY, the ed25519 key import/export, and ECC P-256. That means WOLFSSH_NO_ED25519 is not defined and both the Ed25519 and ECC negative tests compile and run. The CI jobs that build wolfSSL with --enable-all and run the wolfSSH test suite (os-check.yml, sanitizer.yml, plus the sshd/scp/sftp test workflows) exercise these paths, so the Ed25519 and ECC mutation coverage is active in CI, not just the RSA paths.

tests/regress.c
yosuke-wolfssl: Done. Dropped the duplicate AssertHandshakeSucceeds from the four *CorruptSig tests.

yosuke-wolfssl:
Hello @aidangarske ,

aidangarske (approved):

ejohnstown (approved):

PR #1054, opened 2026-06-24
Extended (stderr) data: buffering and flow control

ejohnstown:
- Ignore unknown extended data types: consume, log, and window-credit them per RFC 4254/4251 instead of rejecting with WS_INVALID_EXTDATA.
- Per-channel buffering: move extDataBuffer to WOLFSSH_CHANNEL; accumulate by appending so unread data survives across packets.
- Window flow control (RFC 4254 5.2): charge on receipt, credit on drain, reject data larger than the advertised window.
- Rekey-safe crediting (RFC 4253 7.1): ChannelCreditWindow() parks credit that can't go out mid-KEX or after a failed send; flushed when keying completes. Reads are not blocked during rekey.
- Multi-channel: stream_/extended_data_ read serves only the first channel; other channels use new API wolfSSH_Channel{,Id}{Read,Send}Ext.
- SCP client: drain stderr in a loop, not one chunk.
- Document the mandatory-drain contract in the public header.

Issues: F-864, F-2078

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -11388,9 +11388,6 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
     if (ret == WS_SUCCESS)
         ret = GetUint32(&dataTypeCode, buf, len, &begin);
     if (ret == WS_SUCCESS)
-        ret = (dataTypeCode == CHANNEL_EXTENDED_DATA_STDERR) ?
-            WS_SUCCESS : WS_INVALID_EXTDATA;
-    if (ret == WS_SUCCESS)
         ret = GetSize(&dataSz, buf, len, &begin);
 
     if (ret == WS_SUCCESS) {
@@ -11399,21 +11396,26 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
             ret = WS_INVALID_CHANID;
         else if (dataSz > channel->maxPacketSz)
             ret = WS_RECV_OVERFLOW_E;
-        else {
-            ret = PutBuffer(&ssh->extDataBuffer,  buf + begin, dataSz);
+        else if (dataTypeCode == CHANNEL_EXTENDED_DATA_STDERR) {
+            ret = PutBuffer(&ssh->extDataBuffer, buf + begin, dataSz);
             #ifdef DEBUG_WOLFSSH
             DumpOctetString(buf + begin, dataSz);
             #endif
-            if (ret == WS_SUCCESS) {
+            if (ret == WS_SUCCESS)
                 ret = SendChannelWindowAdjust(ssh, channel->channel, dataSz);
+            if (ret == WS_SUCCESS) {
+                ssh->lastRxId = channelId;
+                ret = WS_EXTDATA;
             }
         }
+        else {
+            WLOG(WS_LOG_INFO, "Ignoring unknown extended data type %u",
+                    dataTypeCode);
+            ret = SendChannelWindowAdjust(ssh, channel->channel, dataSz);
+            if (ret == WS_SUCCESS)
+                ssh->lastRxId = channelId;
+        }
         *idx = begin + dataSz;
-    }
-
-    if (ret == WS_SUCCESS) {
-        ssh->lastRxId = channelId;
-        ret = WS_EXTDATA;
     }
 
     WLOG(WS_LOG_DEBUG, "Leaving DoChannelExtendedData(), ret = %d", ret);
--- a/src/internal.c
+++ b/src/internal.c
@@ -1536,8 +1536,7 @@ WOLFSSH* SshInit(WOLFSSH* ssh, WOLFSSH_CTX* ctx)
     ssh->keyingCompletionCtx = (void*)ssh;
 
     if (BufferInit(&ssh->inputBuffer, 0, ctx->heap) != WS_SUCCESS  ||
-        BufferInit(&ssh->outputBuffer, 0, ctx->heap) != WS_SUCCESS ||
-        BufferInit(&ssh->extDataBuffer, 0, ctx->heap) != WS_SUCCESS) {
+        BufferInit(&ssh->outputBuffer, 0, ctx->heap) != WS_SUCCESS) {
 
         wolfSSH_free(ssh);
         ssh = NULL;
@@ -1556,7 +1555,6 @@ void SshResourceFree(WOLFSSH* ssh, void* heap)
 
     ShrinkBuffer(&ssh->inputBuffer, 1);
     ShrinkBuffer(&ssh->outputBuffer, 1);
-    ShrinkBuffer(&ssh->extDataBuffer, 1);
     WS_FORCEZERO(ssh->k, ssh->kSz);
     HandshakeInfoFree(ssh->handshake, heap);
     WS_FORCEZERO(&ssh->keys, sizeof(Keys));
@@ -3618,6 +3616,14 @@ WOLFSSH_CHANNEL* ChannelNew(WOLFSSH* ssh, byte channelType,
                 newChannel->inputBuffer.buffer = buffer;
                 newChannel->inputBuffer.bufferSz = initialWindowSz;
                 newChannel->inputBuffer.dynamicFlag = 1;
+                if (BufferInit(&newChannel->extDataBuffer, 0, heap)
+                        != WS_SUCCESS) {
+                    WLOG(WS_LOG_DEBUG,
+                         "Unable to init new channel's ext data buffer");
+                    WFREE(buffer, heap, DYNTYPE_BUFFER);
+                    WFREE(newChannel, heap, DYNTYPE_CHANNEL);
+                    newChannel = NULL;
+                }
             }
             else {
                 WLOG(WS_LOG_DEBUG, "Unable to allocate new channel's buffer");
@@ -3654,6 +3660,14 @@ void ChannelDelete(WOLFSSH_CHANNEL* channel, void* heap)
         }
         WFREE(channel->inputBuffer.buffer,
               channel->inputBuffer.heap, DYNTYPE_BUFFER);
+        if (channel->extDataBuffer.length > channel->extDataBuffer.idx) {
+            /* The app never drained it and the channel is going away. */
+            WLOG(WS_LOG_INFO,
+                 "Discarding %u bytes of unread extended data on channel %u",
+                 channel->extDataBuffer.length - channel->extDataBuffer.idx,
+                 channel->channel);
+        }
+        ShrinkBuffer(&channel->extDataBuffer, 1);
         if (channel->command)
             WFREE(channel->command, heap, DYNTYPE_STRING);
         WFREE(channel, heap, DYNTYPE_CHANNEL);
@@ -7301,6 +7315,92 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 }
 
 
+/* Returns amount bytes of receive-window credit to the peer, folding in credit
+ * already parked on the channel. Credit that cannot reach the transport is
+ * parked, not dropped: no WINDOW_ADJUST may be sent mid-rekey (RFC 4253 section
+ * 7.1), an unbundled packet queued nothing, and a socket error can discard what
+ * was bundled. Credit that reached the output buffer counts as delivered. */
+int ChannelCreditWindow(WOLFSSH* ssh, WOLFSSH_CHANNEL* channel, word32 amount)
+{
+    word32 total;
+    int ret;
+    byte bundled = 0;
+
+    if (ssh == NULL || channel == NULL)
+        return WS_BAD_ARGUMENT;
+
+    if (amount > UINT32_MAX - channel->pendingWindowAdjust) {
+        WLOG(WS_LOG_ERROR, "pending window adjust would overflow");
+        return WS_OVERFLOW_E;
+    }
+
+    total = channel->pendingWindowAdjust + amount;
+    if (total == 0)
+        return WS_SUCCESS;
+
+    if (ssh->isKeying) {
+        channel->pendingWindowAdjust = total;
+        return WS_SUCCESS;
+    }
+
+    channel->pendingWindowAdjust = 0;
+    ret = SendChannelWindowAdjust(ssh, channel->channel, total, &bundled);
+
+    /* Keep the credit owed only if it never reached the transport. Once bundled
+     * it sits in the output buffer for a later flush, so re-parking after a
+     * short write (WS_WANT_WRITE) would send it twice and inflate the peer's
+     * window. WS_SOCKET_ERROR_E is the exception: wolfSSH_SendPacket() can
+     * discard what it bundled, so it stays owed. */
+    if (!bundled || ret == WS_SOCKET_ERROR_E) {
+        channel->pendingWindowAdjust = total;
+        WLOG(WS_LOG_ERROR,
+             "ChannelCreditWindow: window adjust send failed (%d) for "
+             "channel %u; %u bytes of credit parked for retry",
+             ret, channel->channel, total);
+    }
+
+    return ret;
+}
+
+
+/* Flush receive-window credit parked on the channels. Called when keying
+ * completes. */
+static int SendPendingChannelWindowAdjust(WOLFSSH* ssh)
+{
+    WOLFSSH_CHANNEL* cur;
+    int savedError;
+    int ret = WS_SUCCESS;
+
+    if (ssh == NULL)
+        return WS_BAD_ARGUMENT;
+
+    /* Still mid-KEX: leave the credit parked until both sides finish. */
+    if (ssh->isKeying)
+        return WS_SUCCESS;
+
+    /* ssh->error is restored only when every channel flushed cleanly, so a
+     * back-pressured or broken transport is not hidden by this incidental
+     * flush. Keep going after a failure: the other channels are independent. */
+    savedError = ssh->error;
+
+    for (cur = ssh->channelList; cur != NULL; cur = cur->next) {
+        if (cur->pendingWindowAdjust != 0) {
+            int adjustResult = ChannelCreditWindow(ssh, cur, 0);
+
+            if (adjustResult != WS_SUCCESS && ret == WS_SUCCESS)
+                ret = adjustResult;
+        }
+    }
+
+    if (ret == WS_SUCCESS)
+        ssh->error = savedError;
+    else
+        ssh->error = ret;
+
+    return ret;
+}
+
+
 static int DoNewKeys(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 {
     int ret = WS_SUCCESS;
@@ -7343,6 +7443,9 @@ static int DoNewKeys(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ssh->isKeying &= ~WOLFSSH_PEER_IS_KEYING;
         HandshakeInfoFree(ssh->handshake, ssh->ctx->heap);
         ssh->handshake = NULL;
+        /* If this was the last keying flag to clear, flush any window credit
+         * that was deferred during the rekey (RFC 4253 section 7.1). */
+        ret = SendPendingChannelWindowAdjust(ssh);
         WLOG(WS_LOG_DEBUG, "Keying completed");
         if (ssh->ctx->keyingCompletionCb)
             ssh->ctx->keyingCompletionCb(ssh->keyingCompletionCtx);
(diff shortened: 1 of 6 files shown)


src/internal.c
```diff
@@ -10214,23 +10211,66 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
...
+        else {
+            /* Unknown extended data type. RFC 4254/4251 call for unknown
+             * values to be ignored rather than rejected, so consume and
+             * discard the payload. Nothing is buffered for the application,
+             * so replenish the window immediately. */
+            WLOG(WS_LOG_INFO, "Ignoring unknown extended data type %u",
+                    dataTypeCode);
+            ret = SendChannelWindowAdjust(ssh, channel->channel, dataSz);
+            if (ret == WS_SUCCESS)
+                ssh->lastRxId = channelId;
```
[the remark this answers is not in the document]
ejohnstown: Both new paths already gate on ssh->isKeying and defer the credit via channel->pendingWindowAdjust, flushed after NEWKEYS -- same as wolfSSH_extended_data_read(). No WINDOW_ADJUST is sent mid-KEX. See DoChannelExtendedData().

wolfssh/internal.h
```diff
@@ -874,7 +874,22 @@ struct WOLFSSH {
...
+                                   * extended data was charged against; its
+                                   * window is replenished as the app reads.
+                                   * If a second channel buffers stderr before
+                                   * the app drains the buffer, this is
+                                   * overwritten: the first channel's window is
+                                   * never credited, and on read this channel is
+                                   * over-credited by the first channel's bytes
+                                   * (the whole shared buffer is credited here).
+                                   * Buffered stderr is intended for the
+                                   * single-session case. */
```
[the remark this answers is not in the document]
ejohnstown: Already updated. The extDataChannelId comment now documents the fail-safe: a second channel's stderr is dropped and credited back rather than overwriting the id, no over-credit.

src/internal.c
```diff
@@ -10214,23 +10264,85 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
...
+                 * one channel's stderr buffers at a time. Buffering a second
+                 * would desync windows (displaced channel never credited, this
+                 * one over-credited). Fail safe: drop it and replenish its
+                 * window (now, or deferred past a rekey) so nothing leaks. */
+                WLOG(WS_LOG_ERROR,
+                     "Extended data from channel %u dropped; channel %u still "
+                     "has unread stderr buffered (only one channel's stderr "
+                     "can be buffered at a time)",
+                     channel->channel, ssh->extDataChannelId);
+                if (ssh->isKeying) {
```
aidangarske: **🟠 [High] Deferred discarded extended data is not charged against the channel receive window during rekey**
*🚫 BLOCK*

The new rekey deferral path accepts discarded extended data while `ssh->isKeying` by adding peer-controlled `dataSz` to `channel->pendingWindowAdjust`, but it leaves `channel->windowSz` unchanged. The only validation is per-packet (`dataSz <= channel->maxPacketSz` and `dataSz <= channel->windowSz`), so it never accounts for cumulative bytes received while window adjustments are deferred. A malicious authenticated peer can send repeated unknown extended-data packets, or repeated second-channel stderr packets that are dropped, each individually within the unchanged window, and this code accepts more bytes than were advertised. When keying completes, `SendPendingChannelWindowAdjust()` flushes the accumulated credit, potentially over-crediting the peer and wrapping the unchecked `word32` `pendingWindowAdjust += dataSz`. The buffered stderr path correctly subtracts from `windowSz`, but the discarded-data rekey branches introduced by this PR do not. Data flow: encrypted peer packet -> DoPacket() -> DoChannelExtendedData() -> GetSize() -> per-packet dataSz checks -> uncharged pendingWindowAdjust accumulation. Impact is connection-level resource / flow-control DoS, not memory corruption. Severity views differ across modes: the review mode rated this High/BLOCK (cumulative window bypass + credit wrap); the security mode rated it Low (DoS only). The stricter High assessment is retained.

**Recommendation:** Charge the receive window whenever payload bytes are accepted from the peer, even when discarded, and guard the deferred-credit addition against overflow, e.g. `if (dataSz > UINT32_MAX - channel->pendingWindowAdjust) ret = WS_OVERFLOW_E; else { channel->windowSz -= dataSz; channel->pendingWindowAdjust += dataSz; ret = WS_SUCCESS; }`. In `SendPendingChannelWindowAdjust()`, restore `windowSz` by `amount` (with overflow check) after `SendChannelWindowAdjust()` returns `WS_SUCCESS`/`WS_WANT_WRITE`. Extend `test_DoChannelExtendedData_keying()` to cover cumulative deferred data and `pendingWindowAdjust` overflow boundaries.

src/ssh.c
```diff
@@ -1425,6 +1438,48 @@ int wolfSSH_extended_data_read(WOLFSSH* ssh, byte* out, word32 outSz)
...
+             * The bytes are already in out, so this is a completed read: it
+             * reports the byte count and must not disturb ssh->error. Stomping
+             * ssh->error here would be invisible to callers that only check the
+             * return value and surprising to callers that inspect
+             * wolfSSH_get_error() after a successful read. SendChannelWindow-
+             * Adjust -> wolfSSH_SendPacket sets ssh->error = WS_WANT_WRITE when
+             * the bundled adjust cannot be flushed in non-blocking mode, so
+             * save ssh->error across the send and restore it; a hard send
+             * failure is logged, not returned. */
+            adjustResult = SendChannelWindowAdjust(ssh, channel->channel,
```
aidangarske: **🔵 [Low] Extended-data read hides nonblocking WINDOW_ADJUST back-pressure**
*💡 SUGGEST*

The PR moves stderr receive-window replenishment from receipt time to `wolfSSH_extended_data_read()`, but then treats `WS_WANT_WRITE` from `SendChannelWindowAdjust()` as an invisible success. In nonblocking mode the adjust may be bundled in `outputBuffer` but not flushed; the function restores `ssh->error` and returns a byte count, so callers are not told to wait for writability or call a flushing path. Read-only loops then keep waiting for inbound data while the peer never receives the window credit it needs to send more, creating a remote-triggerable channel stall. Data flow: peer sends stderr -> DoChannelExtendedData() buffers it and decrements windowSz -> application calls wolfSSH_extended_data_read() -> SendChannelWindowAdjust() returns WS_WANT_WRITE -> error restored and byte count returned -> pending adjust may remain unsent. Impact is session/channel DoS.

**Recommendation:** Do not silently hide `WS_WANT_WRITE` for the window-adjust path. Mirror `wolfSSH_stream_read()` semantics: only advance `extDataBuffer.idx` and credit `channel->windowSz` after the adjust is queued/flushed successfully, and return `WS_WANT_WRITE` with `ssh->error` preserved when the caller must service the writable socket. Alternatively, flush pending output before blocking and preserve a visible writable-wakeup signal for callers.

aidangarske (dismissed):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Deferred discarded extended data is not charged against the channel receive window during rekey** — `src/internal.c:10296-10334`
- [Low] **Extended-data read hides nonblocking WINDOW_ADJUST back-pressure** — `src/ssh.c:1471-1475`

*Review generated by Skoll.*

src/internal.c
ejohnstown: Fixed the overflow guard on pendingWindowAdjust += dataSz (both discard branches) — same pattern as peerWindowSz in DoChannelWindowAdjust, returns WS_OVERFLOW_E. Disagree with the rest: windowSz-never-charged-for-discards is intentional/tested (test_DoChannelExtendedData_keying), and the overflow itself is harmless (no memory touch, just an undercounted adjust to an already-flooding peer). Retaining Low, not High/BLOCK.

src/ssh.c
ejohnstown: Addressed. The version you reviewed restored ssh->error unconditionally; it no longer does. On WS_WANT_WRITE from the adjust send, ssh->error is left as WS_WANT_WRITE so wolfSSH_get_error() tells the caller a flush is still owed, and only a hard failure is logged-and-restored.

I kept the read itself succeeding rather than returning WS_WANT_WRITE: the bytes are already copied into out and idx has advanced, so returning an error would lose them. The adjust is bundled into outputBuffer and goes out on the next flush, and the window is credited locally either way, so no credit leaks.

src/internal.c
```diff
@@ -11202,23 +11314,30 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
...
+                ssh->lastRxId = channelId;
+                ret = WS_EXTDATA;
             }
         }
+        else {
+            WLOG(WS_LOG_INFO, "Ignoring unknown extended data type %u",
+                    dataTypeCode);
+            (void)ChannelCreditWindow(ssh, channel, dataSz);
+            ssh->lastRxId = channelId;
+            ret = WS_SUCCESS;
```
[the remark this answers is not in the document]
ejohnstown: Partly fixed. Hard failures from ChannelCreditWindow() now propagate; WS_WANT_WRITE and WS_OVERFLOW_E return WS_SUCCESS.

Not propagating WS_WANT_WRITE is deliberate. The packet is consumed and the credit is either parked in pendingWindowAdjust or already bundled in outputBuffer, so no credit is lost and the channel can't stall. Returning it from a receive-path decoder would also suppress the HighwaterCheck() fold in DoPacket() and turn "I ignored your packet" into a read-loop wakeup.

Setting ssh->error instead isn't an option either: DoReceive() overwrites it with DoPacket()'s return unconditionally.

src/ssh.c
```diff
@@ -2814,7 +2855,9 @@ int wolfSSH_worker(WOLFSSH* ssh, word32* channelId)
     }
 #endif /* WOLFSSH_TEST_BLOCK */
 
-    if (ret == WS_SUCCESS || ret == WS_CHAN_RXD) {
+    /* WS_EXTDATA reports the channel too, so a multi-channel caller can route
+     * the drain to wolfSSH_ChannelIdReadExt(). */
+    if (ret == WS_SUCCESS || ret == WS_CHAN_RXD || ret == WS_EXTDATA) {
```
aidangarske, line 2972: **🔴 [High] wolfSSH_worker masks extended-data events during rekey** · _Logic / State Management_

The PR added WS_EXTDATA to the same post-receive block that converts successful/channel-data statuses into WS_REKEYING when ssh->isKeying is set. When a MSGID_CHANNEL_EXTENDED_DATA packet arrives during rekey, DoChannelExtendedData() buffers the stderr, charges channel->windowSz, sets lastRxId, and returns WS_EXTDATA; wolfSSH_worker() then overwrites both the return and ssh->error with WS_REKEYING. Callers following the new documented multi-channel contract only drain stderr when worker returns WS_EXTDATA (or branch on wolfSSH_get_error(), e.g. the SCP receive loop), so the buffered stderr may never be drained or credited after rekey completes. Because this PR moved receive-window credit for stderr to the drain path, the hidden buffered data stays charged, which can stall the channel. Severity disagreement: the review mode rates this High/BLOCK; review-security and bugs rate it Medium. Stricter view kept. New tests only exercise worker reporting when ssh->isKeying == 0.

**Fix:** Do not route WS_EXTDATA through the rekey override. Still write *channelId, but apply the keying override only to WS_SUCCESS and WS_CHAN_RXD (e.g. gate with `ssh->isKeying && ret != WS_EXTDATA`) so applications can drain the channel that received stderr during rekey. Add a worker regression with ssh->isKeying set and a received extended-data packet, verifying the app-visible status still triggers a drain.

src/ssh.c
```diff
@@ -3116,6 +3163,66 @@ static int _ChannelRead(WOLFSSH_CHANNEL* channel, byte* buf, word32 bufSz)
...
+
+    bufSz = min(bufSz, extDataBuffer->length - extDataBuffer->idx);
+    WMEMCPY(buf, extDataBuffer->buffer + extDataBuffer->idx, bufSz);
+    extDataBuffer->idx += bufSz;
+
+    if (bufSz > 0) {
+        int adjustResult;
+        int savedError = ssh->error;
+
+        /* Credit locally regardless of the send result; ChannelCreditWindow()
```
aidangarske, line 3310: **🔴 [High] Parked receive-window credit is counted as advertised before the peer is credited** · _Logic / Window Accounting_

The PR added pendingWindowAdjust for deferred WINDOW_ADJUST crediting, but callers add drained bytes back to channel->windowSz before the adjust is actually sent. _ChannelReadExt() (src/ssh.c:3196-3204) increments windowSz before ChannelCreditWindow() can park the credit during rekey or a short/non-blocking write, _UpdateChannelWindow() (src/ssh.c:3130-3135) reopens the local window even when ChannelCreditWindow() only parked the credit under ssh->isKeying, and the unknown extended-data path (src/internal.c:11336-11340) never subtracts dataSz before parking credit. Because DoChannelExtendedData() now uses channel->windowSz to reject data larger than the advertised receive window, pending unsent credit makes that check compare against local buffer capacity rather than what the peer was actually credited to send. A nonconforming/malicious peer can send past the advertised window while credit is parked (e.g. across a long rekey), defeating the advertised-window enforcement this PR added; sustained parking can also grow pendingWindowAdjust toward overflow. Severity disagreement: review rates this High/BLOCK, review-security rates it Medium; stricter view kept.

**Fix:** Keep the receive-window value used for inbound enforcement tied to credit actually sent or at least queued for delivery. For rekey/short-write parked credit, defer increasing channel->windowSz until SendPendingChannelWindowAdjust() successfully queues/sends the adjust, or split local buffer-capacity accounting from advertised-window accounting. Add a regression that parks credit (drain stderr or channel data during rekey), then has the peer send beyond the pre-rekey advertised window before the deferred adjust flushes; it should be rejected.

src/internal.c
```diff
@@ -11202,23 +11314,41 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
...
+                ssh->lastRxId = channelId;
+                ret = WS_EXTDATA;
+            }
+        }
+        else {
+            int creditResult;
+
+            WLOG(WS_LOG_INFO, "Ignoring unknown extended data type %u",
+                    dataTypeCode);
+            creditResult = ChannelCreditWindow(ssh, channel, dataSz);
```
aidangarske, line 11541: **🟠 [Medium] Ignored extended-data credit errors are reported as success** · _Error Handling_

For unknown extended-data types, the PR discards the payload and calls ChannelCreditWindow() to return receive-window credit, but then maps both WS_WANT_WRITE and WS_OVERFLOW_E to WS_SUCCESS. With a non-blocking send that returns want-write, ChannelCreditWindow() has queued a WINDOW_ADJUST in outputBuffer; returning success lets DoReceive() overwrite ssh->error with success, and direct wolfSSH_stream_read() callers do not flush pending output before reading again. With WS_OVERFLOW_E, no credit for this discarded payload was sent or parked, yet the packet is still consumed as successful. Introduced by the new unknown-type ignore path; other ChannelCreditWindow() callers either propagate the status or preserve WS_WANT_WRITE. RFC 4254 5.2 requires extended data to consume the same channel window as ordinary data. This is a distinct error-handling defect co-located with the window-accounting finding at the same lines.

**Fix:** Only treat delivered or deliberately parked credit as success, and surface retry/failure statuses so callers can flush or fail instead of silently losing the credit signal, e.g. `if (creditResult == WS_SUCCESS) ret = WS_SUCCESS; else ret = creditResult;`.

src/ssh.c
```diff
@@ -1253,6 +1254,16 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
             ret = DoReceive(ssh);
             if (ssh->channelList == NULL || ssh->channelList->eofRxd)
                 ret = WS_EOF;
+            if (ret == WS_EXTDATA &&
```
aidangarske, line 1257: **🔵 [Low] stream_read returns WS_ERROR for non-head extended data but leaves ssh-error == WS_EXTDATA**

When `DoReceive()` returns `WS_EXTDATA` for a channel that is not the head of the channel list, the new filter sets `ret = WS_ERROR; break;` without updating `ssh->error`. Because `DoReceive()`/`DoPacket()` already stored `ssh->error = WS_EXTDATA` (src/internal.c:12157), a caller that inspects `wolfSSH_get_error()` after the `WS_ERROR` return sees `WS_EXTDATA` rather than a want/error status. A multi-channel app that (incorrectly, per the documented contract) still calls `wolfSSH_stream_read()` could misinterpret this and, e.g., call `wolfSSH_extended_data_read()` on the head channel — which drains the wrong channel — or treat it as a fatal error and tear down, stranding the buffered stderr on the other channel (its window already charged). Impact is limited: the public header explicitly directs multi-channel apps to `wolfSSH_worker()` + `wolfSSH_ChannelIdReadExt()`, and single-channel apps never hit this branch (lastRxId always equals the head channel). Note this matches the pre-existing WS_CHAN_RXD filter a few lines below (which also sets WS_ERROR without touching ssh->error), so it is consistent with current convention.

**Fix:** Consider clarifying the ssh->error state on this WS_ERROR path; low priority since it matches the existing WS_CHAN_RXD convention and multi-channel apps are steered to the Id-based readers.

aidangarske (commented):
## Skoll Multi-Scan Review

**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [High] [review+review-security+bugs] **wolfSSH_worker masks extended-data events during rekey** — `src/ssh.c:2860-2867`
- [High] [review+review-security] **Parked receive-window credit is counted as advertised before the peer is credited** — `src/ssh.c:3196-3204`
- [Medium] [bugs] **Ignored extended-data credit errors are reported as success** — `src/internal.c:11336-11340`
- [Low] **stream_read returns WS_ERROR for non-head extended data but leaves ssh-error == WS_EXTDATA** — `src/ssh.c:1257-1266`

### Findings not tied to a diff line

#### wolfSSH_extended_data_read now rejects outSz==0 with WS_BAD_ARGUMENT (behavior change)
**File:** `src/ssh.c:1456-1470 (via _ChannelReadExt guard at src/ssh.c:3176)`
**Function:** `wolfSSH_extended_data_read / _ChannelReadExt`
**Severity:** Low

The rewritten `wolfSSH_extended_data_read()` delegates to `_ChannelReadExt()`, whose first guard is `if (channel == NULL || buf == NULL || bufSz == 0) return WS_BAD_ARGUMENT;`. The previous implementation computed `bufSz = min(outSz, length - idx)` and, for `outSz == 0`, copied 0 bytes and returned `0`. So a caller that passes `outSz == 0` now gets a negative error code (`WS_BAD_ARGUMENT`) instead of `0`. A drain loop written as `while ((n = wolfSSH_extended_data_read(ssh, buf, room)) > 0)` where `room` can legitimately reach 0 (e.g. a full application buffer) would previously terminate cleanly on the `0` return and now terminates on a negative value that reads as a failure. The in-tree callers (apps/wolfssh/wolfssh.c, examples/client/client.c, src/wolfscp.c) all pass `SIZE-1 > 0`, so they are unaffected, and the change is asserted intentionally by `test_ChannelReadExtBadArgs`. Flagging so the behavior change is a conscious, documented decision for external API consumers.

**Recommendation:** Confirm the outSz==0 -> WS_BAD_ARGUMENT change is intended and document it for API consumers, or special-case outSz==0 to return 0 to preserve prior behavior.

_Referenced code: `src/ssh.c:1456-1470 (via _ChannelReadExt guard at src/ssh.c:3176)` (2 lines)_

---

src/ssh.c
ejohnstown: Fixed. wolfSSH_worker() no longer folds WS_EXTDATA into WS_REKEYING; it is raised once, on arrival, so masking it stranded the buffer and its credit. Added test_WorkerReportsExtDataChannelKeying.

src/ssh.c
ejohnstown: Disagree. windowSz tracks free buffer capacity, not the peer's belief. Parking only makes the peer's view lower than windowSz, never higher, so the dataSz > windowSz check still bounds buffered bytes exactly.

The proposed fix regresses interop: holding windowSz down until the adjust flushes rejects data the peer legitimately sent before it saw our KEXINIT, which RFC 4253 section 7.1 requires accepting.

Not subtracting dataSz on the unknown-type path is correct -- the payload is discarded, no capacity is consumed, and the credit re-syncs the peer's counter.

pendingWindowAdjust growth needs 4 GB accumulated inside one rekey, and the guard's outcome is an under-credited peer throttling itself.

src/internal.c
ejohnstown: Disagree; same ground as previous. The suggested else ret = creditResult; breaks the flush it is protecting: WS_WANT_WRITE propagates through DoReceive() and then fails the flush guard in wolfSSH_worker(), so the bundled WINDOW_ADJUST never goes out. Returning WS_SUCCESS is what sends it, on the same worker call. It would also suppress the HighwaterCheck() fold.

WS_OVERFLOW_E does lose the credit, but it is unreachable for the reasons in the sibling thread.

src/ssh.c
ejohnstown: Declining, as the finding suggests. It matches the existing WS_CHAN_RXD convention twelve lines below; changing one and not the other makes the path less consistent. Multi-channel apps are directed to wolfSSH_ChannelIdReadExt().

JacobBarthelmeh (dismissed):

src/internal.c
```diff
@@ -11399,23 +11519,41 @@ static int DoChannelExtendedData(WOLFSSH* ssh,
...
+            ssh->lastRxId = channelId;
+            if (creditResult == WS_SUCCESS || creditResult == WS_WANT_WRITE ||
```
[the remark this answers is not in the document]
ejohnstown: Fixed.

tests/unit.c
```diff
@@ -3482,6 +3482,1760 @@ static int test_DoChannelWindowAdjust_overflow(void)
...
+    }
+
+done:
+    wolfSSH_free(ssh);
+    wolfSSH_CTX_free(ctx);
+    return result;
+}
+
+/* Counts IoSend calls so a test can assert whether anything went on the wire. */
+static word32 s_extSendCount = 0;
```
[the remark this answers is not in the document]
ejohnstown: Declined. The test runs sequentially.

JacobBarthelmeh (approved):

PR #1055, opened 2026-06-24
Harden KEXINIT parsing

ejohnstown:
- Strip trailing comma from peer name lists (F-2478): GetNameListRaw folded a legal trailing comma (e.g. aes128-cbc,) into the final name, decoding it as ID_UNKNOWN and failing negotiation. Now trims it, mirroring AlgoListSz.
- Bound KEXINIT language name-list skips (F-5576): the two language lists were skipped via an unchecked GetUint32 + begin += skipSz, letting a forged length wrap the parse offset back into earlier payload bytes and still return success. Both now use GetSkip, which validates the length against the remaining payload.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -4181,6 +4181,11 @@ static int GetNameListRaw(byte* idList, word32* idListSz,
         return WS_BAD_ARGUMENT;
     }
 
+    /* Adjust nameListSz for a trailing comma. */
+    if (nameListSz > 0 && nameList[nameListSz - 1] == ',') {
+        nameListSz--;
+    }
+
     /* Reject oversized name-lists to bound the per-token NameToId scan cost.
      * Applies to every list parsed here; the built-in canned lists are far
      * under these caps. */
--- a/src/internal.c
+++ b/src/internal.c
@@ -5059,17 +5059,13 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     /* Languages - Client to Server, skip */
     if (ret == WS_SUCCESS) {
         WLOG(WS_LOG_DEBUG, "DKI: Languages - Client to Server");
-        ret = GetUint32(&skipSz, buf, len, &begin);
-        if (ret == WS_SUCCESS)
-            begin += skipSz;
+        ret = GetSkip(buf, len, &begin);
     }
 
     /* Languages - Server to Client, skip */
     if (ret == WS_SUCCESS) {
         WLOG(WS_LOG_DEBUG, "DKI: Languages - Server to Client");
-        ret = GetUint32(&skipSz, buf, len, &begin);
-        if (ret == WS_SUCCESS)
-            begin += skipSz;
+        ret = GetSkip(buf, len, &begin);
     }
 
     /* First KEX Packet Follows */
--- a/tests/regress.c
+++ b/tests/regress.c
@@ -141,6 +141,12 @@ static word32 AppendUint32(byte* buf, word32 bufSz, word32 idx, word32 value)
     return idx;
 }
 
+static word32 ReadUint32(const byte* buf)
+{
+    return ((word32)buf[0] << 24) | ((word32)buf[1] << 16) |
+            ((word32)buf[2] << 8) | (word32)buf[3];
+}
+
 static word32 AppendData(byte* buf, word32 bufSz, word32 idx,
         const byte* data, word32 dataSz)
 {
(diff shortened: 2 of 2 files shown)


tests/regress.c
```diff
@@ -2980,6 +2980,88 @@ static void TestKexInitNameListCaps(void)
...
+    AssertNotNull(ctx);
+    ssh = wolfSSH_new(ctx);
+    AssertNotNull(ssh);
+    AssertIntEQ(wolfSSH_SetAlgoListKex(ssh, FPF_KEX_GOOD), WS_SUCCESS);
+    AssertIntEQ(wolfSSH_SetAlgoListKey(ssh, FPF_KEY_GOOD), WS_SUCCESS);
+
+    payloadSz = BuildKexInitPayload(ssh, FPF_KEX_GOOD, FPF_KEY_GOOD,
+            0, payload, (word32)sizeof(payload));
+
+    /* Walk past the cookie and the eight algorithm name-lists to reach the
```
aidangarske, line 3042: **🔵 [Low] Second KEXINIT language skip (server-to-client) has no independent regression coverage** · _test_

The PR hardens both language name-list skips in DoKexInit: the client-to-server list at src/internal.c:4730 and the server-to-client list at src/internal.c:4736, both now bounded via GetSkip. The new regression test only corrupts the first (client-to-server) language length: it walks past the cookie and the eight algorithm name-lists to reach 'off', overwrites that C2S length with 0xFFFFFFFC, and asserts WS_BUFFER_E. Because parsing fails at the C2S list, execution never reaches the S2C GetSkip at internal.c:4736, so that changed branch is never exercised. If a future refactor reverted only the S2C skip back to the unchecked begin += skipSz, this test would still pass. The two call sites are identical, so this is a coverage gap rather than a current correctness bug.

**Fix:** Add a mirrored case (or parameterize the test over the two language fields) that advances past the first language name-list and corrupts the second (server-to-client) language length, confirming it is also rejected with WS_BUFFER_E so both changed branches at internal.c:4730 and internal.c:4736 are independently locked.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 2 total — 1 posted, 1 skipped

### Posted findings
- [Low] [review] **Second KEXINIT language skip (server-to-client) has no independent regression coverage** — `tests/regress.c:3042-3059`

### Skipped findings
- [Info] ~~Single trailing-comma trim changes outcome for doubly-trailing-comma name-lists~~

tests/regress.c
```diff
@@ -2980,6 +2980,88 @@ static void TestKexInitNameListCaps(void)
...
+     * first (client-to-server) language length field. */
+    off = COOKIE_SZ;
+    for (i = 0; i < 8; i++)
+        off += UINT32_SZ + ReadUint32(payload + off);
+
+    /* 0xFFFFFFFC = 2^32 - 4. After reading this 4-byte length the parser is
+     * 4 bytes past 'off', so the old unchecked begin += skipSz wraps begin
+     * back to 'off' itself -- in bounds, not past the buffer. The buggy
```
aidangarske: remove emdash? comment seems a little too long maybe reduce the size

aidangarske (commented):
Looks good!

tests/regress.c
ejohnstown: Changed it. That comment is wordy.

tests/regress.c
ejohnstown: Added the test change.

tests/regress.c
```diff
@@ -2980,6 +2980,111 @@ static void TestKexInitNameListCaps(void)
...
+     * first (client-to-server) language name-list, to reach the second
+     * (server-to-client) language length field. */
+    off = COOKIE_SZ;
+    for (i = 0; i < 8; i++)
+        off += UINT32_SZ + ReadUint32(payload + off);
+    off += UINT32_SZ + ReadUint32(payload + off);
+
+    /* Corrupt the S2C language length with the same overflow value to
+     * verify GetSkip rejects it with WS_BUFFER_E, confirming the fix is
+     * applied to both language name-list skips at internal.c:4730 and 4736. */
```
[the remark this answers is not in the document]
ejohnstown: This comment was rewritten.

src/internal.c
```diff
@@ -4181,6 +4181,11 @@ static int GetNameListRaw(byte* idList, word32* idListSz,
         return WS_BAD_ARGUMENT;
     }
 
+    /* Adjust nameListSz for a trailing comma. */
+    if (nameListSz > 0 && nameList[nameListSz - 1] == ',') {
+        nameListSz--;
+    }
```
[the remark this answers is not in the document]
ejohnstown: I believe this to be a false positive. If the list has just a comma and a length of 1, the length becomes zero. Later on, Match doesn't, and returns ID_UNKNOWN, which becomes an error and we drop out of the function.

aidangarske (approved):

PR #1056, opened 2026-06-24
DH GEX group selection hardening

ejohnstown:
This fixes the server-side Diffie-Hellman Group Exchange (RFC 4419) group selection, which previously ignored the client's requested size window and always sent group 14, and adds a 2048-bit floor to prevent downgrade.

- Discard a guessed KEX packet in DoKexDhGexGroup: honor ignoreNextKexMsg so a wrong first_kex_packet_follows guess is silently discarded rather than parsed as a real group and errored, matching DoKexDhReply / DoKexDhInit (RFC 4253 7.1). (Issue: F-2866)
- Select the DH GEX group from the client's [min, preferred, max] size window instead of always sending group 14: add SelectKexDhGexGroup() to pick the closest fitting built-in group, reject with WS_DH_SIZE_E when none fits, and thread the WOLFSSH through GetDHPrimeGroup() so the sent group, exchange hash, and shared secret stay consistent. (Issue: F-55)
- Enforce a 2048-bit floor (WOLFSSH_DH_GEX_MIN_BITS) on group selection: clamp the client's lower bound up before the scan, reject windows whose max is below the floor, cap the window to MAX_KEX_KEY_SZ, and extend the unit tests for the rejection, clamp-up, and cache-miss fallback paths. (Issue: F-55)

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -5533,6 +5533,23 @@ static const byte dhPrimeGroup16[] = {
 static const word32 dhPrimeGroup16Sz = (word32)sizeof(dhPrimeGroup16);
 #endif
 
+/* RFC 4253 sec. 7.1: discard the KEX packet a peer sent after a wrong
+ * first_packet_follows guess. Leave expectMsgId at MSGID_NONE -- the real
+ * follow-up's ID comes from the negotiated KEX, not the guess, and may cross
+ * the GEX boundary. Returns 1 if consumed, else 0. Caller validates handshake. */
+static int SkipGuessedKexMsg(WOLFSSH* ssh, const char* what,
+        word32 len, word32* idx)
+{
+    WOLFSSH_UNUSED(what);
+    if (!ssh->handshake->ignoreNextKexMsg)
+        return 0;
+    WLOG(WS_LOG_DEBUG, "Skipping %s due to first_packet_follows guess mismatch.",
+            what);
+    ssh->handshake->ignoreNextKexMsg = 0;
+    *idx += len;
+    return 1;
+}
+
 static int DoKexDhInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 {
     /* First get the length of the MP_INT, and then add in the hash of the
@@ -5554,14 +5571,9 @@ static int DoKexDhInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ret = WS_BAD_ARGUMENT;
 
     if (ret == WS_SUCCESS) {
-        if (ssh->handshake->ignoreNextKexMsg) {
-            /* skip this message. */
-            WLOG(WS_LOG_DEBUG, "Skipping client's KEXDH_INIT message due to "
-                               "first_packet_follows guess mismatch.");
-            ssh->handshake->ignoreNextKexMsg = 0;
-            *idx += len;
+        if (SkipGuessedKexMsg(ssh, "client's KEXDH_INIT message",
+                len, idx))
             return WS_SUCCESS;
-        }
 
         begin = *idx;
         ret = GetStringRef(&eSz, &e, buf, len, &begin);
@@ -6781,14 +6793,9 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
     }
 
     if (ret == WS_SUCCESS) {
-        if (ssh->handshake->ignoreNextKexMsg) {
-            /* skip this message. */
-            WLOG(WS_LOG_DEBUG, "Skipping server's KEXDH_REPLY message due to "
-                               "first_packet_follows guess mismatch.");
-            ssh->handshake->ignoreNextKexMsg = 0;
-            *idx += len;
+        if (SkipGuessedKexMsg(ssh, "server's KEXDH_REPLY message",
+                len, idx))
             return WS_SUCCESS;
-        }
     }
 
     if (ret == WS_SUCCESS && len < LENGTH_SZ*2 + *idx) {
@@ -7232,14 +7239,9 @@ static int DoKexDhGexRequest(WOLFSSH* ssh,
         ret = WS_BAD_ARGUMENT;
 
     if (ret == WS_SUCCESS) {
-        if (ssh->handshake->ignoreNextKexMsg) {
-            /* skip this message. */
-            WLOG(WS_LOG_DEBUG, "Skipping client's KEXDH_GEX_REQUEST message "
-                               "due to first_packet_follows guess mismatch.");
-            ssh->handshake->ignoreNextKexMsg = 0;
-            *idx += len;
+        if (SkipGuessedKexMsg(ssh, "client's KEXDH_GEX_REQUEST message",
+                len, idx))
             return WS_SUCCESS;
-        }
 
         begin = *idx;
         ret = GetUint32(&ssh->handshake->dhGexMinSz, buf, len, &begin);
@@ -7411,10 +7413,19 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
     word32 begin;
     int ret = WS_SUCCESS;
 
-    if (ssh == NULL || buf == NULL || len == 0 || idx == NULL)
+    if (ssh == NULL || ssh->handshake == NULL || buf == NULL || len == 0 ||
+            idx == NULL)
         ret = WS_BAD_ARGUMENT;
 
     if (ret == WS_SUCCESS) {
+        /* A conformant server sends GROUP only in response to the client's
+         * REQUEST, so it should never set first_packet_follows here. Discard
+         * the message defensively if a peer sets it anyway, mirroring the other
+         * Do* handlers. */
+        if (SkipGuessedKexMsg(ssh, "server's KEXDH_GEX_GROUP message",
+                len, idx))
+            return WS_SUCCESS;
+
         begin = *idx;
         ret = GetMpint(&primeGroupSz, &primeGroup, buf, len, &begin);
         if (ret == WS_SUCCESS && primeGroupSz > (MAX_KEX_KEY_SZ + 1)) {
@@ -20095,6 +20106,12 @@ int wolfSSH_TestDoKexDhGexRequest(WOLFSSH* ssh, byte* buf, word32 len,
         word32* idx)
 {
     return DoKexDhGexRequest(ssh, buf, len, idx);
+}
+
+int wolfSSH_TestDoKexDhGexGroup(WOLFSSH* ssh, byte* buf, word32 len,
+        word32* idx)
+{
+    return DoKexDhGexGroup(ssh, buf, len, idx);
 }
 
 int wolfSSH_TestValidateKexDhGexGroup(const byte* primeGroup,
--- a/src/internal.c
+++ b/src/internal.c
@@ -12778,11 +12778,77 @@ static int BuildRFC6187Info(WOLFSSH* ssh, int pubKeyID,
 #endif /* WOLFSSH_CERTS */
 
 
+#ifndef WOLFSSH_NO_DH_GEX_SHA256
+/* Pick the in-window group closest to preferredBits, ties favoring the
+ * smaller. WS_DH_SIZE_E if none fits. */
+static int SelectKexDhGexGroup(word32 minBits, word32 preferredBits,
+        word32 maxBits, const byte** primeGroup, word32* primeGroupSz)
+{
+    static const struct {
+        word32 bits;
+        const byte* group;
+        const word32* groupSz;
+    } candidates[] = {
+        #ifndef WOLFSSH_NO_DH_GROUP1_SHA1
+        { 1024, dhPrimeGroup1,  &dhPrimeGroup1Sz  },
+        #endif
+        { 2048, dhPrimeGroup14, &dhPrimeGroup14Sz },
+        #ifndef WOLFSSH_NO_DH_GROUP16_SHA512
+        { 4096, dhPrimeGroup16, &dhPrimeGroup16Sz },
+        #endif
+    };
+    word32 i;
+    word32 best = 0;
+    word32 bestDelta = 0;
+    int haveBest = 0;
+    int ret = WS_SUCCESS;
+
+    if (primeGroup == NULL || primeGroupSz == NULL)
+        return WS_BAD_ARGUMENT;
+
+    for (i = 0; i < (word32)(sizeof(candidates) / sizeof(candidates[0])); i++) {
+        word32 bits = candidates[i].bits;
+        word32 delta;
+
+        if (bits < minBits || bits > maxBits)
+            continue;
+        delta = (bits > preferredBits) ? (bits - preferredBits)
+                                       : (preferredBits - bits);
+        /* Ascending scan with a strict '<' keeps the smaller group on a tie. */
+        if (!haveBest || delta < bestDelta) {
+            best = i;
+            bestDelta = delta;
+            haveBest = 1;
+        }
+    }
+
+    if (!haveBest) {
+        WLOG(WS_LOG_DEBUG,
+                "DH GEX: no built-in group within client window [%u, %u]",
+                minBits, maxBits);
+        ret = WS_DH_SIZE_E;
+    }
+    else {
+        *primeGroup = candidates[best].group;
+        *primeGroupSz = *candidates[best].groupSz;
+    }
+
+    return ret;
+}
+#endif /* !WOLFSSH_NO_DH_GEX_SHA256 */
+
+
 #ifndef WOLFSSH_NO_DH
-static int GetDHPrimeGroup(int kexId, const byte** primeGroup,
+static int GetDHPrimeGroup(WOLFSSH* ssh, const byte** primeGroup,
     word32* primeGroupSz, const byte** generator, word32* generatorSz)
 {
     int ret = WS_SUCCESS;
+    int kexId;
+
+    if (ssh == NULL || ssh->handshake == NULL)
+        return WS_BAD_ARGUMENT;
+
+    kexId = ssh->handshake->kexId;
 
     switch (kexId) {
         #ifndef WOLFSSH_NO_DH_GROUP1_SHA1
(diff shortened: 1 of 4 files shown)


src/internal.c
```diff
@@ -6765,10 +6765,21 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
...
-    if (ssh == NULL || buf == NULL || len == 0 || idx == NULL)
+    if (ssh == NULL || ssh->handshake == NULL || buf == NULL || len == 0 ||
+            idx == NULL)
         ret = WS_BAD_ARGUMENT;
 
     if (ret == WS_SUCCESS) {
+        if (ssh->handshake->ignoreNextKexMsg) {
+            /* Defensive symmetry with the other Do* handlers; a server never
+             * guesses the GROUP response, so this is not expected to fire. */
+            WLOG(WS_LOG_DEBUG, "Skipping server's KEXDH_GEX_GROUP message due "
```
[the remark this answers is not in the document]
ejohnstown: The comment wording was worth fixing, but the rationale isn't right. In RFC 4419 GEX the client sends the first KEX packet (KEX_DH_GEX_REQUEST); the server's GROUP is a response to the client's window, so a conformant server can't send it as a guess and never sets first_packet_follows. Hence this handler is client-side — the branch only fires for a non-conformant/adversarial server, which is exactly what the regress test covers. Reworded so it reads as deliberate defense (RFC 4253 7.1) rather than dead code.

src/internal.c
```diff
@@ -11757,11 +11770,107 @@ static int BuildRFC6187Info(WOLFSSH* ssh, int pubKeyID,
...
+                                       : (preferredBits - bits);
+        /* Ascending scan with a strict '<' keeps the smaller group on a tie. */
+        if (!haveBest || delta < bestDelta) {
+            best = i;
+            bestDelta = delta;
+            haveBest = 1;
+        }
+    }
+
+    if (!haveBest) {
```
aidangarske, line 12836: **🟠 [Medium] Server now rejects GEX windows it can't satisfy (behavior change vs. always-group14)** · _Behavior change / interop_

Previously the server unconditionally sent group 14 (2048-bit) for every GEX request. Now SelectKexDhGexGroup clamps the client's minBits up to WOLFSSH_DH_GEX_MIN_BITS (2048) and returns WS_DH_SIZE_E when no built-in group falls in the effective window. This is the intended hardening, but it is an externally observable behavior change: a client that requests a window not covered by the compiled-in groups now gets a hard handshake failure instead of a working exchange. Two interop cases worth confirming: (a) a client requesting min > 2048 when WOLFSSH_NO_DH_GROUP16_SHA512 is defined (only 2048 available) — the server now fails the handshake; (b) a build where WOLFSSH_DEFAULT_GEXDH_MAX/MAX_KEX_KEY_SZ caps maxBits below a client's minimum. The stock wolfSSH client (min=1024, pref=3072, max=8192) still resolves to 2048, so wolfSSH↔wolfSSH is unaffected. Verified: minBits is unconditionally raised to the 2048 floor (11822-11823), maxBits is capped to MAX_KEX_KEY_SZ*8 (11828-11829), and the empty-window path returns WS_DH_SIZE_E (11847-11851).

**Fix:** Confirm this rejection path is acceptable for the interop matrix (especially clients configured with min>2048 on builds without group 16). Optionally log at WS_LOG_INFO/WARN rather than DEBUG so the rejection is diagnosable in the field.

tests/unit.c
```diff
@@ -1131,6 +1131,270 @@ static int test_DhGexGroupValidate(void)
...
+
+/* End-to-end consistency check for DH GEX server group selection: the group
+ * SendKexDhGexGroup puts on the wire must be the exact group GetDHPrimeGroup
+ * hands the exchange-hash and key-agreement path. The default window resolves
+ * to the historical 2048-bit group 14, so every stock GEX handshake test would
+ * pass even if the two paths disagreed. This drives a client window that forces
+ * the 4096-bit group 16 and confirms the cached wire group and the group that
+ * GetDHPrimeGroup returns are one and the same -- the property the send/hash
+ * cache exists to guarantee. */
+static int test_DhGexGroupSendHashConsistency(void)
```
aidangarske: **🔵 [Low] Send/hash consistency and cache-miss tests are compiled out on group14-only builds** · _Test coverage_

Both test_DhGexGroupSendHashConsistency and test_DhGexGroupCacheMissFallback are entirely wrapped in #ifndef WOLFSSH_NO_DH_GROUP16_SHA512 and return 0 otherwise (verified: unit.c:1248/1318-1322 and 1330). Since group 16 (4096-bit) is disabled in many default configurations, the send-vs-hash cache consistency guarantee — the core property this PR adds — goes completely unexercised on those builds. The consistency is trivially true when only group 14 exists, so this is a reduced-coverage note rather than a correctness gap, but the wire-vs-hash divergence the cache exists to prevent is only tested when group 16 is present.

**Fix:** Consider a group14-only variant that still asserts GetDHPrimeGroup returns the exact cached handshake->primeGroup pointer/size after SendKexDhGexGroup, so the caching path is covered on default builds too.

tests/unit.c
```diff
@@ -1131,6 +1131,270 @@ static int test_DhGexGroupValidate(void)
...
+    const byte* group;
+    word32 groupSz;
+    int ret;
+
+    /* Default-ish window: with group 16 enabled, preferred 3072 is equidistant
+     * from 2048 and 4096 and the tie favors the smaller group; without group 16,
+     * 2048 is simply the closest in-window candidate. Either way a stock client
+     * still receives the historical 2048-bit group (256 bytes). */
+    group = NULL; groupSz = 0;
+    ret = wolfSSH_TestSelectKexDhGexGroup(1024, 3072, 8192, &group, &groupSz);
```
aidangarske: **🔵 [Low] DH GEX selection tests ignore configurable floor** · _Logic Errors_

The PR adds WOLFSSH_DH_GEX_MIN_BITS as an overridable compile-time floor in src/internal.c, but the new unit tests assume that floor is always exactly 2048 bits and hard-code 256-byte group14 outcomes. With a valid non-default build such as -DWOLFSSH_DH_GEX_MIN_BITS=4096 and group16 enabled, wolfSSH_TestSelectKexDhGexGroup(1024, 3072, 8192, ...) correctly selects the 4096-bit group, but this test fails because it still expects groupSz == 256. With a lower override, the sub-2048 rejection/floor-clamp checks can similarly fail by allowing group1. This is introduced by the PR because it adds both the configurable floor and tests that do not derive expectations from it.

**Fix:** Make the floor visible to the tests and condition expected results on it, or only run the 2048-specific assertions when the floor is 2048. For example, move the default WOLFSSH_DH_GEX_MIN_BITS definition to a shared internal header, then guard these checks with #if WOLFSSH_DH_GEX_MIN_BITS == 2048 or compute the expected group size from the configured floor.

src/internal.c
```diff
@@ -13798,6 +13924,32 @@ int SendKexDhGexGroup(WOLFSSH* ssh)
...
+                ssh->handshake->dhGexMaxSz,
+                &primeGroup, &primeGroupSz);
+    }
+
+    /* Cache the selected group on the handshake so the exchange hash and the
+     * shared secret (GetDHPrimeGroup) reuse exactly what goes on the wire here,
+     * making the wire group the single source of truth instead of re-running
+     * the selection independently. */
+    if (ret == WS_SUCCESS) {
+        if (ssh->handshake->primeGroup)
```
aidangarske: **🔵 [Low] New null checks use implicit-pointer form instead of != NULL** · _style_

The new caching block null-checks with `if (ssh->handshake->primeGroup)` (verified at internal.c:13941) rather than the wolfSSL-preferred `if (ssh->handshake->primeGroup != NULL)`. GetDHPrimeGroup (line 11915) correctly uses != NULL. Behavior is identical and it mirrors the adjacent pre-existing style in DoKexDhGexGroup, so it is locally consistent and purely a style-consistency note, not a defect. Both reviewers flagged the same line: the review mode rated it Low/NIT and the review-security mode rated it Info; the stricter Low is retained here.

**Fix:** Use `if (ssh->handshake->primeGroup != NULL)` for the new pointer check to match wolfSSL C style; leave the pre-existing block unchanged and do not reformat surrounding lines.

src/internal.c
```diff
@@ -11757,11 +11770,107 @@ static int BuildRFC6187Info(WOLFSSH* ssh, int pubKeyID,
...
+ * for (RFC 4419 sec. 3). */
+static int SelectKexDhGexGroup(word32 minBits, word32 preferredBits,
+        word32 maxBits, const byte** primeGroup, word32* primeGroupSz)
+{
+    static const struct {
+        word32 bits;
+        const byte* group;
+        const word32* groupSz;
+    } candidates[] = {
+        #ifndef WOLFSSH_NO_DH_GROUP1_SHA1
```
aidangarske, line 12794: **🔵 [Low] 1024-bit group 1 candidate is unreachable (dead entry)** · _convention_

The candidates[] table includes the 1024-bit group 1 (guarded by WOLFSSH_NO_DH_GROUP1_SHA1), but minBits is unconditionally raised to WOLFSSH_DH_GEX_MIN_BITS (2048) before the scan (verified at internal.c:11822-11823), so the bits < minBits check at 11835 always skips the 1024 entry. It can never be selected. This is intentional/defensive per the header comment, and harmless, but the entry is effectively dead code that could mislead a future reader into thinking 1024 is selectable via GEX.

**Fix:** Either drop the group 1 entry (the floor already guarantees it is never chosen) or keep a one-line note that it is present only for documentary parity; no functional change needed.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 6 total — 5 posted, 1 skipped

### Posted findings
- [Medium] [review] **Server now rejects GEX windows it can't satisfy (behavior change vs. always-group14)** — `src/internal.c:11847-11851`
- [Low] [review] **Send/hash consistency and cache-miss tests are compiled out on group14-only builds** — `tests/unit.c:1246-1330`
- [Low] [bugs] **DH GEX selection tests ignore configurable floor** — `tests/unit.c:1148-1209`
- [Low] [review+review-security] **New null checks use implicit-pointer form instead of != NULL** — `src/internal.c:13941`
- [Low] [review] **1024-bit group 1 candidate is unreachable (dead entry)** — `src/internal.c:11802-11804`

### Skipped findings
- [Info] ~~Client can now steer server to a 4096-bit GEX group (modest CPU amplification)~~

src/internal.c
ejohnstown: Declining. WOLFSSH_DH_GEX_MIN_BITS is #ifndef-guarded. Under -DWOLFSSH_DH_GEX_MIN_BITS=1024 the group 1 entry passes the bits < minBits filter and is selected - it's what makes a lowered floor work. Dropping it would silently reject every sub-2048 window instead of honoring it.

This also contradicts the "tests ignore configurable floor" comment from the same review, which assumes the floor is configurable.

Took your alternative: one-line note at the entry, and moved the default to wolfssh/internal.h so the knob is visible.

src/internal.c
ejohnstown: No working exchange was lost. Case (a) already failed before this PR: the old server sent group 14 regardless, and the client rejects it in ValidateKexDhGexGroup (internal.c:6684, bits < (int)minBits). This moves the failure server-side where it's reported accurately. Against a client that skips that check, the old behavior was a silent downgrade below the stated minimum — the bug being fixed.

Case (b) is near-unreachable: MAX_KEX_KEY_SZ is ML-KEM-sized by default (internal.h:479, 3168 bytes = 25344 bits), so the cap only binds if ML-KEM is disabled and WOLFSSH_DEFAULT_GEXDH_MAX is under 4096.

Keeping the rejection.

tests/unit.c
ejohnstown: Both Taken; same root cause. WOLFSSH_DH_GEX_MIN_BITS was only visible inside src/internal.c. Moved the default to wolfssh/internal.h.

On coverage: group16 is off more often than noted — internal.h:154-157 force-defines WOLFSSH_NO_DH_GROUP16_SHA512 whenever WOLFSSH_NO_SHA512 is set. Both tests now run on every build. The consistency test always exercises the default window (pointer identity between cached wire group and GetDHPrimeGroup), and forces group 16 when available; the cache-miss test compares against the recorded wire size instead of a hard-coded 512.

Verified by fault injection on a group14-only build: bypassing the cache now fails with hash group sz 256 != wire 256, where before it compiled to return 0. Both sizes are 256, so the pointer check is what catches it — a size-only assertion wouldn't.

tests/unit.c
ejohnstown: Both Taken; same root cause. WOLFSSH_DH_GEX_MIN_BITS was only visible inside src/internal.c. Moved the default to wolfssh/internal.h.

Confirmed your floor prediction: under -DWOLFSSH_DH_GEX_MIN_BITS=4096 the old test failed with default window got ret 0 sz 512 — selection correct, expectation wrong. 2048-specific assertions are now gated on #if WOLFSSH_DH_GEX_MIN_BITS == 2048, with floor-relative checks otherwise.

src/internal.c
ejohnstown: Style nit applied to the new line only; internal.c:6804 left alone as you suggested.

src/internal.c
```diff
@@ -7426,9 +7437,16 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
...
+
+        if (minBits < WOLFSSH_DH_GEX_MIN_BITS)
+            minBits = WOLFSSH_DH_GEX_MIN_BITS;
+
         ret = ValidateKexDhGexGroup(primeGroup, primeGroupSz,
                 generator, generatorSz,
-                ssh->handshake->dhGexMinSz,
+                minBits,
                 ssh->handshake->dhGexMaxSz,
                 ssh->rng);
```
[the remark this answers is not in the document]
ejohnstown: Declining. CreateMpint() is only on our server side, and it's fed a copy of a built-in group, dhPrimeGroup1/14/16 all start with 0xFF, so the leading-zero scan stops at i == 0 and the WMEMMOVE guarded by if (i > 0) never fires. The buffers DoKexDhGexGroup caches are client-side and never reach CreateMpint at all: the client hash path computes the pad by hand and hashes verbatim. If a server sends extra leading pad bytes and hashes without them, the client hashes them back as received, the hashes diverge, and the handshake fails at signature verification. Fails safe, no desync.

src/internal.c
```diff
@@ -5533,6 +5533,23 @@ static const byte dhPrimeGroup16[] = {
 static const word32 dhPrimeGroup16Sz = (word32)sizeof(dhPrimeGroup16);
 #endif
 
+/* RFC 4253 sec. 7.1: discard the KEX packet a peer sent after a wrong
+ * first_packet_follows guess. Leave expectMsgId at MSGID_NONE -- the real
```
aidangarske: remove emdash? shorten comment?

src/internal.c
```diff
@@ -5533,6 +5533,23 @@ static const byte dhPrimeGroup16[] = {
...
+        word32 len, word32* idx)
+{
+    WOLFSSH_UNUSED(what);
+    if (!ssh->handshake->ignoreNextKexMsg)
+        return 0;
+    WLOG(WS_LOG_DEBUG, "Skipping %s due to first_packet_follows guess mismatch.",
+            what);
+    ssh->handshake->ignoreNextKexMsg = 0;
+    *idx += len;
+    return 1;
```
aidangarske, line 5550: do we normally return magic numbers like this for failure codes?

src/internal.c
```diff
@@ -12767,11 +12772,94 @@ static int BuildRFC6187Info(WOLFSSH* ssh, int pubKeyID,
...
+        word32 bits;
+        const byte* group;
+        word32 groupSz;
+    } candidates[] = {
+        /* Cast sizeof to word32 to match the file's dhPrimeGroupNSz style; the
+         * named constants are not constant expressions usable in this static
+         * initializer. */
+        #ifndef WOLFSSH_NO_DH_GROUP1_SHA1
+        /* Only reachable with a lowered WOLFSSH_DH_GEX_MIN_BITS. */
+        { 1024, dhPrimeGroup1,  (word32)sizeof(dhPrimeGroup1)  },
```
aidangarske, line 12796: these bare scope bare should probably be removed

src/internal.c
```diff
@@ -12808,10 +12896,41 @@ static int GetDHPrimeGroup(int kexId, const byte** primeGroup,
...
-            *primeGroupSz = dhPrimeGroup14Sz;
-            *generator = dhGenerator;
-            *generatorSz = dhGeneratorSz;
+            /* Reuse the group SendKexDhGexGroup cached on the handshake so the
+             * exchange hash and the shared secret match the group that was put
+             * on the wire. An unset cache means no GROUP was ever sent (the
+             * peer skipped GEX_REQUEST); there is no wire group to match, so
+             * re-select from whatever window the handshake holds.
+             *
+             * handshake->primeGroup has a role-dependent meaning: on the server
```
aidangarske: Shorten comment?

src/internal.c
```diff
@@ -14989,6 +15138,33 @@ int SendKexDhGexGroup(WOLFSSH* ssh)
...
+        ret = SelectKexDhGexGroup(ssh->handshake->dhGexMinSz,
+                ssh->handshake->dhGexPreferredSz,
+                ssh->handshake->dhGexMaxSz,
+                &primeGroup, &primeGroupSz);
+    }
+
+    /* Cache the selected group so the exchange hash and shared secret
+     * (GetDHPrimeGroup) reuse exactly what goes on the wire here. */
+    if (ret == WS_SUCCESS) {
+        if (ssh->handshake->primeGroup != NULL)
```
aidangarske: can we use {} on these? kinda hard to read

wolfssh/internal.h
```diff
@@ -511,14 +511,63 @@ enum NameIdType {
...
+          !defined(WOLFSSH_NO_CURVE25519_MLKEM768_SHA256)
+        /* Size of ML-KEM-768 public key (1184) plus ECC/X25519 component. */
+        #define KEX_F_SIZE 1300
+    #elif !defined(WOLFSSH_NO_DH_GROUP16_SHA512)
+        #define KEX_F_SIZE (512 + 1)
+    #else
+        #define KEX_F_SIZE (256 + 1)
+    #endif
+#endif
+/* A user-supplied KEX_F_SIZE override must still hold the largest f/ciphertext
```
aidangarske: shorten comment? remove emdash?

aidangarske (commented):
Looks good just some nits

src/internal.c
ejohnstown: Shortened the comment. Replaced the double-dash with a comma.

src/internal.c
ejohnstown: Declining. It's a true or false return.

src/internal.c
ejohnstown: Decline. It's not a bare scope. That's a member of the candidates[] array initializer.

src/internal.c
ejohnstown: Trimmed the comment and put it in a better spot. The comment now describes the two roles those pointers have in server vs client connections.

src/internal.c
ejohnstown: Sure.

wolfssh/internal.h
ejohnstown: Reduced the size of the comment. It was restating the error string.

aidangarske (approved):

JacobBarthelmeh (approved):

PR #1057, opened 2026-06-24
Test DoChannelWindowAdjust overflow guard

ejohnstown:
Adds unit coverage for the overflow guard in `DoChannelWindowAdjust`, which rejects a peer window adjustment that would wrap `peerWindowSz` past `UINT32_MAX` with `WS_OVERFLOW_E`. The guard existed but was untested. (F-2874)

Test-only; no production behavior changes.


aidangarske (approved):
Looks good.

JacobBarthelmeh (approved):

PR #1058, opened 2026-06-24
Run threaded api-test SFTP/SCP tests on Windows

yosuke-wolfssl:
## Summary

Enables the threaded echoserver+client **SFTP/SCP integration tests** in `api-test` on Windows CI (MSVC + AddressSanitizer). These tests were previously stubbed out on Windows because the readiness handshake was POSIX-only, so a whole class of SFTP/SCP coverage never ran there. Running them surfaced several Windows-only bugs in wolfSSH, which this PR also fixes.

Net change is small (7 files, +105/−32) — most of the work was diagnosis.

### Test harness / CI
- **`windows-check.yml`** — fix the `user_settings.h` `#if 0`→`#if 1` flip so SSHD/SFTP/SCP/CERTS/X509 are actually compiled and tested (they were silently stubbing to `{ ; }`).
- **`wolfssh/test.h`** — implement the **Win32 readiness handshake** for `tcp_ready` (a `CRITICAL_SECTION` + manual-reset event, with `Init/Free/Wait/SignalTcpReady` Win32 branches). `WaitForSingleObject` is bounded by a 30s timeout so a stuck server can't hang CI. This lets the threaded tests run instead of being skipped.
- **`tests/api.c`** — drop the `!USE_WINDOWS_API` guards that stubbed the threaded SFTP/SCP tests; pair `InitTcpReady`/`FreeTcpReady`; use an **ephemeral port (`-p 0`)** on Windows like POSIX, so concurrent CI runs don't collide on a fixed port.

### wolfSSH Windows fixes uncovered by the tests
- **`MakeScpCmd` (`src/wolfscp.c`)** — the length-probe used `WSNPRINTF(NULL, 0, ...)`, but on Windows that maps to `_snprintf_s`, which rejects the NULL/0 form and aborts via `__fastfail`. Use `_scprintf` through a new Windows-only `WSCPRINTF` wrapper (`wolfssh/port.h`) for the length measurement.
- **`WSTRNCPY` call-site hardening** — on Windows `WSTRNCPY(s1,s2,n)` maps to `strncpy_s((s1),(n),(s2),(n))`, which `__fastfail`s when `strlen(src) >= n`. Audited every `src/` and example call site; fixed the unguarded ones:
  - `ScpCheckForRename` — pass the real buffer size (`sizeof(buf)`) rather than a copy count.
  - `ScpPushDir` — add an upstream length guard returning `WS_BUFFER_E`.
  - `echoserver.c` / `sftpclient.c` — guard the user/server-supplied string copies (skip or return on over-long input).
- **SFTP rekey-test shutdown flake (`tests/api.c`)** — on Windows the post-transfer close races the client's graceful `wolfSSH_shutdown`; Winsock raises an RST that the client sees on the **recv** path. There, wolfSSH follows its convention of returning the generic `WS_ERROR` while stashing the specific code in `ssh->error`. The test only absorbed the direct `WS_SOCKET_ERROR_E` return (the send-path manifestation), so the recv-path reset slipped through ~30–40% of runs. The fix absorbs the benign reset via `wolfSSH_get_error(ssh) == WS_SOCKET_ERROR_E`, gated on `argsCount == WS_ERROR` so a genuine error with a different `ssh->error` still fails — no masking.

## Verification
- **Windows** (winvs Release + MSVC AddressSanitizer, run from repo root): `api-test` + `unit-test` pass **15/15** (was ~30–40% flaky at the SFTP-rekey shutdown).
- **POSIX** (macOS): `api-test` + `testsuite` clean under `-fsanitize=address,undefined`.

## Notes
- No change to the `WSTRNCPY` macro mapping — fixes are at the call sites only.
- Win32 sync primitives (`CreateEvent`, `EnterCriticalSection`, `WaitForSingleObject`) are used directly, consistent with existing `USE_WINDOWS_API` code in the repo.

Supersedes #1030 (which this branch already contains in full; #1058 also un-stubs the threaded tests that #1030 had excluded).

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -2249,6 +2249,11 @@ static int LoadPasswdList(StrList* strList, PwMapList* mapList)
     int count = 0;
 
     while (strList) {
+        if (WSTRLEN(strList->str) >= sizeof names - 1) {
+            fprintf(stderr, "Ignoring over-long entry: %.32s\n", strList->str);
+            strList = strList->next;
+            continue;
+        }
         WSTRNCPY(names, strList->str, sizeof names - 1);
         passwd = WSTRCHR(names, ':');
         if (passwd != NULL) {
@@ -2278,6 +2283,11 @@ static int LoadKeyboardList(StrList* strList, PwMapList* mapList,
     int count = 0;
 
     while (strList) {
+        if (WSTRLEN(strList->str) >= sizeof names - 1) {
+            fprintf(stderr, "Ignoring over-long entry: %.32s\n", strList->str);
+            strList = strList->next;
+            continue;
+        }
         WSTRNCPY(names, strList->str, sizeof names - 1);
         passwd = WSTRCHR(names, ':');
         if (passwd != NULL) {
@@ -2314,6 +2324,11 @@ static int LoadPubKeyList(StrList* strList, int format, PwMapList* mapList)
         buf = NULL;
         bufSz = 0;
 
+        if (WSTRLEN(strList->str) >= sizeof names - 1) {
+            fprintf(stderr, "Ignoring over-long entry: %.32s\n", strList->str);
+            strList = strList->next;
+            continue;
+        }
         WSTRNCPY(names, strList->str, sizeof names - 1);
         fileName = WSTRCHR(names, ':');
         if (fileName != NULL) {
@@ -2834,6 +2849,12 @@ static INLINE void SignalTcpReady(tcp_ready* ready, word16 port)
     ready->port = port;
     pthread_cond_signal(&ready->cond);
     pthread_mutex_unlock(&ready->mutex);
+#elif defined(USE_WINDOWS_API) && defined(NO_MAIN_DRIVER) && \
+    !defined(SINGLE_THREADED)
+    ready->ready = 1;
+    ready->port = port;
+    /* SetEvent is a full barrier; this one-shot signal needs no lock */
+    SetEvent(ready->readyEvent);
 #else
     WOLFSSH_UNUSED(ready);
     WOLFSSH_UNUSED(port);
@@ -2979,7 +3000,7 @@ THREAD_RETURN WOLFSSH_THREAD echoserver_test(void* args)
                     }
                     else {
                         port = (word16)atoi(myoptarg);
-                        #if !defined(NO_MAIN_DRIVER) || defined(USE_WINDOWS_API)
+                        #if !defined(NO_MAIN_DRIVER)
                             if (port == 0) {
                                 ES_ERROR("port number cannot be 0");
                             }
--- a/examples/sftpclient/sftpclient.c
+++ b/examples/sftpclient/sftpclient.c
@@ -156,6 +156,7 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
     static word32 lastOutputTime = 0;
     static word32 lastPrintedBytes[2] = {0, 0};
     word32 elapsedTime;
+    word32 nameSz;
 #endif
     char buf[80];
     word64 longBytes = ((word64)bytes[1] << 32) | bytes[0];
@@ -176,13 +177,19 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
     }
     lastPrintedBytes[0] = bytes[0];
     lastPrintedBytes[1] = bytes[1];
-    if (WSTRNCMP(currentFile, name, WSTRLEN(name)) != 0) {
+    /* clamp over-long names: currentFile holds the truncated prefix, so the
+     * comparison must use the same bound or every call looks like a new file */
+    nameSz = (word32)WSTRLEN(name);
+    if (nameSz >= sizeof(currentFile))
+        nameSz = (word32)sizeof(currentFile) - 1;
+    if (WSTRNCMP(currentFile, name, nameSz) != 0 ||
+            currentFile[nameSz] != '\0') {
         startTime = current_time(1);
         lastOutputTime = 0; /* Reset timer for new file transfer */
         lastPrintedBytes[0] = 0;
         lastPrintedBytes[1] = 0;
-        WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
-        WSTRNCPY(currentFile, name, WOLFSSH_MAX_FILENAME);
+        WMEMSET(currentFile, 0, sizeof(currentFile));
+        WMEMCPY(currentFile, name, nameSz);
     }
     elapsedTime = currentTime - startTime;
     WSNPRINTF(buf, sizeof(buf), "Processed %8llu\t bytes in %d seconds\r",
@@ -191,7 +198,7 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
     if (elapsedTime > TIMEOUT_VALUE) {
         WSNPRINTF(buf, sizeof(buf), "\nProcess timed out at %d seconds, "
                 "stopping\r", elapsedTime);
-        WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
+        WMEMSET(currentFile, 0, sizeof(currentFile));
         wolfSSH_SFTP_Interrupt(ssh);
     }
 #endif
@@ -667,7 +674,7 @@ static int doCmds(func_args* args)
                     ret == WS_CHAN_RXD || ret == WS_REKEYING);
 
 #ifndef WOLFSSH_NO_TIMESTAMP
-            WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
+            WMEMSET(currentFile, 0, sizeof(currentFile));
 #endif
 
             if (ret != WS_SUCCESS) {
@@ -778,7 +785,7 @@ static int doCmds(func_args* args)
                     ret == WS_CHAN_RXD || ret == WS_REKEYING);
 
 #ifndef WOLFSSH_NO_TIMESTAMP
-            WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
+            WMEMSET(currentFile, 0, sizeof(currentFile));
 #endif
 
             if (ret != WS_SUCCESS) {
@@ -1445,6 +1452,7 @@ static int doAutopilot(int cmd, char* local, char* remote)
     int ret = WS_SUCCESS;
     char fullpath[128] = ".";
     WS_SFTPNAME* name  = NULL;
+    word32 remoteSz;
     byte remoteAbsPath = 0;
 
     /* check if is absolute path before making it one */
@@ -1462,8 +1470,11 @@ static int doAutopilot(int cmd, char* local, char* remote)
 
     if (remoteAbsPath) {
        /* use remote absolute path if provided */
+       remoteSz = (word32)WSTRLEN(remote);
+       if (remoteSz >= sizeof(fullpath))
+           return WS_BUFFER_E;
        WMEMSET(fullpath, 0, sizeof(fullpath));
-       WSTRNCPY(fullpath, remote, sizeof(fullpath) - 1);
+       WMEMCPY(fullpath, remote, remoteSz + 1);
     }
     else {
         err = doBuildRemotePath(remote, fullpath, sizeof(fullpath), &name);
--- a/src/wolfscp.c
+++ b/src/wolfscp.c
@@ -1952,12 +1952,19 @@ int wolfSSH_SCP_connect(WOLFSSH* ssh, byte* cmd)
 }
 
 
+/* Shared format so the size probe and the write cannot drift. */
+#define SCP_CMD_FMT "scp -%c %s"
+
 static char* MakeScpCmd(const char* name, char dir, void* heap)
 {
     char* cmd;
     int sz;
 
-    sz = WSNPRINTF(NULL, 0, "scp -%c %s", dir, name) + 1;
+#ifdef USE_WINDOWS_API
+    sz = WSCPRINTF(SCP_CMD_FMT, dir, name) + 1;
+#else
+    sz = WSNPRINTF(NULL, 0, SCP_CMD_FMT, dir, name) + 1;
+#endif
     if (sz <= 0) {
         return NULL;
     }
@@ -1965,7 +1972,7 @@ static char* MakeScpCmd(const char* name, char dir, void* heap)
     if (cmd == NULL) {
         return NULL;
     }
-    sz = WSNPRINTF(cmd, sz, "scp -%c %s", dir, name);
+    sz = WSNPRINTF(cmd, sz, SCP_CMD_FMT, dir, name);
     if (sz <= 0) {
         WFREE(cmd, heap, DYNTYPE_STRING);
         return NULL;
@@ -2712,9 +2719,14 @@ static ScpDir* ScpNewDir(void *fs, const char* path, void* heap)
 int ScpPushDir(void *fs, ScpSendCtx* ctx, const char* path, void* heap)
 {
     ScpDir* entry;
+    word32  pathSz;
 
     if (ctx == NULL || path == NULL)
         return WS_BAD_ARGUMENT;
+
+    pathSz = (word32)WSTRLEN(path);
+    if (pathSz >= sizeof(ctx->dirName))
+        return WS_BUFFER_E;
 
     entry = ScpNewDir(fs, path, heap);
     if (entry == NULL) {
@@ -2729,12 +2741,27 @@ int ScpPushDir(void *fs, ScpSendCtx* ctx, const char* path, void* heap)
         ctx->currentDir = entry;
     }
 
-    /* append directory name to ctx->dirName */
-    WSTRNCPY(ctx->dirName, path, DEFAULT_SCP_FILE_NAME_SZ-1);
-    ctx->dirName[DEFAULT_SCP_FILE_NAME_SZ-1] = '\0';
+    /* append directory name to ctx->dirName, terminator included; the guard
+     * above bounds pathSz so the copy always fits */
+    WMEMCPY(ctx->dirName, path, pathSz + 1);
 
     return WS_SUCCESS;
 }
+
+#ifdef WOLFSSH_TEST_INTERNAL
+int wolfSSH_TestScpPushDir(const char* path)
+{
+    ScpSendCtx ctx;
+    int ret;
+
+    WMEMSET(&ctx, 0, sizeof(ctx));
+    ret = ScpPushDir(NULL, &ctx, path, NULL);
+    if (ret == WS_SUCCESS)
+        ScpSendCtxFreeDirs(NULL, &ctx, NULL);
+
+    return ret;
+}
+#endif /* WOLFSSH_TEST_INTERNAL */
 
 /* Remove top ScpDir from directory stack, remove dir from ctx->dirName */
 int ScpPopDir(void *fs, ScpSendCtx* ctx, void* heap)
--- a/tests/api.c
+++ b/tests/api.c
@@ -1687,6 +1687,20 @@ static void test_wolfSSH_agent_signrequest_success(void)
 #if defined(WOLFSSH_SFTP) && !defined(NO_WOLFSSH_CLIENT) && \
     !defined(SINGLE_THREADED)
 
+/* A peer gone at shutdown is not a failure: the send path returns the reset,
+ * the recv path returns generic WS_ERROR with the code in wolfSSH_get_error.
+ * WS_SOCKET_ERROR_E is generic, so a real shutdown failure is tolerated too. */
+static int AbsorbBenignReset(WOLFSSH* ssh, int ret)
+{
+    if (ret == WS_SOCKET_ERROR_E ||
+            (ret == WS_ERROR &&
+             wolfSSH_get_error(ssh) == WS_SOCKET_ERROR_E)) {
+        ret = WS_SUCCESS;
+    }
+
+    return ret;
+}
+
 byte userPassword[256];
 
 static int sftpUserAuth(byte authType, WS_UserAuthData* authData, void* ctx)
@@ -1820,10 +1834,8 @@ static void test_wolfSSH_SFTP_SendReadPacket(void)
     argsCount = 0;
     args[argsCount++] = ".";
     args[argsCount++] = "-1";
-#ifndef USE_WINDOWS_API
     args[argsCount++] = "-p";
     args[argsCount++] = "0";
-#endif
     ser.argv   = (char**)args;
     ser.argc    = argsCount;
     ser.signal = &ready;
@@ -2004,11 +2016,7 @@ static void test_wolfSSH_SFTP_SendReadPacket(void)
         wolfSSH_worker(ssh, NULL);
     }
 
-    argsCount = wolfSSH_shutdown(ssh);
-    if (argsCount == WS_SOCKET_ERROR_E) {
-        /* If the socket is closed on shutdown, peer is gone, this is OK. */
-        argsCount = WS_SUCCESS;
-    }
+    argsCount = AbsorbBenignReset(ssh, wolfSSH_shutdown(ssh));
 
 #if DEFAULT_HIGHWATER_MARK < 8000
     if (argsCount == WS_REKEYING) {
@@ -2030,6 +2038,7 @@ static void test_wolfSSH_SFTP_SendReadPacket(void)
     k_sleep(Z_TIMEOUT_TICKS(100));
 #endif
     ThreadJoin(serThread);
+    FreeTcpReady(&ready);
 }
 
 
@@ -2090,10 +2099,8 @@ static void test_wolfSSH_SFTP_PartialSend(void)
     argsCount = 0;
     args[argsCount++] = ".";
     args[argsCount++] = "-1";
-#ifndef USE_WINDOWS_API
     args[argsCount++] = "-p";
     args[argsCount++] = "0";
-#endif
     ser.argv   = (char**)args;
     ser.argc    = argsCount;
     ser.signal = &ready;
@@ -2279,10 +2286,7 @@ static void test_wolfSSH_SFTP_PartialSend(void)
         wolfSSH_worker(ssh, NULL);
     }
 
-    argsCount = wolfSSH_shutdown(ssh);
-    if (argsCount == WS_SOCKET_ERROR_E) {
-        argsCount = WS_SUCCESS;
-    }
+    argsCount = AbsorbBenignReset(ssh, wolfSSH_shutdown(ssh));
 #if DEFAULT_HIGHWATER_MARK < 8000
     if (argsCount == WS_REKEYING) {
         argsCount = WS_SUCCESS;
@@ -2299,6 +2303,7 @@ static void test_wolfSSH_SFTP_PartialSend(void)
     k_sleep(Z_TIMEOUT_TICKS(100));
 #endif
     ThreadJoin(serThread);
+    FreeTcpReady(&ready);
 #endif /* WOLFSSH_TEST_INTERNAL */
 }
 
@@ -2330,10 +2335,8 @@ static void sftp_rekey_test(int nonBlock)
     argsCount = 0;
     args[argsCount++] = ".";
     args[argsCount++] = "-1";
-#ifndef USE_WINDOWS_API
     args[argsCount++] = "-p";
     args[argsCount++] = "0";
-#endif
     ser.argv   = (char**)args;
     ser.argc    = argsCount;
     ser.signal = &ready;
@@ -2417,9 +2420,7 @@ static void sftp_rekey_test(int nonBlock)
      * before the WS_REKEYING fixup below could otherwise mask it. The loop cap
      * is one above the assert threshold to leave last-iteration headroom. */
     AssertIntLE(tries, SFTP_REKEY_MAX_TRIES);
-    if (argsCount == WS_SOCKET_ERROR_E) {
-        argsCount = WS_SUCCESS;
-    }
+    argsCount = AbsorbBenignReset(ssh, argsCount);
 #if DEFAULT_HIGHWATER_MARK < 8000
     if (argsCount == WS_REKEYING) {
         /* in cases where highwater mark is really small a re-key could happen */
@@ -2437,6 +2438,7 @@ static void sftp_rekey_test(int nonBlock)
     k_sleep(Z_TIMEOUT_TICKS(100));
 #endif
     ThreadJoin(serThread);
+    FreeTcpReady(&ready);
 }
 
 static void test_wolfSSH_SFTP_ReKey(void)
@@ -2615,10 +2617,8 @@ static void test_wolfSSH_SFTP_Confinement(void)
     argsCount = 0;
     args[argsCount++] = ".";
     args[argsCount++] = "-1";
-#ifndef USE_WINDOWS_API
     args[argsCount++] = "-p";
     args[argsCount++] = "0";
-#endif
     ser.argv = (char**)args;
     ser.argc = argsCount;
     ser.signal = &ready;
@@ -2791,10 +2791,7 @@ static void test_wolfSSH_SFTP_Confinement(void)
     while (wolfSSH_get_error(ssh) == WS_REKEYING)
         wolfSSH_worker(ssh, NULL);
 
-    ret = wolfSSH_shutdown(ssh);
-    if (ret == WS_SOCKET_ERROR_E) {
-        ret = WS_SUCCESS;
-    }
+    ret = AbsorbBenignReset(ssh, wolfSSH_shutdown(ssh));
 #if DEFAULT_HIGHWATER_MARK < 8000
     if (ret == WS_REKEYING) {
         ret = WS_SUCCESS;
@@ -2809,6 +2806,7 @@ static void test_wolfSSH_SFTP_Confinement(void)
     k_sleep(Z_TIMEOUT_TICKS(100));
 #endif
     ThreadJoin(serThread);
+    FreeTcpReady(&ready);
 
     /* remove staged targets; escMkdir/escDest only exist if confinement
      * leaked, and inJailDir only if the positive-case RMDIR did not run, so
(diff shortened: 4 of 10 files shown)


src/wolfscp.c
```diff
@@ -1904,7 +1904,11 @@ static char* MakeScpCmd(const char* name, char dir, void* heap)
     char* cmd;
     int sz;
 
+#ifdef USE_WINDOWS_API
+    sz = WSCPRINTF("scp -%c %s", dir, name) + 1;
```
aidangarske: **🟠 [High] New WSCPRINTF macro breaks Windows WSTRING_USER builds**
*🚫 BLOCK*

The PR changes the Windows length probe from `WSNPRINTF(NULL, 0, ...)` to `WSCPRINTF(...)`, but `WSCPRINTF` is only defined inside `#ifndef WSTRING_USER` in `wolfssh/port.h`. A supported Windows build that defines `WSTRING_USER` and enables SCP will now compile `MakeScpCmd` with `USE_WINDOWS_API` and hit an undefined `WSCPRINTF` macro. This is introduced by the PR because `MakeScpCmd` previously only depended on the existing `WSNPRINTF` wrapper.

**Recommendation:** Provide a `WSCPRINTF` fallback outside the `#ifndef WSTRING_USER` block, or avoid adding a new required string-port macro by using a local Windows size-probe helper that is always available for `USE_WINDOWS_API` builds. Fix the macro availability before merge so custom string-port Windows builds do not regress.

aidangarske (dismissed):
## 🐺 Skoll Code Review

### Posted findings
- [High] **New WSCPRINTF macro breaks Windows WSTRING_USER builds** — `src/wolfscp.c:1908`
- [Medium] **Boundary checks reject one valid maximum-length path** — `src/wolfscp.c:2594; examples/sftpclient/sftpclient.c:1467`

*Review generated by Skoll.*

src/wolfscp.c
aidangarske: **🟠 [High] New WSCPRINTF macro breaks Windows WSTRING_USER builds**
*🚫 BLOCK*

The PR changes the Windows length probe from `WSNPRINTF(NULL, 0, ...)` to `WSCPRINTF(...)`, but `WSCPRINTF` is only defined inside `#ifndef WSTRING_USER` in `wolfssh/port.h`. A supported Windows build that defines `WSTRING_USER` and enables SCP will now compile `MakeScpCmd` with `USE_WINDOWS_API` and hit an undefined `WSCPRINTF` macro. This is introduced by the PR because `MakeScpCmd` previously only depended on the existing `WSNPRINTF` wrapper.

**Recommendation:** Provide a `WSCPRINTF` fallback outside the `#ifndef WSTRING_USER` block, or avoid adding a new required string-port macro by using a local Windows size-probe helper that is always available for `USE_WINDOWS_API` builds. Fix the macro availability before merge so custom string-port Windows builds do not regress.

src/wolfscp.c
```diff
@@ -2587,6 +2591,9 @@ int ScpPushDir(void *fs, ScpSendCtx* ctx, const char* path, void* heap)
     if (ctx == NULL || path == NULL)
         return WS_BAD_ARGUMENT;
 
+    if (WSTRLEN(path) >= DEFAULT_SCP_FILE_NAME_SZ - 1)
```
aidangarske: **🟡 [Medium] Boundary checks reject one valid maximum-length path**
*💡 SUGGEST*

The new `WSTRNCPY` guards compare against `buffer_size - 1` and then pass `buffer_size - 1` as the Windows secure-CRT destination size. That avoids the `strncpy_s` fastfail, but it also rejects strings that exactly fill the destination with one byte left for the terminator. For example, `ctx->dirName` is `DEFAULT_SCP_FILE_NAME_SZ` bytes, so a `DEFAULT_SCP_FILE_NAME_SZ - 1` byte path should fit, but the new guard returns `WS_BUFFER_E`. The same off-by-one behavior exists for absolute autopilot paths in `fullpath`.

**Recommendation:** Use the real destination size in the `WSTRNCPY` call and guard against `>= sizeof(destination)`, then add a boundary test for an input whose length is exactly `sizeof(destination) - 1`. Adjust the copy sizes so the Windows hardening does not reduce the accepted maximum path length.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **New WSCPRINTF macro breaks Windows WSTRING_USER builds** — `src/wolfscp.c:1908`
- [Medium] **Boundary checks reject one valid maximum-length path** — `src/wolfscp.c:2594; examples/sftpclient/sftpclient.c:1467`

*Review generated by Skoll.*

src/wolfscp.c
yosuke-wolfssl: Thanks for reviewing.
Confirmed. `WSCPRINTF` was defined at `port.h:595`, inside the `#ifndef WSTRING_USER` block that closes at line 637. Since `WSTRING_USER` means the integrator supplies every string macro themselves, a `WSTRING_USER` + `USE_WINDOWS_API` + `WOLFSSH_SCP` build would hit an undefined `WSCPRINTF` — and would have no way to know it needed to define it, because before this PR `MakeScpCmd` only depended on the already-required `WSNPRINTF`. A genuine build regression introduced here.

Moved the definition out of the block, behind its own guard so it is still overridable.

yosuke-wolfssl: It's duplicated and fixed on above comment

src/wolfscp.c
yosuke-wolfssl: Confirmed, and it's a behavior regression on POSIX too, not only Windows. Both guards compared against `size - 1` while the destination is explicitly NUL-terminated, so they rejected a string that previously fit exactly.
Both sites now pass the true destination size to `WSTRNCPY` and guard on `>= sizeof(dest)`.

yosuke-wolfssl:
Hello @aidangarske ,

tests/api.c
```diff
@@ -1502,6 +1502,20 @@ static void test_wolfSSH_agent_signrequest_success(void)
 #if defined(WOLFSSH_SFTP) && !defined(NO_WOLFSSH_CLIENT) && \
     !defined(SINGLE_THREADED)
 
+/* A peer that is already gone at shutdown is not a test failure. The send path
+ * reports the reset as the return value; the recv path returns the generic
+ * WS_ERROR and stashes the specific code in wolfSSH_get_error. */
+static int AbsorbBenignReset(WOLFSSH* ssh, int ret)
```
aidangarske, line 1693: **🟠 [Medium] AbsorbBenignReset widens shutdown tolerance beyond peer-reset** · _Test Quality_

Verified against the tree. The helper absorbs `ret == WS_ERROR && wolfSSH_get_error(ssh) == WS_SOCKET_ERROR_E` in addition to the direct `WS_SOCKET_ERROR_E` return the old code handled. Confirmed that `WS_ERROR` and `WS_FATAL_ERROR` are the same value (-1001, wolfssh/error.h:42-43), so the `ret == WS_ERROR` gate matches every generic wolfSSH failure return, not just the recv path. `WS_SOCKET_ERROR_E` (-1009) is wolfSSH's generic socket-layer failure code — it is not specific to "peer sent RST". So a server that dies mid-shutdown for a genuine reason (rather than the benign post-transfer race this targets) now also reports WS_SUCCESS across all five call sites. This is a deliberate and reasonable trade (the old code already absorbed the send-path form, and the PR documents the ~30-40% flake it fixes), but the tests can no longer distinguish a benign reset from a real socket failure at shutdown, and the widening applies uniformly to SendReadPacket / PartialSend / rekey / Confinement / KeyboardInteractive.

**Fix:** Consider gating on the underlying `WSAGetLastError() == WSAECONNRESET` / `errno == ECONNRESET` so a genuine socket failure at shutdown still fails the test. If that proves too brittle across the platforms api-test runs on, keep the current form but note the limitation in the helper comment.

src/wolfscp.c
```diff
@@ -2663,10 +2667,15 @@ static ScpDir* ScpNewDir(void *fs, const char* path, void* heap)
 int ScpPushDir(void *fs, ScpSendCtx* ctx, const char* path, void* heap)
 {
     ScpDir* entry;
+    word32  pathSz;
 
     if (ctx == NULL || path == NULL)
         return WS_BAD_ARGUMENT;
 
+    pathSz = (word32)WSTRLEN(path);
```
aidangarske, line 2727: **🟠 [Medium] No test coverage for the new input-validation reject paths** · _Test Coverage_

Verified against the tree. The PR adds four new bounds rejects — `ScpPushDir` -> `WS_BUFFER_E` (wolfscp.c:2676), `doAutopilot` -> `WS_BUFFER_E` (sftpclient.c:1471), the three echoserver `LoadPasswdList`/`LoadKeyboardList`/`LoadPubKeyList` skip-and-continue guards (echoserver.c:2263, 2297, 2338), and the `myStatusCb` name guard — and none of them gain a test. The stated verification is that the pre-existing Windows suite stops crashing, which exercises only the under-limit path in every case; the new over-limit branches are entirely uncovered on both POSIX and Windows. `ScpPushDir` is the one worth covering: it is a genuine behavior change (was silently truncating at `DEFAULT_SCP_FILE_NAME_SZ - 1` and returning WS_SUCCESS, now rejects at >= 1024 — confirmed `dirName[DEFAULT_SCP_FILE_NAME_SZ]` at wolfscp.h:95 and the 1024 define at wolfscp.h:47), and its three callers map the failure to three different outcomes — `WS_INVALID_PATH_E`/`WS_FATAL_ERROR` (wolfscp.c:1383), `WS_SCP_ABORT` (wolfscp.c:2965), and a bare log (wolfscp.c:3243). Confirmed `ScpPushDir` is forward-declared `static` at wolfscp.c:56 (the definition at line 2667 omits `static`, but the prior internal-linkage declaration is visible, so linkage stays internal) — it cannot be unit-tested directly, and coverage would need an integration case driving a >1023-char base path through `wolfSSH_SCP_to`.

**Fix:** At minimum cover the `ScpPushDir` over-long-path reject, since it changes library behavior for all three callers. The example-program guards (echoserver/sftpclient) are lower value and can reasonably be left uncovered.

examples/sftpclient/sftpclient.c
```diff
@@ -181,8 +182,11 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
...
         lastPrintedBytes[0] = 0;
         lastPrintedBytes[1] = 0;
-        WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
-        WSTRNCPY(currentFile, name, WOLFSSH_MAX_FILENAME);
+        WMEMSET(currentFile, 0, sizeof(currentFile));
+        nameSz = (word32)WSTRLEN(name);
+        if (nameSz >= sizeof(currentFile))
+            return;
+        WMEMCPY(currentFile, name, nameSz + 1);
     }
```
aidangarske, line 193: **🔵 [Low] myStatusCb early return permanently suppresses all progress output for over-long file names** · _Logic_

All three scan modes independently confirmed this against the tree. The new length guard sits AFTER the state reset, so when `nameSz >= sizeof(currentFile)` (257, since `currentFile` is `static char currentFile[WOLFSSH_MAX_FILENAME + 1]` at sftpclient.c:104 and WOLFSSH_MAX_FILENAME is 256 per wolfssh/ssh.h:572) the callback has already zeroed `currentFile` and reset `startTime`/`lastOutputTime`/`lastPrintedBytes` before it bails. Because `currentFile` is left empty, the next call's `WSTRNCMP(currentFile, name, WSTRLEN(name))` still mismatches, so every subsequent call re-enters the reset branch and returns again — progress reporting is permanently suppressed for the entire transfer, and the `WSNPRINTF`/`WFPUTS` print at lines 192-207 is never reached. Previously on POSIX `WSTRNCPY(currentFile, name, WOLFSSH_MAX_FILENAME)` truncated the name into the buffer and fell through to print, so this is a regression on POSIX. On Windows the old `WSTRNCPY` expanded to `strncpy_s(currentFile, 256, name, 256)`, a constraint violation that `__fastfail`s for `strlen(name) >= 256` — the guard genuinely fixes that crash, so the direction is right; the issue is that it trades a crash for silence rather than clamping. The guard itself is buffer-correct: `nameSz == 256` copies 257 bytes and fits exactly, no off-by-one. Triggering input is ordinary: `statusCb` receives the caller-supplied local path from `wolfSSH_SFTP_Get`/`wolfSSH_SFTP_Put`, and POSIX PATH_MAX is 4096. Severity views differed — the `review` mode rated this Medium/SUGGEST, `review-security` and `bugs` rated it Low; the…

**Fix:** Clamp `nameSz` to `sizeof(currentFile) - 1` and fall through instead of returning. This keeps the strncpy_s-safe property the PR is after, populates `currentFile` so the name-change branch stops re-triggering, and preserves the pre-existing truncate-and-print behavior on POSIX. It does not address the pre-existing dead timeout for long names — to fix that too, compare over the stored length (e.g. `WSTRNCMP(currentFile, name, WSTRLEN(currentFile))`) or track `nameSz` in a static alongside `currentFile`.

wolfssh/test.h
```diff
@@ -891,6 +917,11 @@ static INLINE void WaitTcpReady(tcp_ready* ready)
      * seems to help. This is not ideal. (XXX) */
     k_sleep(Z_TIMEOUT_TICKS(300));
 #endif /* WOLFSSH_ZEPHYR */
+#elif defined(USE_WINDOWS_API) && defined(NO_MAIN_DRIVER) && \
```
aidangarske, line 927: **🔵 [Low] WaitTcpReady Win32 path discards WaitForSingleObject result, turning a server-startup failure into a 30s stall plus a…** · _Error Handling_

All three scan modes independently confirmed this against the tree (TCP_READY_TIMEOUT_MS is 30000 at wolfssh/test.h:860). The Win32 branch calls `WaitForSingleObject(ready->readyEvent, TCP_READY_TIMEOUT_MS)` and discards the DWORD result, and unlike the POSIX branch directly above (test.h:905-913, `while (!ready->ready) pthread_cond_wait(...)`) it never inspects the `ready->ready` predicate — `WAIT_OBJECT_0`, `WAIT_TIMEOUT` and `WAIT_FAILED` are indistinguishable to the caller. Triggering condition: `SignalTcpReady` is invoked from exactly one site (echoserver.c:3469, immediately after `tcp_listen()`), so every `ES_ERROR` path before that line expands to `WOLFSSL_RETURN_FROM_THREAD(0)` and exits the server thread WITHOUT signalling the manual-reset event created at test.h:875. Reachable examples on the Windows api-test build (`ide/winvs/api-test/api-test.vcxproj` defines `NO_MAIN_DRIVER` project-wide, so the whole Win32 handshake is live): `-p` with a NULL `myoptarg`, a host-key load failure, or `-R` under `NO_FILESYSTEM`. In all of those `WaitTcpReady` stalls the full 30s then returns exactly as if the server were up; `ready.port` is still 0 (set in `InitTcpReady`, test.h:866), the client connects to 127.0.0.1:0, `connect()` fails, and `AssertNotNull(ctx)` from `sftp_client_connect(&ctx, &ssh, 0)` aborts. The in-code claim that the caller "fails fast" is only half true: the run aborts rather than hangs, but only after burning 30s per affected test (up to six call sites in tests/api.c -> ~3 minutes), and the operator sees `ctx => NULL` — a connect failure pointing at the…

**Fix:** Check for `WAIT_OBJECT_0` and the `ready->ready` predicate, and call `err_sys()` otherwise. `err_sys` is `WS_NORETURN`/`exit()` on Windows (test.h:299-315) and is the established convention in this header — it already handles `CreateEvent` failure the same way at test.h:876-879 — so this keeps the fail-fast intent, makes the cause visible at the point it happens, avoids the racy `ready.port` read, and realigns the Win32 branch with the POSIX predicate loop. If aborting inside the harness is undesirable, at minimum print the wait result and drop the 'fails fast' wording from the comment, since the observed behaviour is a 30s stall followed by an unrelated-looking assertion.

wolfssh/test.h
```diff
@@ -826,6 +826,9 @@ typedef struct tcp_ready {
 #if defined(_POSIX_THREADS) && !defined(__MINGW32__)
     pthread_mutex_t mutex;
     pthread_cond_t  cond;
+#elif defined(USE_WINDOWS_API)
```
aidangarske, line 840: **🔵 [Low] CRITICAL_SECTION in tcp_ready is never acquired by any reader** · _Dead Code_

Verified against the tree; the `bugs` mode independently noted the same thing as a related observation on its WaitTcpReady finding ("`ready->ready` is declared as the predicate (test.h:823) and is written under `cs` by the signalling side, but is never read by `WaitTcpReady`, so the predicate the `CRITICAL_SECTION` protects has no reader on this path"). The new Win32 `tcp_ready` carries both a `CRITICAL_SECTION cs` and a manual-reset `readyEvent`, but only the writer (`SignalTcpReady` in echoserver.c:2861-2865) ever enters the CS. `WaitTcpReady` waits solely on the event (test.h:924) and never enters `cs` nor reads the `ready->ready` predicate; callers read `ready.port` directly with no lock. `SetEvent`/`WaitForSingleObject` are already full barriers, so they alone order the `port` write ahead of the reader's load — the CS protects nothing and adds `InitializeCriticalSection`/`DeleteCriticalSection` lifecycle (test.h:874-877, 896) to every call site. This is a faithful transliteration of the POSIX mutex+cond shape, but a Win32 event needs no companion lock for a one-shot signal. Note the event handle still needs `FreeTcpReady`, so dropping `cs` does not remove the Init/Free pairing — it just removes dead state. Also flagging the asymmetry: the struct members are guarded by `#elif defined(USE_WINDOWS_API)` alone while every function that touches them additionally requires `NO_MAIN_DRIVER && !SINGLE_THREADED`. This is harmless (the members are simply unused in those builds) and mirrors the pre-existing POSIX asymmetry, but tightening the struct guard to match would prevent a…

**Fix:** Drop `cs` and the `EnterCriticalSection`/`LeaveCriticalSection` pair in `SignalTcpReady`; `SetEvent` already publishes the `port` write. If the CS is kept deliberately, have `WaitTcpReady` actually take it and check `ready->ready` so the lock has a reader — which would also fix the predicate gap in the WaitTcpReady finding above.

src/wolfscp.c
```diff
@@ -1908,7 +1908,11 @@ static char* MakeScpCmd(const char* name, char dir, void* heap)
     char* cmd;
     int sz;
 
+#ifdef USE_WINDOWS_API
```
aidangarske, line 1963: **🔵 [Low] MakeScpCmd now repeats the "scp -%c %s" format string three times** · _Maintainability_

Verified against the tree. Adding the Windows `WSCPRINTF` branch brings the format-string count in this one function to three: the Windows length probe, the POSIX length probe, and the actual write. The probe result sizes the `WMALLOC`, so any future divergence between a probe copy and the write copy silently becomes either a heap overflow or a truncated command. The two probes are also mutually exclusive at compile time, so no build ever cross-checks them against each other. The current logic is correct — `_scprintf` returns chars-excluding-NUL (-1 on error, caught by `sz <= 0`), and the subsequent `_snprintf_s(cmd, sz, _TRUNCATE, ...)` writes at most `sz-1` chars plus NUL, which fits exactly — but the duplication is the fragile part.

**Fix:** Hoist the format into a file-local `#define` so the probe and the write cannot drift. Use a macro rather than a `const char*` local — MSVC emits C4774 for non-literal format strings passed to the CRT `_s`/`_scprintf` family.

examples/echoserver/echoserver.c
```diff
@@ -2260,6 +2260,11 @@ static int LoadPasswdList(StrList* strList, PwMapList* mapList)
     int count = 0;
 
     while (strList) {
+        if (WSTRLEN(strList->str) >= sizeof names - 1) {
```
aidangarske, line 2252: **🔵 [Low] Over-long entry guard copy-pasted three times with a non-identifying message** · _Maintainability_

Verified against the tree — the same five-line guard appears verbatim at lines 2263, 2297 and 2338. The bound is correct: `names[256]`, guard rejects `len >= 255`, and `WSTRNCPY(names, strList->str, sizeof names - 1)` maps to `strncpy_s(names, 255, src, 255)`, which `__fastfail`s exactly when `strlen(src) >= 255`, so the guard covers precisely the crashing set. Two small polish points: the message `"Ignoring over-long entry\n"` does not say which entry, unlike the adjacent `fprintf(stderr, "Ignoring password: %s\n", names)` a few lines down; and the guard's `sizeof names - 1` must stay in lockstep with the `WSTRNCPY` count argument or the protection silently lapses. (The skipped entries also no longer bump `count`, but every caller at lines 3204-3249 discards the return value, so that is a non-issue.)

**Fix:** Print a truncated prefix of the offending entry so the operator can identify it. Optionally factor the guard into a small `CopyListEntry(names, sizeof names, strList->str)` helper shared by the three loaders.

examples/sftpclient/sftpclient.c
```diff
@@ -181,8 +182,11 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
...
     elapsedTime = currentTime - startTime;
     WSNPRINTF(buf, sizeof(buf), "Processed %8llu\t bytes in %d seconds\r",
```
aidangarske, line 195: **🔵 [Low] Leftover WOLFSSH_MAX_FILENAME memsets inconsistent with new sizeof(currentFile)** · _Consistency_

Verified against the tree. Line 185 was updated to `WMEMSET(currentFile, 0, sizeof(currentFile))`, but lines 198, 674 and 785 still use `WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME)`. `currentFile` is declared `char currentFile[WOLFSSH_MAX_FILENAME + 1]` (line 102), so those three leave the final byte untouched. This is harmless today — after the new `WMEMCPY(currentFile, name, nameSz + 1)` with `nameSz == 256`, byte [256] is the NUL, so a 256-byte memset still yields an empty string — but the mismatch is precisely the off-by-one class this PR set out to eliminate, and it now reads as an oversight next to the updated line 185.

**Fix:** Switch lines 198, 674 and 785 to `sizeof(currentFile)` for consistency with line 185.

---
> **Note:** Referenced line (`examples/sftpclient/sftpclient.c:198`) is outside the diff hunk. Comment anchored to nearest changed region.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 8 total — 8 posted, 0 skipped
**8 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] [review] **AbsorbBenignReset widens shutdown tolerance beyond peer-reset** — `tests/api.c:1508-1517`
- [Medium] [review] **No test coverage for the new input-validation reject paths** — `src/wolfscp.c:2675-2677`
- [Low] [review+review-security+bugs] **myStatusCb early return permanently suppresses all progress output for over-long file names** — `examples/sftpclient/sftpclient.c:180-190`
- [Low] [review+review-security+bugs] **WaitTcpReady Win32 path discards WaitForSingleObject result, turning a server-startup failure into a 30s stall plus a misleading connect failure** — `wolfssh/test.h:920-924`
- [Low] [review+bugs] **CRITICAL_SECTION in tcp_ready is never acquired by any reader** — `wolfssh/test.h:829-831`
- [Low] [review] **MakeScpCmd now repeats the "scp -%c %s" format string three times** — `src/wolfscp.c:1911-1923`
- [Low] [review] **Leftover WOLFSSH_MAX_FILENAME memsets inconsistent with new sizeof(currentFile)** — `examples/sftpclient/sftpclient.c:198`
- [Low] [review] **Over-long entry guard copy-pasted three times with a non-identifying message** — `examples/echoserver/echoserver.c:2263-2267`

examples/sftpclient/sftpclient.c
yosuke-wolfssl: Fixed

wolfssh/test.h
yosuke-wolfssl: Fixed

wolfssh/test.h
yosuke-wolfssl: Fixed

src/wolfscp.c
yosuke-wolfssl: Fixed

examples/echoserver/echoserver.c
yosuke-wolfssl: Fixed

examples/sftpclient/sftpclient.c
yosuke-wolfssl: Fixed

tests/api.c
yosuke-wolfssl: Confirmed the widening is real — WS_ERROR/WS_FATAL_ERROR are both -1001 and
WS_SOCKET_ERROR_E is wolfSSH's generic socket code, so this does absorb any
generic socket failure at shutdown across all five call sites, not just a peer
RST.

I kept the current form rather than gating on WSAGetLastError()==WSAECONNRESET /
errno==ECONNRESET. That gating is brittle here: AbsorbBenignReset runs on the
value wolfSSH_shutdown() already returned, and wolfSSH does its own socket
cleanup on the way out, so the OS last-error/errno is not guaranteed to still
hold the reset code by the time the helper reads it — exactly the cross-platform
brittleness the finding notes. Taking the documented-limitation option instead:
the helper now carries a comment stating the tolerance is generic-socket-wide
and is acceptable because it only applies after the transfer has completed and
shutdown is the final step.

yosuke-wolfssl:
Hello @aidangarske ,
I fixed the issues and updated comments. Can you check it please ?

examples/sftpclient/sftpclient.c
```diff
@@ -181,8 +182,13 @@ static void myStatusCb(WOLFSSH* sshIn, word32* bytes, char* name)
...
         lastPrintedBytes[1] = 0;
-        WMEMSET(currentFile, 0, WOLFSSH_MAX_FILENAME);
-        WSTRNCPY(currentFile, name, WOLFSSH_MAX_FILENAME);
+        WMEMSET(currentFile, 0, sizeof(currentFile));
+        nameSz = (word32)WSTRLEN(name);
+        /* clamp over-long names: returning would leave currentFile empty
+         * and suppress progress output for the rest of the transfer */
+        if (nameSz >= sizeof(currentFile))
+            nameSz = (word32)sizeof(currentFile) - 1;
+        WMEMCPY(currentFile, name, nameSz);
```
aidangarske, line 192: **🟠 [Medium] Over-long SFTP status names reset progress state every callback** · _Logic Errors_

The PR replaces the Windows-crashing WSTRNCPY with a manual clamp/copy of over-long `name` values into `currentFile`, but the file-change detection still compares using the full `WSTRLEN(name)`. When `name` is at least `sizeof(currentFile)` bytes, `currentFile` stores only the truncated prefix plus NUL; on the next progress callback `WSTRNCMP(currentFile, name, WSTRLEN(name))` reads past the stored prefix (NUL vs. the remaining source byte) and treats the same transfer as a new file. That resets `startTime` and `lastOutputTime` on every callback, so the progress timeout never reaches TIMEOUT_VALUE and output throttling is bypassed for that transfer. Severity views differ across modes — the review mode rated this Medium and the bugs mode rated it Low; the stricter Medium is retained here. Confidence is High given both modes independently identified the same defect.

**Fix:** Compute the bounded name length before the comparison and use that same bounded length for both comparison and copy, e.g.:

nameSz = (word32)WSTRLEN(name);
if (nameSz >= sizeof(currentFile))
    nameSz = (word32)sizeof(currentFile) - 1;

if (WSTRNCMP(currentFile, name, nameSz) != 0 ||
        currentFile[nameSz] != '\0') {
    startTime = current_time(1);
    lastOutputTime = 0;
    lastPrintedBytes[0] = 0;
    lastPrintedBytes[1] = 0;
    WMEMSET(currentFile, 0, sizeof(currentFile));
    WMEMCPY(currentFile, name, nameSz);
}

This keeps the clamped filename stable across repeated callbacks for the same transfer.

src/wolfscp.c
```diff
@@ -2712,10 +2719,15 @@ static ScpDir* ScpNewDir(void *fs, const char* path, void* heap)
 int ScpPushDir(void *fs, ScpSendCtx* ctx, const char* path, void* heap)
 {
     ScpDir* entry;
+    word32  pathSz;
 
     if (ctx == NULL || path == NULL)
         return WS_BAD_ARGUMENT;
 
+    pathSz = (word32)WSTRLEN(path);
```
aidangarske, line 2727: **🟠 [Medium] Boundary paths added for Windows string hardening are not covered** · _Test Coverage_

This PR converts several Windows crash-prone WSTRNCPY call sites into explicit length checks or manual copies, but the enabled integration tests exercise only normal SFTP/SCP transfers. There is no coverage for the new boundary behavior: ScpPushDir accepting `sizeof(ctx->dirName) - 1` bytes and rejecting `sizeof(ctx->dirName)` with WS_BUFFER_E, the echoserver over-long auth-list entries being skipped, or the sftpclient absolute autopilot path returning WS_BUFFER_E. Since the PR specifically targets Windows-only boundary failures, these branches can regress without CI catching it.

**Fix:** Add focused boundary tests for at least the core SCP path-length guard and one Windows WSTRNCPY replacement path: exact-fit succeeds, one-byte-too-long returns/skips cleanly, and no invalid-parameter abort occurs on Windows.

wolfssh/test.h
```diff
@@ -860,6 +862,14 @@ typedef struct func_args {
...
 
+#if defined(USE_WINDOWS_API) && defined(NO_MAIN_DRIVER) && \
+    !defined(SINGLE_THREADED)
+/* Upper bound on the wait for the server to become ready. The signal fires
+ * right after the listen() so the real wait is sub-second; this generous
+ * bound only keeps a dead server from hanging the run. */
+#define TCP_READY_TIMEOUT_MS 30000
+#endif
+
 static INLINE void InitTcpReady(tcp_ready* ready)
```
aidangarske, line 873: **🔵 [Low] Windows tcp_ready event can leak in existing normal-return callers** · _Resource Leak_

Before this PR, InitTcpReady allocated no Windows resource, so callers that omitted FreeTcpReady on Windows did not leak a handle. After the PR, InitTcpReady creates a manual-reset event under USE_WINDOWS_API && NO_MAIN_DRIVER && !SINGLE_THREADED, and FreeTcpReady is now required to close it. The PR adds cleanup to the changed tests/api.c paths, but existing normal-return callers such as tests/auth.c:test_client call InitTcpReady, join the server thread, and return without FreeTcpReady. Because the Windows Visual Studio test projects define NO_MAIN_DRIVER, this leaks one event handle per successful test_client invocation on Windows. Test-only and bounded, so severity is Low, but the shared helper change introduced a new cleanup contract without updating all normal callers.

**Fix:** Update all normal-return InitTcpReady callers to call FreeTcpReady, including tests/auth.c:test_client, or change the helper pattern so Windows allocation and cleanup are harder to unpair.

aidangarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+bugs] **Over-long SFTP status names reset progress state every callback** — `examples/sftpclient/sftpclient.c:180-191`
- [Medium] [review] **Boundary paths added for Windows string hardening are not covered** — `src/wolfscp.c:2727-2746`
- [Low] [review-security] **Windows tcp_ready event can leak in existing normal-return callers** — `wolfssh/test.h:873-904`

examples/sftpclient/sftpclient.c
yosuke-wolfssl: Confirmed. The clamp landed after the comparison, so `currentFile` held the
truncated prefix while `WSTRNCMP` still compared over the full `WSTRLEN(name)` —
the compare could never match, so every callback re-entered the reset branch and
`elapsedTime` never grew, leaving the `TIMEOUT_VALUE` check dead.

Fixed as suggested: `nameSz` is clamped before the comparison and both the
compare and the copy use that bound (sftpclient.c:182-190). Added the
`currentFile[nameSz] != '\0'` check so a longer name sharing a prefix still
registers as a new file.

src/wolfscp.c
yosuke-wolfssl: Confirmed, and covered — but not by the suggested route.

An integration case through `wolfSSH_SCP_to` does not reach this guard on a
default POSIX build: `wsScpSendCallback` already rejects an over-long recursive
root at wolfscp.c:3273 (`rootLen >= DEFAULT_SCP_FILE_NAME_SZ` -> `WS_SCP_ABORT`)
before `ScpPushDir` is called, whenever `WOLFSSH_HAVE_SYMLINK` is defined. Such a
test would pass on the upstream guard and prove nothing about this one. The
`ScpProcessEntry` caller needs a real >1023-char directory, which is at
`PATH_MAX` on macOS.

Used the repo's existing `WOLFSSH_TEST_INTERNAL` shim pattern instead, as with
`wolfSSH_TestScpGetFileMode` — `ScpPushDir` keeps internal linkage and the shim
only exists in `libwolfssh_test.la`, so no ABI change. `wolfSSH_TestScpPushDir()`
at wolfscp.c:2752, declared at internal.h:1537, driven by `test_ScpPushDir()` at
unit.c:8713: 1024 -> `WS_BUFFER_E`, 1023 -> asserts the guard did not fire, NULL
-> `WS_BAD_ARGUMENT`.

The exact-fit case asserts "not `WS_BUFFER_E`" rather than `WS_SUCCESS`, since no
1023-char directory exists to open — it still pins the boundary at 1024.
Confirmed non-vacuous: deleting the guard fails the test, restoring it passes.

Left the example-program guards uncovered, per your note that they're lower value.

src/wolfscp.c
yosuke-wolfssl: Same function as the earlier test-coverage thread — answered in full there.

Short version: the SCP path-length guard is now covered by `test_ScpPushDir()`
(unit.c:8713) via a `WOLFSSH_TEST_INTERNAL` shim. The integration route both
threads suggested cannot reach the guard, because `wsScpSendCallback` rejects the
over-long recursive root upstream at wolfscp.c:3273. The echoserver and
sftpclient guards are example-program code and stay uncovered.

wolfssh/test.h
yosuke-wolfssl: Confirmed. `InitTcpReady` allocating a Windows event made `FreeTcpReady`
mandatory, and `test_client` joined the server thread and returned without it —
one handle leaked per invocation on builds defining `NO_MAIN_DRIVER`.

Added `FreeTcpReady(&ready)` at auth.c:1839. Swept the repo for the same pattern:
`api.c`, `kex.c`, `sftp.c`, `testsuite.c`, the other three `auth.c` sites and the
examples all pair correctly — `test_client` was the only one missing.

aidangarske (approved):
Nice job!

ejohnstown (approved):

PR #1059, opened 2026-06-24
Harden remote port forwarding (tcpip-forward)

ejohnstown:
- Report allocated port in tcpip-forward reply (F-5573): Answer a port-0 (dynamic) tcpip-forward with the actually-bound port per RFC 4254 §7.1; callback returns >= WS_FWD_PORT_CHECK (1024) as the port, below it as a WS_FwdCbError status. A success that reports no port is undone and rejected (mapped to WS_RESOURCE_E so a no-reply request keeps the link).
- Gate forwarded-tcpip opens like direct-tcpip (F-6275): Both forwarding channel types now require a fwdCb and fail closed without one; a forwarded-tcpip open reaching a server is rejected as wrong-direction before any policy hook runs.

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -529,6 +529,8 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
     else if (action == WOLFSSH_FWD_REMOTE_SETUP) {
         struct sockaddr_in addr;
         socklen_t addrSz = 0;
+        socklen_t boundSz = sizeof(addr);
+        word32 allocatedPort = 0;
 
         fwdCbCtx->hostName = WSTRDUP(name, NULL, 0);
         fwdCbCtx->hostPort = port;
@@ -553,7 +555,7 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
             }
             else {
                 printf("Not using IPv6 yet.\n");
-                ret = WS_FWD_SETUP_E;
+                ret = -1;
             }
         }
 
@@ -566,8 +568,35 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
             ret = listen(appCtx->listenFd, 5);
         }
 
+        if (ret == 0 && port == 0) {
+            /* The peer requested port 0, so the OS picked the port during
+             * bind(). Recover it to report back to the caller. */
+            WMEMSET(&addr, 0, sizeof addr);
+            if (getsockname(appCtx->listenFd,
+                    (struct sockaddr*)&addr, &boundSz) == 0) {
+                allocatedPort = (word32)ntohs(addr.sin_port);
+                /* The library reads a return below WS_FWD_PORT_CHECK as a
+                 * status, not a port, so an allocated port must be reportable.
+                 * An unprivileged OS-chosen port always is; guard anyway. */
+                if (allocatedPort < WS_FWD_PORT_CHECK) {
+                    printf("Allocated port %u not reportable.\n", allocatedPort);
+                    ret = -1;
+                }
+                else {
+                    fwdCbCtx->hostPort = allocatedPort;
+                }
+            }
+            else {
+                printf("getsockname failed for forwarded port.\n");
+                ret = -1;
+            }
+        }
+
         if (ret == 0) {
             appCtx->state = APP_STATE_LISTEN;
+            /* Report any dynamically allocated port to the library through the
+             * return value; 0 keeps the port the peer requested. */
+            ret = (int)allocatedPort;
         }
         else {
             if (fwdCbCtx->hostName != NULL) {
--- a/ide/Espressif/ESP-IDF/examples/wolfssh_echoserver/main/echoserver.c
+++ b/ide/Espressif/ESP-IDF/examples/wolfssh_echoserver/main/echoserver.c
@@ -521,6 +521,8 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
     else if (action == WOLFSSH_FWD_REMOTE_SETUP) {
         struct sockaddr_in addr;
         socklen_t addrSz = 0;
+        socklen_t boundSz = sizeof(addr);
+        word32 allocatedPort = 0;
 
         ctx->hostName = WSTRDUP(name, NULL, 0);
         ctx->hostPort = port;
@@ -545,7 +547,7 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
             }
             else {
                 printf("Not using IPv6 yet.\n");
-                ret = WS_FWD_SETUP_E;
+                ret = -1;
             }
         }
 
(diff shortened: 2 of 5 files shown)


src/internal.c
```diff
@@ -9235,11 +9270,23 @@ static int DoChannelOpen(WOLFSSH* ssh,
...
+                /* A forwarded-tcpip open is a server-to-client message sent in
+                 * response to a tcpip-forward request, so only a client should
+                 * ever receive one. Reject opens arriving in the wrong
+                 * direction before applying any forwarding policy. */
+                if (typeId == ID_CHANTYPE_TCPIP_FORWARD &&
+                        ssh->ctx->side == WOLFSSH_ENDPOINT_SERVER) {
+                    WLOG(WS_LOG_WARN, "Rejecting forwarded-tcpip channel open "
+                            "received by a server (wrong direction)");
+                    fail_reason = OPEN_ADMINISTRATIVELY_PROHIBITED;
+                    ret = WS_ERROR;
```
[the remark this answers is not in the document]
ejohnstown: Done. Moved the wrong-direction check above channelOpenCb and gated the policy callback and channel-state updates on ret == WS_SUCCESS, so invalid-direction opens trigger no hooks.

src/internal.c
```diff
@@ -8975,15 +8976,28 @@ static int DoGlobalRequestFwd(WOLFSSH* ssh,
...
+             * below it is a WS_FwdCbError status, where WS_FWD_SUCCESS is
+             * success and any other value is a rejection. */
+            if (!isCancel && cbRet >= WS_FWD_PORT_CHECK) {
+                bindPort = (word32)cbRet;
+            }
+            else if (cbRet != WS_FWD_SUCCESS) {
+                WLOG(WS_LOG_WARN, "Forward callback rejected the request, "
+                        "WS_FwdCbError = %d", cbRet);
+                ret = WS_RESOURCE_E;
+            }
```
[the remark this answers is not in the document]
ejohnstown: Fixed. The allocated-port return is now applied only when requestedPort == 0 and cbRet <= 65535; a port-like return on a non-zero request is ignored.

examples/echoserver/echoserver.c
```diff
@@ -566,8 +568,26 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
...
 
+        if (ret == 0 && port == 0) {
+            /* The peer requested port 0, so the OS picked the port during
+             * bind(). Recover it to report back to the caller. */
+            WMEMSET(&addr, 0, sizeof addr);
+            if (getsockname(appCtx->listenFd,
+                    (struct sockaddr*)&addr, &boundSz) == 0) {
+                allocatedPort = (word32)ntohs(addr.sin_port);
+                fwdCbCtx->hostPort = allocatedPort;
+            }
```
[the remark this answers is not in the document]
ejohnstown: Added: rejects an allocated port < WS_FWD_PORT_CHECK instead of reporting it.

ide/Espressif/ESP-IDF/examples/wolfssh_echoserver/main/echoserver.c
```diff
@@ -558,8 +560,26 @@ static int wolfSSH_FwdDefaultActions(WS_FwdCbAction action, void* vCtx,
...
 
+        if (ret == 0 && port == 0) {
+            /* The peer requested port 0, so the OS picked the port during
+             * bind(). Recover it to report back to the caller. */
+            WMEMSET(&addr, 0, sizeof addr);
+            if (getsockname(ctx->listenFd,
+                    (struct sockaddr*)&addr, &boundSz) == 0) {
+                allocatedPort = (word32)ntohs(addr.sin_port);
+                ctx->hostPort = allocatedPort;
+            }
```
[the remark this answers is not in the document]
ejohnstown: Same guard added here.

tests/regress.c
```diff
@@ -1152,6 +1164,54 @@ static int AcceptFwdCb(WS_FwdCbAction action, void* ctx,
...
+
+static int AllocatePortFwdCb(WS_FwdCbAction action, void* ctx,
+        const char* host, word32 port)
+{
+    (void)ctx;
+    (void)host;
+
+    /* A positive return value reports the allocated port for a port-0
+     * request; 0 (WS_SUCCESS) otherwise. */
+    if (action == WOLFSSH_FWD_REMOTE_SETUP && port == 0)
```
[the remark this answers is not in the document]
ejohnstown: Updated to describe the WS_FWD_PORT_CHECK boundary.

wolfssh/ssh.h
```diff
@@ -204,8 +204,19 @@ typedef enum WS_FwdCbError {
     WS_FWD_NOT_AVAILABLE,
     WS_FWD_INVALID_ACTION,
     WS_FWD_PEER_E,
+    /* Not an error code: the lowest unprivileged (ephemeral) port, and the
```
yosuke-wolfssl: This word "ephemeral" may be come from my commit but it's technically wrong now.
1024 is the privileged boundary (ports 1–1023 are privileged); it is not the lowest ephemeral port — ephemeral ranges start at 32768 (Linux) or 49152 (IANA)

yosuke-wolfssl (dismissed):
It looks good to me.
Maybe we can move WS_FWD_PORT_CHECK out from WS_FwdCbError since it's not pure error code, but it's acceptable.

wolfssh/ssh.h
```diff
@@ -204,8 +204,19 @@ typedef enum WS_FwdCbError {
...
+     * boundary of the callback return convention below. A return below this
+     * value is a WS_FwdCbError status; a return at or above it is the port a
+     * WOLFSSH_FWD_REMOTE_SETUP allocated for a port-0 (dynamic) request. */
+    WS_FWD_PORT_CHECK = 1024,
 } WS_FwdCbError;
 
+/* Return value convention: the callback returns either a WS_FwdCbError status
```
padelsbach: This big comment basically reiterates the new one above I'd say.

Also, should we mention the new case(s) where cleanup is called?

wolfssh/ssh.h
ejohnstown: Changed to call it privileged instead of ephemeral. Whether 1024, 32768, or 49152, it is still greater than WS_FWD_PEER_E. I'll move it to a #define that can be overridden. It's fine.

tests/regress.c
```diff
@@ -1465,6 +1525,58 @@ static void TestDirectTcpipNoFwdCbSendsOpenFail(void)
...
+     * admitted by the default-accept channelOpenCb path. */
+    extraSz = BuildDirectTcpipExtra("127.0.0.1", 8080, "127.0.0.1", 2222,
+            extra, sizeof(extra));
+    inSz = BuildChannelOpenPacket("forwarded-tcpip", 9, 0x4000, 0x8000,
+            extra, extraSz, in, sizeof(in));
+
+    InitChannelOpenHarness(&harness, in, inSz);
+    /* Intentionally do NOT register fwdCb */
+
+    ret = DoReceive(harness.ssh);
```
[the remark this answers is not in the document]
ejohnstown: Addressed. The forwarded-tcpip-with-no-fwdCb test was redundant: as a server harness it failed at the wrong-direction check before reaching the no-fwdCb gate, making it a duplicate of TestForwardedTcpipOnServerSendsOpenFail that couldn't catch a regression in the fail-closed path. Deleted it and its call site; TestDirectTcpipNoFwdCbSendsOpenFail already covers the no-fwdCb gate via the un-direction-checked direct-tcpip path.

src/internal.c
```diff
@@ -8992,6 +9014,27 @@ static int DoGlobalRequestFwd(WOLFSSH* ssh,
...
+         * port-0 (dynamic) request to be answered with the actual unprivileged
+         * port allocated. A successful callback that did not report one leaves
+         * bindPort below WS_FWD_PORT_CHECK, so we cannot comply: undo the setup
+         * and reject instead of sending a non-compliant success. */
+        if (requestedPort == 0 && bindPort < WS_FWD_PORT_CHECK) {
+            int cleanupRet;
+
+            WLOG(WS_LOG_WARN, "Forward callback reported no unprivileged port "
+                    "for a port-0 request; rejecting.");
+            cleanupRet = ssh->ctx->fwdCb(WOLFSSH_FWD_REMOTE_CLEANUP,
```
padelsbach: AI says we may want to check `ssh->ctx->fwdCb` for null. Looks like the other calls have the check.

wolfssh/ssh.h
```diff
@@ -204,8 +204,19 @@ typedef enum WS_FwdCbError {
     WS_FWD_NOT_AVAILABLE,
     WS_FWD_INVALID_ACTION,
     WS_FWD_PEER_E,
+    /* Not an error code: the lowest unprivileged (ephemeral) port, and the
+     * boundary of the callback return convention below. A return below this
```
padelsbach: This seems like a minor drift in the API. Any callers looking for `ret == WH_SUCCESS` specifically could break. Just want to make sure this is ok.

padelsbach (dismissed):

wolfssh/ssh.h
ejohnstown: This is a necessary change in the API. This change doesn't affect the function signature. It originally changed the port to a pointer so it could be updated. It was the easiest way to get the port number back out.

src/internal.c
ejohnstown: If you follow the logic, you can't get to the second call of fwdCb without already going through a NULL check. I still added the check, just not an else case.

wolfssh/ssh.h
ejohnstown: Simplified the comment.

padelsbach (approved):

PR #1060, opened 2026-06-25
Add OpenSSH certificate user authentication

yosuke-wolfssl:
## Add OpenSSH certificate user authentication

Adds OpenSSH certificate (`*-cert-v01@openssh.com`) user auth to wolfSSHd behind
`--enable-ossh-certs`, which is **opt-in and deliberately not part of
`--enable-all`** — existing `--enable-all` deployments are unaffected by this PR.
Operators can authorize users by trusting a single CA via `TrustedUserCAKeys`
instead of distributing every user's public key into `authorized_keys`.

### What's implemented (Unix)

**Library (`src/ossh.c`, `src/internal.c`)** — portable, no platform deps:
- Parse the certificate wire format; verify the **CA signature** against the
  embedded CA key (RSA incl. `rsa-sha2-256/512`, ECDSA P-256/384/521, Ed25519).
- **Critical options** parsed in strict ascending order; out-of-order, duplicate,
  or unknown names rejected. Only `force-command` and `source-address` are
  supported, and an empty or malformed value is **rejected** rather than treated
  as unrestricted.
- **Extensions** are parsed and tolerant (unknown, duplicate, and out-of-order
  accepted) but **not enforced** — see Known limitations.
- Reconstruct the certified base user key and verify the **user's signature**
  against it, using the on-the-wire signed length.
- Policy fields (`caKey`, principals, validity, `source-address`,
  `force-command`) are handed to the user-auth callback.

**Daemon (`apps/wolfsshd/auth.c`, `wolfsshd.c`)** — `CheckPublicKeyUnix` enforces
fail-closed, in order: CA trust → principal binding (a principal-less cert is
rejected, matching `sshd`) → validity window → source-address → force-command.

### Security model

The library proves only that the certificate is self-consistent (signed by its
embedded CA) and that the client holds the certified private key. It does **not**
decide trust — the callback checks `caKey` against a trust store and enforces
policy, mirroring how public-key auth delegates the `authorized_keys` check. The
library fails closed when no user-auth callback is set.

### ForceCommand and certificate force-command

A **certificate `force-command`** restricts what the session may do — it denies
file transfer unless the command is exactly `internal-sftp`. A configured
**`ForceCommand`** replaces what is run for shell and exec, and does not by
itself deny SFTP or SCP.

| Config `ForceCommand` | Cert `force-command` | SFTP | SCP | Shell/exec runs |
|---|---|---|---|---|
| `/bin/wrapper` | — | allowed | allowed | `/bin/wrapper` |
| — | `/bin/backup` | denied | denied | `/bin/backup` |
| — | `internal-sftp` | allowed | denied | `internal-sftp` |
| `/bin/wrapper` | `/bin/backup` | denied | denied | `/bin/wrapper` |

Row 1 is unchanged from master. Row 4 deliberately diverges from OpenSSH, which
checks `adm_forced_command` first and never reaches the certificate's command, so
a configured `ForceCommand` lifts the CA's restriction entirely; wolfSSHd honors
both. wolfSSHd also denies where `sshd` substitutes the forced command and lets
the client fail against the wrong program — the observable outcome matches in
every row.

### Configuration

```
# wolfsshd config
TrustedUserCAKeys /etc/wolfssh/trusted-user-ca-keys.pub
```

```sh
./configure --enable-ossh-certs --enable-sshd   # or --enable-all --enable-ossh-certs
```

### Platform support

- **Unix:** fully supported and enforced.
- **Windows:** intentionally **fails closed** — a certificate is never accepted on
  the threaded path. Follow-up in two PRs (per-connection cert-state relocation,
  then enforcement plus a Windows CI cert-auth job).

### Testing

- **`tests/api.c`** — parse, CA-signature verify, critical-option handling and
  extension tolerance, `OsshCertBaseId` mapping, `OsshRsaCertSigId` selection, and
  rejection of a CA blob whose curve name and key type disagree.
- **`tests/unit.c`** — Ed25519 public-key parse, the user-signature wire-length
  path, and the `wolfSSH_RsaVerify` NULL-signature guard.
- **`test_configuration.c`** — principal binding (incl. empty-list rejection),
  validity window, source-address CIDR matching and enforcement through
  `CheckPublicKeyUnix`, and force-command precedence.
- **`sshd_ossh_cert_test.sh`** — end-to-end via the wolfSSH and OpenSSH clients:
  valid certs (Ed25519/RSA/ECDSA CAs and user keys), untrusted CA, wrong and empty
  principal, expired, unknown critical option, source-address match/deny,
  force-command, `internal-sftp`, SFTP/SCP gating, and a configured `ForceCommand`
  alongside a certificate.
- Wired into CI explicitly, since `--enable-all` no longer implies the feature:
  `sshd-test.yml` builds `--enable-all --enable-ossh-certs` (so `make check` and
  the `run_all_sshd_tests.sh` e2e suite cover it), and `os-check.yml` adds an
  `--enable-all --enable-ossh-certs` matrix entry on Ubuntu and macOS. The plain
  `--enable-all` entries stay, so both states are built. `sshd_ossh_cert_test.sh`
  self-skips (exit 77) when the feature is off.

### Known limitations

- **Certificate extensions are parsed but not enforced.** `permit-pty`,
  `permit-port-forwarding`, `permit-agent-forwarding`, `permit-X11-forwarding`
  and `permit-user-rc` are recognized and validated, then ignored. In OpenSSH
  these are opt-in permissions where absence means deny, so wolfSSH is the
  permissive direction here. The live gap is `permit-pty`: a
  `ssh-keygen -O clear -O force-command=...` certificate still gets an
  interactive pty, because `SHELL_Subsystem()` decides on `SHELL_IsPty(ssh)`
  alone. The four forwarding/user-rc extensions have no corresponding wolfSSHd
  feature to gate today. Enforcement is a follow-up.
- **Windows:** fail-closed only (see above).
- **Source-address negation** (`!entry`) is not implemented; a negated entry fails
  the whole list closed and logs the offending entry.
- **Host certificates** are not implemented, and this PR adds **no public API**
  for them — the loader arrives with the feature. With `--enable-ossh-certs` and
  X.509 disabled, a configured `HostCertificate` now fails daemon startup with
  `WS_UNIMPLEMENTED_E` — the previous guard was a never-defined macro, so the
  directive was read and silently discarded. Remove `HostCertificate` or enable
  X.509. This only affects builds that opt into `--enable-ossh-certs`;
  `--enable-all` keeps its current behavior, including the `TrustedUserCAKeys`
  X.509 load failure.
- No C-level test drives the certificate reconstruction and user-signature path in
  `DoUserAuthRequestPublicKey`; it is covered end-to-end only. Follow-up.

### Notes for reviewers

- `WS_UserAuthData_PublicKey`: the cert fields are appended after every existing
  member, and `isOsshCert` sits after `dataToSignSz` rather than packing with
  `isCert`, so every member a user-auth callback can reach keeps the same offset
  whether or not `WOLFSSH_OSSH_CERTS` is enabled (verified with `offsetof` in both
  builds). A flag mismatch therefore truncates rather than misaligns. The struct
  is a stack local in `DoUserAuthRequest()`, so the padding is not a
  per-connection cost.
- Client-side host-key advertisement uses `cannedKeyAlgoNamesHostKey`, which omits
  the `*-cert-v01@openssh.com` names (host-certificate verification is
  unimplemented) while keeping the plain, X.509, and ML-DSA names in sync.
- `src/ossh.c` is in the Windows MSVC project because the OpenSSH key decoders it
  defines are referenced unconditionally.


src/internal.c
```diff
@@ -16002,6 +15858,24 @@ static int PrepareUserAuthRequestPublicKey(WOLFSSH* ssh, word32* payloadSz,
         keySig->sigId = matchId;
         keySig->sigName = IdToName(matchId);
         keySig->sigNameSz = (word32)WSTRLEN(keySig->sigName);
+    #ifdef WOLFSSH_OSSH_CERTS
```
aidangarske: **🟠 [High] RSA certificate guards assume rsa-sha2-256 is always available**
*🚫 BLOCK*  `bug`

The new OpenSSH RSA certificate path can reference `cannedKeyAlgoOsshRsaSha2_256CertName` even when `WOLFSSH_NO_RSA_SHA2_256` is defined. `internal.h` allows `WOLFSSH_NO_RSA_SHA2_256` to be set by `NO_SHA256` while `rsa-sha2-512` remains available, so an `--enable-ossh-certs` build with SHA-256 disabled and SHA-512 enabled fails to compile on this fallback. The same guard mismatch also advertises/maps `rsa-sha2-512-cert-v01@openssh.com` while parts of the cert handling gate `ID_OSSH_CERT_RSA` on SHA2-256, leaving SHA512-only builds inconsistent.

**Recommendation:** Use one consistent RSA-cert availability guard based on RSA plus at least one RSA SHA2 signature algorithm. Only reference the SHA2-256 cert name under `#ifndef WOLFSSH_NO_RSA_SHA2_256`, choose SHA2-512 when SHA2-256 is disabled, and apply the same guard to `cannedKeyAlgoNames`, `NameIdMap`, `cannedKeyAlgoClient`, `OsshCertBaseId`, and `OsshRsaCertSigId`. Add a compile configuration covering `WOLFSSH_OSSH_CERTS` with `WOLFSSH_NO_RSA_SHA2_256` and SHA2-512 still enabled.

src/ossh.c
```diff
@@ -0,0 +1,1234 @@
...
+        idx = 0;
+        ret = GetStringRef(&sigTypeSz, &sigType, cert->signature,
+                cert->signatureSz, &idx);
+    }
+    if (ret == WS_SUCCESS) {
+        ret = GetStringRef(&sigSz, &sig, cert->signature, cert->signatureSz,
+                &idx);
+    }
+
+    /* select hash from the signature algorithm name */
```
aidangarske, line 1125: **🟡 [Medium] OpenSSH RSA certificate CA signatures bypass SHA-1 soft-disable policy**
*💡 SUGGEST*

The PR introduces OpenSSH certificate verification and accepts an RSA CA certificate signature whose signature algorithm string is `ssh-rsa`, mapping it to `WC_HASH_TYPE_SHA`. Data Flow: remote SSH userauth public-key packet -> certificate blob parsed by `DoUserAuthRequestPublicKey` -> `OsshCertParse` extracts the attacker-supplied signature algorithm from the certificate signature field -> `OsshCertVerifySignature` dispatches to `OsshVerifyRsa` -> SHA-1 RSA verification is accepted -> the daemon later trusts the embedded CA key via `TrustedUserCAKeys`. This is inconsistent with the existing user-auth algorithm list, where plain `ID_SSH_RSA` is only advertised when `WOLFSSH_NO_SHA1_SOFT_DISABLE` is defined. As a result, builds that otherwise soft-disable SHA-1 RSA can still authenticate OpenSSH certificates signed by a trusted RSA CA using SHA-1. A malicious or compromised certificate issuer, or an attacker able to exploit SHA-1 collision weaknesses in a CA-signing workflow, gets a weaker authentication path than the build policy intends. This issue is introduced by the PR's new RSA OpenSSH certificate verifier; no existing test asserts that SHA-1 OSSH CA signatures are rejected under the soft-disable configuration.

**Recommendation:** Apply the same SHA-1 soft-disable policy used by normal public-key authentication, or reject `ssh-rsa` CA signatures entirely by default. For example gate the `ssh-rsa` branch under `#ifdef WOLFSSH_NO_SHA1_SOFT_DISABLE`. Also add a negative OpenSSH certificate test that signs an RSA-CA certificate with `ssh-rsa` and verifies rejection when SHA-1 is soft-disabled.

tests/api.c
```diff
@@ -1388,6 +1391,388 @@ static void test_wolfSSH_agent_signrequest_success(void)
...
+  0x17, 0xb8, 0x5f, 0x8c, 0xe7, 0x00, 0x00, 0x00, 0x20, 0x70, 0xfa, 0xfa,
+  0xf1, 0x0e, 0xd5, 0x70, 0x22, 0x47, 0xea, 0x6a, 0x7c, 0xa6, 0x98, 0xc3,
+  0x2c, 0x9e, 0xca, 0x2c, 0xb5, 0xe0, 0xcc, 0x46, 0x36, 0x6a, 0xb7, 0x8c,
+  0x49, 0xda, 0x40, 0x97, 0x30
+};
+
+/* Parse, verify the CA signature, and validate the options of each committed
+ * certificate vector; then flip the final signature byte and confirm the
+ * verification fails while the parse still succeeds. */
+static void test_wolfSSH_OsshCert_valid(void)
```
aidangarske, line 1962: **🟡 [Medium] OSSH certificate API test assumes RSA and ECDSA are enabled**
*💡 SUGGEST*  `test`

The new certificate verification test is guarded only by `WOLFSSH_OSSH_CERTS` and `!WOLFSSH_NO_ED25519`, but it always verifies vectors signed by Ed25519, RSA, and ECDSA CAs. Builds that enable OSSH certs and Ed25519 while disabling RSA or ECDSA will compile out the corresponding verifier and then fail this test, even though that algorithm-disabled configuration is otherwise valid.

**Recommendation:** Build the vector list conditionally: include the Ed25519 CA vector under Ed25519, the RSA CA vector under RSA support, and the ECDSA CA vector under ECDSA plus the needed curve support. Apply the same per-algorithm guards to the malformed verifier `caTypes` list.

apps/wolfsshd/test/run_all_sshd_tests.sh
```diff
@@ -176,6 +176,14 @@ else
         printf "Shutting down test wolfSSHd\n"
```
aidangarske, line 413: **🟡 [Medium] New sshd OSSH cert test cannot be selected or excluded**
*💡 SUGGEST*  `test`

`sshd_ossh_cert_test.sh` is invoked manually at the end of the local-host path, but it is not added to the `test_cases` array used by `--match` validation and `--exclude`. As a result, `--match sshd_ossh_cert_test.sh` is rejected as unknown, and `--exclude sshd_ossh_cert_test.sh` does not suppress the new test.

**Recommendation:** Add `sshd_ossh_cert_test.sh` to the selectable test list, or add a separate extra-test list that participates in `--match` and `--exclude` validation before running the self-contained local-host test.

aidangarske (dismissed):
## 🐺 Skoll Code Review

### Posted findings
- [High] **RSA certificate guards assume rsa-sha2-256 is always available** — `src/internal.c:15861-15876`
- [Medium] **OpenSSH RSA certificate CA signatures bypass SHA-1 soft-disable policy** — `src/ossh.c:1013-1025`
- [Medium] **OSSH certificate API test assumes RSA and ECDSA are enabled** — `tests/api.c:1579-1598`
- [Medium] **New sshd OSSH cert test cannot be selected or excluded** — `apps/wolfsshd/test/run_all_sshd_tests.sh:5-14`

*Review generated by Skoll.*

src/internal.c
yosuke-wolfssl: Thanks, fixed as suggested.

src/ossh.c
yosuke-wolfssl: Fixed using WOLFSSH_NO_SHA1_SOFT_DISABLE

tests/api.c
yosuke-wolfssl: You are correct, I fixed it

apps/wolfsshd/test/run_all_sshd_tests.sh
yosuke-wolfssl: Thanks — accurate as stated, but I'd like to push back on scoping this to the new test.

`test_cases` only covers 7 of the tests this script runs. `sshd_forcedcmd_test.sh`,
`sshd_window_full_test.sh`, `sshd_empty_password_test.sh`, `sshd_login_grace_test.sh`
and `sshd_x509_test.sh` are all likewise invoked directly outside the array, and are
equally unreachable via `--match` and unaffected by `--exclude`. `sshd_ossh_cert_test.sh`
follows that existing convention — `sshd_x509_test.sh`, the closest analogue, does exactly
the same thing (it also needs its own daemon started with a specific config).

So this is a pre-existing limitation of the runner rather than something this PR
introduces. I'd rather not special-case the new test: either we leave it consistent with
its neighbors, or we fix the runner for all of them (e.g. an `extra_test_cases` list that
participates in `--match`/`--exclude` validation) as a separate change that isn't tied to
the OpenSSH cert feature.

Happy to do the latter as a follow-up, or let me know if you prefer to fix them in this PR now.

apps/wolfsshd/test/test_configuration.c
```diff
@@ -1899,6 +1899,519 @@ static int test_OpenSecureFile(void)
...
+    user = pw->pw_name;
+    nameLen = (word32)WSTRLEN(user);
+    if (nameLen == 0 || nameLen + 4 > (word32)sizeof(principals)) {
+        return WS_FATAL_ERROR;
+    }
+    /* principals list holding the running account: [uint32 len][name]. */
+    principals[0] = (byte)((nameLen >> 24) & 0xff);
+    principals[1] = (byte)((nameLen >> 16) & 0xff);
+    principals[2] = (byte)((nameLen >> 8) & 0xff);
+    principals[3] = (byte)(nameLen & 0xff);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/test/sshd_ossh_cert_test.sh
```diff
@@ -0,0 +1,287 @@
...
+
+[ -x "$WOLFSSHD" ] || skip "wolfsshd not built, skipping OpenSSH cert test"
+[ -x "$CLIENT" ]   || skip "wolfSSH client not built, skipping OpenSSH cert test"
+command -v ssh-keygen >/dev/null 2>&1 || \
+    skip "ssh-keygen not found, skipping OpenSSH cert test"
+
+WORK=$(mktemp -d)
+trap 'pkill -f "wolfsshd .*sshd_config_ossh" 2>/dev/null; rm -rf "$WORK"' EXIT
+
+# The suite may run under sudo (CI): this script runs as root while the daemon
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/test/sshd_ossh_cert_test.sh
```diff
@@ -0,0 +1,287 @@
...
+MARKERDIR="$WORK/markers"
+mkdir -p "$MARKERDIR"
+chmod 777 "$MARKERDIR"
+
+# The host private key is a secret loaded through the secure gate, which refuses
+# a group/world readable file. The committed key is 644, so use a 600 copy.
+HOSTKEY="$WORK/hostkey.pem"
+cp "$ROOT/keys/server-key.pem" "$HOSTKEY"
+chmod 600 "$HOSTKEY"
+
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

yosuke-wolfssl:
Hello @aidangarske , @ejohnstown ,
Could you review this again ?

aidangarske:
This is something @ejohnstown should review when he has time

keys/fred-ossh-forcecmd-cert.pub
ejohnstown, line 1: This committed cert bakes a fixed force-command touch /tmp/wolfsshd_ossh_forced_marker (visible via ssh-keygen -L), which is the pre-fix world-writable /tmp marker path we moved away from in the daemon test. It's harmless in practice — but only because nothing reads it: sshd_ossh_cert_test.sh regenerates all cert variants at runtime via renew-ossh-certs.sh "$LOGINUSER" into $LOGINUSER-ossh-*-cert.pub, and the force-command is injected there from the environment (-O force-command="touch $OSSH_FORCED_MARKER", pointing at the per-run $MARKERDIR). The committed fred-ossh-* certs are never read by the shell test, the C tests (which embed wire vectors inline), or anything in src/; they only appear as EXTRA_DIST in keys/include.am.

So the stale path is really a symptom of a bigger question: do these 14 fred-ossh-*-cert.pub files need to be checked in at all? Since they're regenerated per-login-user and never consumed, I'd suggest dropping them (and their include.am lines) and keeping only the CA/user keypairs, which are used (reused for signing and cat'd into TrustedUserCAKeys). make dist stays self-contained — the tarball ships the keypairs plus renew-ossh-certs.sh, and the test skips cleanly (77) without ssh-keygen.

If you'd rather keep the certs as human-readable reference material, that's fine too — but then please regenerate this one so the committed baseline doesn't advertise the old /tmp path.

ejohnstown (changes requested):

keys/fred-ossh-forcecmd-cert.pub
yosuke-wolfssl: Thanks, You are correct.
I removed the unused certs.

yosuke-wolfssl:
Hello @ejohnstown ,
All tests are passed. Can you review this again please ?

ejohnstown (changes requested):
Review comments provided directly.

ejohnstown (approved):

apps/wolfsshd/test/run_all_sshd_tests.sh
ejohnstown: Done. See the follow-up.

PR #1065, opened 2026-06-26
Port wolfSSH Zephyr test sample to Zephyr 4.4.0

ejohnstown:
Brings the Zephyr test sample and its CI up to date with Zephyr 4.x APIs while keeping it buildable on the older 3.4.0 release.

- Fix keyboard-interactive build without TERM.
- Port Zephyr test sample to 3.4.0 and 4.4.0.
- Test Zephyr sample on 3.4.0 and 4.4.0.


.github/workflows/zephyr.yml
```diff
@@ -13,34 +13,23 @@ jobs:
...
+          # Each leg runs in the Zephyr Project CI image whose bundled SDK
+          # matches the Zephyr release under test: v0.26.4 ships SDK 0.16.1
+          # (for Zephyr 3.4.0), v0.29.0 ships SDK 1.0.0 (for Zephyr 4.4.0).
           - zephyr-ref: v3.4.0
-            zephyr-sdk: 0.16.1
+            docker-image: v0.26.4
+          - zephyr-ref: v4.4.0
+            docker-image: v0.29.0
     runs-on: ubuntu-22.04
+    container:
```
aidangarske, line 25: **🔴 [High] Unsupported job container entrypoint option breaks Zephyr CI** · _API Contract Violations_

The PR introduces a GitHub Actions job container and sets `container.options` to `--entrypoint /bin/bash`. GitHub's workflow syntax for `jobs.\<job_id>.container.options` explicitly documents that the `--entrypoint` (and `--network`) options are not supported. GitHub manages the job-container entrypoint itself to keep the container alive and exec each step into it, so overriding it via container options is unsupported and can prevent the container from starting or being kept alive before any `west`/`twister` step runs. The base workflow used no job container, so this regression is introduced by the PR and gates every Zephyr matrix leg. Verified against the referenced code: lines 24-26 add the container with the unsupported option.

**Fix:** Remove the unsupported option. If the image needs a neutral entrypoint, publish/use an image with that entrypoint baked in instead of overriding it through job-container options.

container:
  image: ghcr.io/zephyrproject-rtos/ci:${{ matrix.config.docker-image }}

.github/workflows/zephyr.yml
```diff
@@ -13,34 +13,23 @@ jobs:
     strategy:
```
aidangarske, line 13: **🔵 [Low] Version matrix lacks fail-fast: false, so one Zephyr leg failing cancels the other** · _convention_

The stated goal of the PR is to validate the sample on BOTH Zephyr 3.4.0 and 4.4.0. GitHub Actions matrix strategy defaults to `fail-fast: true`, which cancels all in-progress matrix legs as soon as one fails. If the 3.4.0 leg (the harder-to-keep-working legacy port) fails to build, the 4.4.0 leg will be cancelled before it finishes, hiding whether 4.4.0 still passes (and vice-versa). For a version-compatibility matrix you almost always want both results independently. Verified: the `strategy:` block contains only `matrix:` with no `fail-fast:` key.

**Fix:** Add `fail-fast: false` under `strategy:` so each Zephyr version reports its own pass/fail independently.

zephyr/samples/tests/CMakeLists.txt
```diff
@@ -1,5 +1,41 @@
...
+if(NOT DEFINED ZEPHYR_VER_MAJOR OR NOT DEFINED ZEPHYR_VER_MINOR)
+  message(FATAL_ERROR
+    "Could not determine Zephyr version from $ENV{ZEPHYR_BASE}/VERSION")
+endif()
+set(ZEPHYR_VER "${ZEPHYR_VER_MAJOR}.${ZEPHYR_VER_MINOR}")
+
+# RAM disk: configured via the devicetree "zephyr,ram-disk" node on Zephyr
+# >= 3.5, via Kconfig before that. Only the filesystem-backed scenarios need
+# it; the nofs build (CONF_FILE="prj_nofs.conf") has no disk.
+if(NOT "${CONF_FILE}" MATCHES "nofs")
```
aidangarske, line 28: **🔵 [Low] RAM-disk selection keys off a substring match of the CONF_FILE filename** · _convention_

`if(NOT "${CONF_FILE}" MATCHES "nofs")` decides whether to add the RAM disk purely from the presence of the substring "nofs" in the CONF_FILE filename. This works for today's three scenarios (`prj.conf`, `prj_kbi.conf`, `prj_nofs.conf`), but it is an implicit coupling: a future filesystem-less config that does not happen to contain "nofs" in its name would silently pull in the RAM disk overlay, and a filesystem config that happened to contain "nofs" would lose it. The intent ("this scenario has no filesystem") is not self-documenting from the filename convention alone. Verified present in CMakeLists.txt.

**Fix:** Optional: drive the RAM disk decision off an explicit toggle (e.g. a `-DWOLFSSH_WITH_FS=y/n` extra_arg per scenario in sample.yaml) rather than a filename substring, so the coupling is explicit. Not blocking given the current fixed scenario set.

zephyr/samples/tests/posix_legacy.conf
```diff
@@ -0,0 +1,14 @@
+# Pre-4.1 POSIX layer.
```
aidangarske: **🔵 [Low] New config/overlay files open with multi-line description blocks contrary to repo comment convention** · _style_

The new `posix_legacy.conf`, `ramdisk_legacy.conf`, and `ramdisk.overlay` each open with a multi-line file-header comment block, and `CMakeLists.txt` adds several multi-line comment blocks. The wolfSSL-ecosystem convention favors minimal comments — a one-line WHY when it would surprise a reader — rather than multi-line file-header/explanation blocks. The WHYs here (version boundaries, POSIX reorg) are genuinely worth a note, but could be condensed to a single line each, with the longer rationale living in the PR/commit description. Verified: posix_legacy.conf opens with a 10-line comment block, and CMakeLists.txt carries multiple multi-line explanatory blocks.

**Fix:** Optionally collapse the multi-line header blocks in the new .conf/.overlay files (and CMakeLists comments) to a single WHY line each; move the extended rationale to the PR description. Purely stylistic.

aidangarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [High] [bugs] **Unsupported job container entrypoint option breaks Zephyr CI** — `.github/workflows/zephyr.yml:24-26`
- [Low] [review] **Version matrix lacks fail-fast: false, so one Zephyr leg failing cancels the other** — `.github/workflows/zephyr.yml:13-22`
- [Low] [review] **RAM-disk selection keys off a substring match of the CONF_FILE filename** — `zephyr/samples/tests/CMakeLists.txt:24`
- [Low] [review] **New config/overlay files open with multi-line description blocks contrary to repo comment convention** — `zephyr/samples/tests/posix_legacy.conf:1-10`

.github/workflows/zephyr.yml
ejohnstown: Changed. The entrypoint ends up being a no-op on the action runner. Removed it anyway.

.github/workflows/zephyr.yml
ejohnstown: Disabled fail-fast.

zephyr/samples/tests/posix_legacy.conf
ejohnstown: Trimmed the comments.

zephyr/samples/tests/CMakeLists.txt
ejohnstown: Fair point, but with the scenario set fixed at three (prj.conf, prj_kbi.conf, prj_nofs.conf) the filename match is unambiguous, and the intent is spelled out in the adjacent comment. Leaving as-is; would revisit if we add filesystem-less scenarios that don't carry nofs in the name.

aidangarske (approved):
Looks good

PR #1067, opened 2026-06-28
wolfsshd: fail closed when a per-connection privilege drop fails

yosuke-wolfssl:
# wolfsshd: fail closed when a per-connection privilege drop fails

## Summary

Fixes a privilege-handling defect in `wolfsshd` where a failed per-connection
drop to the authenticated user's uid/gid could leave the connection handler
running at an elevated privilege level (root, or the sshd-daemon uid) instead
of terminating.

Addressed by f_5850.

## Background

When servicing a channel request, the subsystem handlers
(`SHELL_Subsystem`, `SCP_Subsystem`, `SFTP_Subsystem`) drop privileges to the
authenticated user via `wolfSSHD_AuthReducePermissionsUser()`. On failure, the
old code attempted a fallback to `wolfSSHD_AuthReducePermissions()` and then
did `return WS_FATAL_ERROR`:

```c
if (wolfSSHD_AuthReducePermissionsUser(conn->auth, pPasswd->pw_uid,
        pPasswd->pw_gid) != WS_SUCCESS) {
    wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error setting user ID");
    if (wolfSSHD_AuthReducePermissions(conn->auth) != WS_SUCCESS) {
        exit(1);
    }
    return WS_FATAL_ERROR;
}
```

Two problems combined into a privilege-management hole:

1. **The fallback is a no-op when privilege separation is off.** With
   `UsePrivilegeSeparation` set to `WOLFSSHD_PRIV_OFF`,
   `wolfSSHD_AuthReducePermissions()` performs no uid change and returns
   `WS_SUCCESS`. So after a failed drop-to-user, the fallback "succeeds"
   without lowering anything.

2. **`HandleConnection` discarded the `SHELL_Subsystem` return value.** The
   `WS_FATAL_ERROR` was never captured, so `ret` stayed `WS_SUCCESS`. The
   post-switch error check consults `wolfSSH_get_error(ssh)`, which a kernel
   `setreuid()`/`setregid()` failure does not set, so `wolfSSH_shutdown` and up
   to ten `wolfSSH_worker` iterations would execute while the effective uid was
   still above the authenticated user.

Triggering the drop failure itself requires an environmental condition that
forces `setreuid`/`setregid` to fail (e.g. `RLIMIT_NPROC` saturation at the
target uid, an LSM denial, or user-namespace mapping limits), so this is not
routinely reachable in default deployments — but when it does occur the handler
must fail closed.

## Fix

- Every per-connection privilege-drop failure path now calls `exit(1)`
  directly instead of attempting the no-op fallback and returning. This is
  strictly stronger than propagating an error into the connection handler:
  the per-connection process terminates before any work runs at the wrong
  privilege level. Applied consistently across `SHELL_Subsystem` (the forked
  child branch and the parent branch, plus the `dup2`/`setgroups`/`chroot`
  failure paths), `SCP_Subsystem`, and `SFTP_Subsystem`.

- `HandleConnection` now captures the return value at both subsystem call
  sites (`ret = SHELL_Subsystem(...)` for the shell and exec sessions), so a
  failure is reflected in `ret` as well (defense in depth; `exit(1)` already
  makes the path unreachable).

- In the `SHELL_Subsystem` parent branch the `forkpty` child has already been
  spawned, so it is killed with `kill(childPid, SIGKILL)` before `exit(1)` to
  guarantee deterministic teardown rather than relying on pty/pipe closure,
  matching the existing unexpected-error handling later in the same function.

The daemon's own startup drop in `main()` is unchanged: it already fails closed
by setting `ret = WS_FATAL_ERROR`, which prevents the listen/accept loop from
starting.

## Testing

- `wolfsshd` builds cleanly with the project configuration.
- No functional behavior change on the success path; only the
  privilege-drop-failure paths are affected, and they now terminate the
  per-connection process.

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -601,12 +601,10 @@ static int SCP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     if (wolfSSHD_AuthReducePermissionsUser(conn->auth, pPasswd->pw_uid,
             pPasswd->pw_gid) != WS_SUCCESS) {
         wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error setting user ID");
-        if (wolfSSHD_AuthReducePermissions(conn->auth) != WS_SUCCESS) {
-            /* stop everything if not able to reduce permissions level */
-            exit(1);
-        }
-
-        return WS_FATAL_ERROR;
+        /* could not drop to the authenticated user; terminate the
+         * per-connection process rather than continue at a higher
+         * privilege level */
+        exit(1);
     }
 #else
     /* impersonate the logged on user for file permissions */
@@ -720,12 +718,10 @@ static int SFTP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     if (wolfSSHD_AuthReducePermissionsUser(conn->auth, pPasswd->pw_uid,
             pPasswd->pw_gid) != WS_SUCCESS) {
         wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error setting user ID");
-        if (wolfSSHD_AuthReducePermissions(conn->auth) != WS_SUCCESS) {
-            /* stop everything if not able to reduce permissions level */
-            exit(1);
-        }
-
-        return WS_FATAL_ERROR;
+        /* could not drop to the authenticated user; terminate the
+         * per-connection process rather than continue at a higher
+         * privilege level */
+        exit(1);
     }
 #else
     char r[MAX_PATH];
@@ -1410,7 +1406,8 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         return WS_FATAL_ERROR;
     }
     else if (childPid == 0) {
-        /* Child process */
+        /* Child process. Any setup failure below exit(1)s rather than returning
+         * into the connection handler while still at a raised privilege level. */
         const char *args[] = {"-sh", NULL, NULL, NULL};
         char cmd[MAX_COMMAND_SZ];
         char shell[MAX_COMMAND_SZ];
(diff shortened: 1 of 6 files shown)


apps/wolfsshd/wolfsshd.c
```diff
@@ -595,12 +595,10 @@ static int SCP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     if (wolfSSHD_AuthReducePermissionsUser(conn->auth, pPasswd->pw_uid,
```
aidangarske, line 601: **🟡 [Medium] Add regression coverage for fail-closed subsystem paths**
*💡 SUGGEST*  `test`

The PR changes several privilege-drop failure paths from returning `WS_FATAL_ERROR` after a fallback attempt to calling `exit(1)`, and it changes `HandleConnection` to retain `SHELL_Subsystem` return values. Existing tests cover `wolfSSHD_AuthReducePermissionsUser()` itself, but I did not find tests that exercise the subsystem call sites or verify that `HandleConnection` no longer discards shell/exec subsystem failures. Without a regression test around these changed paths, a future edit could accidentally reintroduce a return path into the connection handler after a failed per-connection privilege drop.

**Suggestion:**
```suggestion
Add a focused regression test using the existing `WOLFSSHD_UNIT_TEST` privilege-drop stubs. Run the subsystem failure case in a child process and assert that it exits non-zero, and add coverage that a `SHELL_Subsystem` failure is assigned into `ret` at the shell and exec call sites.
```

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Add regression coverage for fail-closed subsystem paths** — `apps/wolfsshd/wolfsshd.c:595-601`

*Review generated by Skoll.*

apps/wolfsshd/wolfsshd.c
yosuke-wolfssl: Thanks — added the regression coverage in `apps/wolfsshd/test/sshd_privdrop_fail_test.sh`.

One correction to the suggestion: asserting that `SHELL_Subsystem`'s return value "is
assigned into `ret`" wouldn't prove anything. `HandleConnection`'s post-switch gate reads
`wolfSSH_get_error(ssh)` and never consults `ret` (`wolfsshd.c:2230`) — that's also true on
master for `SFTP_Subsystem`/`SCP_Subsystem`, whose return values have always been captured
and then discarded by the same gate. So a test asserting `ret` would pass on the buggy code
too. The guarantee is `exit(1)`, so the test asserts process termination instead.

It runs a `noinst` daemon build whose privilege drop is forced to fail, drives all four
converted paths (exec/shell, sftp, scp), and per connection checks the daemon log:

- must contain `Error setting user ID` — the failure path was actually reached
- must NOT contain `Attempting to close down connection` — the handler died instead of
  running `wolfSSH_shutdown` + `wolfSSH_worker` at the un-dropped euid

Negative control: restoring the old fallback-and-return makes the handler log
`Attempting to close down connection` and the test fails. Reintroducing it in SCP/SFTP only
fails just those phases. The fault injection is gated behind both `WOLFSSHD_UNIT_TEST` and a
new `WOLFSSHD_FAULT_INJECT`, so it's absent from the shipped `wolfsshd` binary (confirmed
with `strings`).

yosuke-wolfssl:
Hi @aidangarske . could you please review this ?

apps/wolfsshd/auth.c
```diff
@@ -93,8 +93,34 @@
 #endif
 
 #if defined(WOLFSSHD_UNIT_TEST) && !defined(_WIN32)
+#ifdef WOLFSSHD_FAULT_INJECT
```
aidangarske: Is there any way we can do this test without adding fault injections into the src?

yosuke-wolfssl: Thanks, I dropped the in-source fault injection entirely. auth.c and include.am
are now back to master (no callbacks, no WOLFSSHD_FAULT_INJECT, no faultinj daemon
build).

The test now drives the stock wolfsshd binary with a small LD_PRELOAD interposer
(apps/wolfsshd/test/sshd_privdrop_preload.c) that returns EPERM from
setregid()/setreuid() only when WOLFSSHD_FAULT_PRIVDROP is set, forwarding the real
call otherwise. It's compiled on-demand by the test, Linux-only (SKIPs where
DYLD stripping applies), and never touches the shipped daemon or library — so we're
now testing the actual production binary, not a special build.

yosuke-wolfssl:
Hi @aidangarske ,
Sorry to bother you again, could you review this please ?

aidangarske (approved):

ejohnstown (approved):

PR #1068, opened 2026-06-29
Add server-side IsMessageAllowed pre-auth gate regression tests

yosuke-wolfssl:
### Background

`IsMessageAllowed` (`src/internal.c`) dispatches to `IsMessageAllowedServer`
or `IsMessageAllowedClient` based on `ssh->ctx->side`, and is the per-message
pre-authentication gate invoked from `DoPacket`. Every
`wolfSSH_TestIsMessageAllowed` call in `tests/regress.c` ran against the single
client-endpoint `ssh` created in `main()`, so only the client branch was
exercised. The entire server gate (`IsMessageAllowedServer`) had no test of its
blocking behavior, so mutations of its pre-auth checks could survive the suite.

This is a test-coverage gap. There is no production code change.
Addressed by f_6519.

### What this PR does

Adds server-endpoint regression tests that mirror the existing client gate
tests. A single shared server `ctx`/`ssh` is created in `main()` (guarded by
`#ifndef NO_WOLFSSH_SERVER`), symmetric with the existing client pair, and each
test calls `ResetSession` then sets `acceptState` before asserting.

New tests in `tests/regress.c`:

- `TestServerChannelBlockedBeforeAuth` / `TestServerChannelAllowedAfterAuth` -
  `CHANNEL_OPEN` and `GLOBAL_REQUEST` are blocked one state below the
  auth-complete boundary and allowed at `ACCEPT_SERVER_USERAUTH_SENT`. Together
  these pin the `acceptState < ACCEPT_SERVER_USERAUTH_SENT` comparison.
- `TestServerUserauthBlockedBeforeKeyed` - the user auth request is blocked
  while the server is still keying and accepted once keyed (pins the pre-keyed
  message-range bound).
- `TestServerOnlyUserauthMsgsBlocked` - the user auth messages only the server
  sends (`USERAUTH_FAILURE`, `USERAUTH_SUCCESS`) are rejected, while the
  keyboard-interactive `USERAUTH_INFO_RESPONSE` the server receives is allowed.
- `TestServerServiceRequestStateGated` - `SERVICE_REQUEST` is accepted only
  once keyed, `SERVICE_ACCEPT` is never accepted, and the `SERVICE_ACCEPT`
  reject sets `WS_MSGID_NOT_ALLOWED_E`.

`ResetSession` now also clears `acceptState` so the harness baseline is
symmetric for both endpoint kinds.

### Note on mutation coverage

Of the two mutants originally suspected to survive, only one is real and
killable:

- Relaxing `acceptState < ACCEPT_SERVER_USERAUTH_SENT` to `<=` is a real mutant:
  it blocks a legitimate post-auth `CHANNEL_OPEN`. `TestServerChannelAllowedAfterAuth`
  catches it.
- Deleting the `if (MSGIDLIMIT_POST_USERAUTH(msg)) return 0;` block is an
  equivalent mutant for the return value: `MSGIDLIMIT_POST_USERAUTH(x)` is
  `x >= 80`, and every such message also trips the following
  `msg > MSGID_USERAUTH_REQUEST && msg != MSGID_USERAUTH_INFO_RESPONSE` check,
  which returns 0 anyway. Neither path sets `ssh->error`, so no test can
  distinguish it. It is intentionally not targeted.

### Testing

Built with `./configure --enable-all` and ran the regression suite:

```
$ ./tests/regress.test
...
regress: PASS
```

Negative control: relaxing the boundary comparison to `<=` makes
`TestServerChannelAllowedAfterAuth` fail, confirming the new test detects the
mutation; reverted afterward.

### Files changed

- `tests/regress.c` - five new server gate tests, shared server fixture in
  `main()`, and `acceptState` reset in `ResetSession`.

### DIFF
--- a/tests/regress.c
+++ b/tests/regress.c
@@ -79,6 +79,7 @@ static void ResetSession(WOLFSSH* ssh)
     }
     ssh->isKeying = 0;
     ssh->connectState = CONNECT_BEGIN;
+    ssh->acceptState = ACCEPT_BEGIN;
     ssh->error = 0;
 }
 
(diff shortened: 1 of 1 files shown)


aidangarske (approved):
Nice!

PR #1048, opened 2026-06-22
ML-DSA Support

stenslae:
- Implemented ML-DSA-44, ML-DSA-65, and ML-DSA-87 as public key authentication algorithms and their x509v3 certificate variants.

PR #1069, opened 2026-06-29
Fix WS_KeySignature leak on DoAsn1Key private-key error paths

yosuke-wolfssl:
## Description

`DoAsn1Key` (`src/ssh.c`) calls `IdentifyAsn1Key()`, which on success heap-allocates a `WS_KeySignature` and initializes an inner wolfCrypt key (`wc_InitRsaKey`/`wc_ecc_init`). That allocation is meant to be released at the end of the function via `wolfSSH_KEY_clean(key)` / `WFREE(key, ...)`.

In the `else if (ret > 0 && isPrivate)` branch, two early `return ret;` statements bypassed that cleanup:

- malloc-failure path (`WS_MEMORY_E`)
- caller-buffer-too-small path (`WS_BUFFER_E`)

Both run after `key` has been allocated, so each leaks the `WS_KeySignature` plus the `mp_int`/ECC internals owned by the embedded wolfCrypt key. The `*outSz < inSz` path is reachable from the public `wolfSSH_ReadKey_buffer` / `wolfSSH_ReadKey_buffer_ex` APIs whenever a caller passes a fixed `*out` buffer smaller than the DER private key.

Addressed by f_6513.

## Fix

Replaced the two early `return ret;` statements with fallthrough control flow so every error path reaches the existing `wolfSSH_KEY_clean(key)` / `WFREE(key, ...)` cleanup:

- malloc failure and buffer-too-small now set `ret` instead of returning.
- The success body is guarded with `if (ret > 0)`, which also preserves the positive key-type id needed by `IdToName(ret)`.
- The existing tail `if (*out == NULL) WFREE(newKey, ...)` remains correct for both error paths.

## Testing

Added a regression test in `tests/api.c` (`test_wolfSSH_ReadKey`) that drives the ASN.1/DER private-key error path with a caller-provided buffer one byte too small. Existing ASN.1 tests only used allocate-mode, leaving this path uncovered.

The test asserts:

- `ret == WS_BUFFER_E`
- caller buffer pointer is preserved (`key == keyCheck`)
- output params are left untouched on the error path (`keySz`, `keyType`, `keyTypeSz`)

The `WS_MEMORY_E` early-return path is fixed the same way but is not directly tested (it requires allocation-failure injection).

## Verification

- `./configure --enable-all` build + `tests/api.test`: pass.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1892,23 +1892,23 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
             newKey = (byte*)WMALLOC(inSz, heap, DYNTYPE_PRIVKEY);
             if (newKey == NULL) {
                 ret = WS_MEMORY_E;
-                return ret;
             }
         }
+        else if (*outSz < inSz) {
+            WLOG(WS_LOG_DEBUG, "DER private key output size too small");
+            ret = WS_BUFFER_E;
+        }
         else {
-            if (*outSz < inSz) {
-                WLOG(WS_LOG_DEBUG, "DER private key output size too small");
-                ret = WS_BUFFER_E;
-                return ret;
-            }
             newKey = *out;
         }
 
-        *out = newKey;
-        *outSz = inSz;
-        WMEMCPY(newKey, in, inSz);
-        *outType = (const byte*)IdToName(ret);
-        *outTypeSz = (word32)WSTRLEN((const char*)*outType);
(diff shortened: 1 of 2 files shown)


aidangarske (approved):
Looks good! please rebase then I will merge

yosuke-wolfssl:
Close this since the issue was solved in #1048 .

PR #1070, opened 2026-06-29
keyboard-interactive: send USERAUTH_FAILURE on response-count mismatch

yosuke-wolfssl:
### Problem

`DoUserAuthInfoResponse()` tore down the transport instead of sending an
`SSH_MSG_USERAUTH_FAILURE` when a keyboard-interactive `INFO_RESPONSE` carried a
`responseCount` that did not match the server's pending `promptCount`.

The guard set `ret = WS_USER_AUTH_E`, but the failure-response gate only sends a
message when `authFailure || partialSuccess` is set, so nothing was emitted;
`DoReceive()` converts `WS_USER_AUTH_E` to `WS_FATAL_ERROR` and the connection
dies. RFC 4256 requires the server to answer every `INFO_RESPONSE` with
`SUCCESS`, `FAILURE`, or another `INFO_REQUEST`. Impact is limited to
malformed/malicious clients, but the response is non-conformant.

Addressed by f_6514.

### Fix (`src/internal.c`)

Set `authFailure` in the mismatch guard so the existing gate sends
`SSH_MSG_USERAUTH_FAILURE` and the connection survives; the client may retry.
`ret` stays non-success so the response allocation, parse, and user-auth callback
are skipped on a protocol-violating message.

Two removals came out of review:

- The adjacent `responseCount > WOLFSSH_MAX_PROMPTS` guard was dead code. The
  mismatch check forces `responseCount == promptCount`, and `promptCount` is
  capped at `WOLFSSH_MAX_PROMPTS` by both of its writers
  (`SendUserAuthKeyboardRequest()` server-side, `DoUserAuthInfoRequest()`
  client-side), so the allocation below is bounded transitively.
- An `*idx = len` line on the failure path was an unobservable no-op:
  `DoReceive()` calls `ShrinkBuffer(&ssh->inputBuffer, 1)` on every consumed
  packet, and `forcedFree = 1` zeroes `length`/`idx`, so a short payload index
  never reaches the next packet.

Net change is one functional line plus a comment.

### Tests (`tests/regress.c`)

Both use the in-memory packet harness, feeding crafted packets into `DoReceive()`
and inspecting the bytes the server writes back:

- `TestKbInfoResponseCountMismatchSendsFailure` — `responseCount` 2 vs
  `promptCount` 1; asserts `WS_SUCCESS` and a `MSGID_USERAUTH_FAILURE` reply.
- `TestKbInfoResponseMismatchKeepsFraming` — two back-to-back malformed
  responses; asserts the second packet still parses and draws its own failure,
  covering parser state after a rejected response.

### Verification

- `make check`: 10 passed, 1 skipped (external), 0 failed.
- Clean under `-Werror` across 6 gcc-13 configs; lint clean.
- Negative controls: reverting `authFailure = 1` aborts the first test with
  `-1001`; the framing test fails only when `*idx = len` is absent *and*
  `ShrinkBuffer()` is forced non-free, confirming it detects real desync.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8223,12 +8223,10 @@ static int DoUserAuthInfoResponse(WOLFSSH* ssh,
     if ((ret == WS_SUCCESS) &&
         (ssh->kbAuth.promptCount != kb->responseCount)) {
         WLOG(WS_LOG_DEBUG, "DUARKB: Invalid number of responses received");
-        ret = WS_USER_AUTH_E;
-    }
-
-    if (ret == WS_SUCCESS && kb->responseCount > WOLFSSH_MAX_PROMPTS) {
-        WLOG(WS_LOG_DEBUG, "DUARKB: Received too many responses (%d), max: %d",
-             kb->responseCount, WOLFSSH_MAX_PROMPTS);
+        /* Answer with USERAUTH_FAILURE instead of tearing down the transport.
+         * Keeping ret non-success skips the allocation, parse, and callback
+         * below; past here responseCount equals the capped promptCount. */
+        authFailure = 1;
         ret = WS_USER_AUTH_E;
     }
 
--- a/tests/regress.c
+++ b/tests/regress.c
@@ -340,6 +340,103 @@ static void FreeChannelOpenHarness(ChannelOpenHarness* harness)
         wolfSSH_CTX_free(harness->ctx);
 }
 
+#ifdef WOLFSSH_KEYBOARD_INTERACTIVE
+/* Build a plaintext SSH_MSG_USERAUTH_INFO_RESPONSE. The wire response count is
+ * "responseCount"; "stringCount" "x" response strings are actually appended.
+ * The count-mismatch guard rejects before parsing the body, so a short body is
+ * sufficient to drive it. */
+static word32 BuildInfoResponsePacket(word32 responseCount, word32 stringCount,
+        byte* out, word32 outSz)
+{
+    byte payload[128];
+    word32 idx = 0;
+    word32 i;
+
+    idx = AppendUint32(payload, sizeof(payload), idx, responseCount);
+    for (i = 0; i < stringCount; i++) {
+        idx = AppendString(payload, sizeof(payload), idx, "x");
+    }
+
+    return WrapPacket(MSGID_USERAUTH_INFO_RESPONSE, payload, idx, out, outSz);
+}
+
+/* Place the server mid keyboard-interactive auth, expecting an INFO_RESPONSE
+ * for "promptCount" prompts. */
+static void InitKbInfoResponseHarness(ChannelOpenHarness* harness,
+        byte* in, word32 inSz, word32 promptCount)
+{
+    InitChannelOpenHarness(harness, in, inSz);
+    harness->ssh->acceptState = ACCEPT_CLIENT_USERAUTH_REQUEST_DONE;
+    harness->ssh->authId = ID_USERAUTH_KEYBOARD;
+    harness->ssh->kbAuth.promptCount = promptCount;
+}
+
+/* Post-condition for a rejected INFO_RESPONSE: the transport survives and the
+ * server's reply is a USERAUTH_FAILURE rather than a teardown. */
+static void AssertKbInfoResponseSentFailure(const ChannelOpenHarness* harness,
+        int ret)
+{
+    AssertIntEQ(ret, WS_SUCCESS);
+    AssertIntEQ(harness->io.inOff, harness->io.inSz);
+    AssertTrue(harness->io.outSz > 0);
+    AssertIntEQ(ParseMsgId(harness->io.out, harness->io.outSz),
+            MSGID_USERAUTH_FAILURE);
+}
+
+/* A response count that disagrees with the server's prompt count must yield a
+ * USERAUTH_FAILURE, not a transport teardown (RFC 4256). */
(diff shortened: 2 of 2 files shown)


src/internal.c
```diff
@@ -7275,12 +7275,17 @@ static int DoUserAuthInfoResponse(WOLFSSH* ssh,
     if ((ret == WS_SUCCESS) &&
         (ssh->kbAuth.promptCount != kb->responseCount)) {
         WLOG(WS_LOG_DEBUG, "DUARKB: Invalid number of responses received");
+        /* Respond with USERAUTH_FAILURE rather than
```
aidangarske: **🟡 [Medium] Malformed keyboard response failure leaves challenge state active**
*💡 SUGGEST*  `bug`

The new count-mismatch and too-many branches now turn a malformed `INFO_RESPONSE` into a successful `SendUserAuthFailure()` path, but the pending keyboard-interactive state is left intact. `ssh->authId` remains `ID_USERAUTH_KEYBOARD` and `ssh->kbAuth.promptCount` remains the old challenge count, so after receiving `USERAUTH_FAILURE` a malformed client can send another `MSGID_USERAUTH_INFO_RESPONSE` without first sending a new `USERAUTH_REQUEST`. Before this PR these malformed count cases tore down the connection, so this retry-without-new-request path is introduced by the change. That is at least protocol-state drift from the intended retry flow described in the PR.

**Recommendation:** After sending `USERAUTH_FAILURE` for a rejected keyboard-interactive `INFO_RESPONSE`, clear the pending keyboard challenge state or add an explicit pending-challenge flag so a later `INFO_RESPONSE` is rejected unless a fresh `USERAUTH_REQUEST` caused a new `INFO_REQUEST`. Add a regression test that sends a malformed `INFO_RESPONSE`, observes `USERAUTH_FAILURE`, then sends a second `INFO_RESPONSE` without a new `USERAUTH_REQUEST` and verifies it is not accepted as an answer to the old challenge.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Malformed keyboard response failure leaves challenge state active** — `src/internal.c:7278-7385`

*Review generated by Skoll.*

src/internal.c
yosuke-wolfssl: Thanks, and sorry for late response.
I rebased onto master, which bounds this.

212e976c ("Add server-side max failed userauth attempt limit") landed after this
review and rewired the failure gate to `SendUserAuthFailureCount(ssh, partialSuccess, 1)`,
so every `USERAUTH_FAILURE` this PR triggers is charged against `maxAuthAttempts`
(default 6) and disconnects at the cap. That commit also added the explicit
pending-challenge flag suggested here, `ssh->kbSetupPending`.

`authId` and `kbAuth.promptCount` do stay set, but that is pre-existing behavior,
not new: the ordinary rejected-password path already does
`authFailure = 1; ret = WS_SUCCESS` and leaves the same state
(`src/internal.c:8299-8303`). This PR just makes the malformed-count case behave
like the wrong-password case.

Rejecting a stray `INFO_RESPONSE` outright would change that long-standing path for
every keyboard-interactive failure, so I'd rather do it in its own PR — happy to
open one if you agree.

tests/regress.c
```diff
@@ -340,6 +340,71 @@ static void FreeChannelOpenHarness(ChannelOpenHarness* harness)
...
+{
+    InitChannelOpenHarness(harness, in, inSz);
+    harness->ssh->acceptState = ACCEPT_CLIENT_USERAUTH_REQUEST_DONE;
+    harness->ssh->authId = ID_USERAUTH_KEYBOARD;
+    harness->ssh->kbAuth.promptCount = promptCount;
+}
+
+/* Post-condition for a rejected INFO_RESPONSE: the transport survives and the
+ * server's reply is a USERAUTH_FAILURE rather than a teardown. */
+static void AssertKbInfoResponseSentFailure(const ChannelOpenHarness* harness,
```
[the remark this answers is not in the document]
yosuke-wolfssl: Checked this against the code and it doesn't reproduce — the High is a false positive.

`DoPacket()` does leave `ssh->inputBuffer.idx` short when `*idx` stays 0, but
`DoReceive()` then calls `ShrinkBuffer(&ssh->inputBuffer, 1)` on every consumed
packet, and `forcedFree = 1` sets `length = 0, idx = 0` and scrubs the buffer.
The short index never reaches the next packet.

Added `TestKbInfoResponseMismatchKeepsFraming` (tests/regress.c) to settle it:
two back-to-back malformed `INFO_RESPONSE`s, `DoReceive()` twice, asserting the
second still parses and draws its own `USERAUTH_FAILURE`. Results:

| `*idx = len` | `ShrinkBuffer` forcedFree | Test |
|---|---|---|
| absent (current) | `1` (current) | PASS |
| present | `1` | PASS |
| present | `0` | PASS |
| absent | `0` | FAIL (`-1001` on the second `DoReceive`) |

The last row confirms the test detects the desync when it is reachable; the first
two confirm it isn't reachable as the receive path stands. `*idx = len` was
removed earlier in this PR as an unobservable no-op, and this measurement is why
it stays removed.

That also covers the Low finding: the test now asserts parser state after the
mismatch by feeding a follow-up packet.

yosuke-wolfssl:
Hello @aidangarske, @ejohnstown ,
Sorry for being late. Could you review this again ?

ejohnstown (approved):

PR #1071, opened 2026-06-29
SSHD/Echoserver: Fix memory leaks and public-key lookup

stenslae:
Fixed: Memory leaks, multiple public-key lookups per user in the echoserver, and authentication privilege errors.

- **internal.c**: Readability refactor for RSA/ECC keys.
- **auth.c**: Freed `promptData.prompts` and `promptData.promptLengths` on allocation failure pathways.
- **echoserver.c**: Fixed a copy-paste error and added cleanups on allocations failure.
 - **Key Hash Lookup Correctness**: Updated `wsUserAuth` in echoserver.c to traverse all
  registered public keys for a matching user name instead of immediately returning an invalid
  key error on the first hash mismatch.
  - **Stack Buffer Cleanup**: Freed `keyLoadBuf` under `WOLFSSH_SMALL_STACK` configurations for all early exit paths inside `echoserver_test`.
  - **auth.c**: Returns success immediately if privilege separation is explicitly disabled.
  - **auth.c**: Limits changing target user/group identifiers to environments where privilege separation or sandboxing is active.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -95,6 +95,10 @@
 #if defined(WOLFSSHD_UNIT_TEST) && !defined(_WIN32)
 int (*wsshd_setregid_cb)(WGID_T, WGID_T) = setregid;
 int (*wsshd_setreuid_cb)(WUID_T, WUID_T) = setreuid;
+int (*wsshd_setegid_cb)(WGID_T) = setegid;
+int (*wsshd_seteuid_cb)(WUID_T) = seteuid;
+struct passwd* (*wsshd_getpwnam_cb)(const char*) = getpwnam;
+#define getpwnam wsshd_getpwnam_cb
 #endif
 
 struct WOLFSSHD_AUTH {
@@ -1835,13 +1839,8 @@ static int SetDefaultPublicKeyCheck(WOLFSSHD_AUTH* auth)
     return ret;
 }
 
-#ifndef WOLFSSH_SSHD_USER
-    #define WOLFSSH_SSHD_USER sshd
-#endif
-#define WOLFSSH_USER_GET_STRING(x) #x
-#define WOLFSSH_USER_STRING(x) WOLFSSH_USER_GET_STRING(x)
 
-static int SetDefualtUserID(WOLFSSHD_AUTH* auth)
+static int SetDefaultUserID(WOLFSSHD_AUTH* auth)
 {
 #ifdef _WIN32
     /* TODO: Implement for Windows. */
@@ -1849,6 +1848,15 @@ static int SetDefualtUserID(WOLFSSHD_AUTH* auth)
 #else
     struct passwd* pwInfo;
     int ret = WS_SUCCESS;
+
+    if (wolfSSHD_ConfigGetPrivilegeSeparation(auth->conf) ==
+            WOLFSSHD_PRIV_OFF) {
+        auth->gid  = getgid();
+        auth->uid  = getuid();
+        auth->sGid = auth->gid;
+        auth->sUid = auth->uid;
+        return WS_SUCCESS;
+    }
 
     pwInfo = getpwnam(WOLFSSH_USER_STRING(WOLFSSH_SSHD_USER));
     if (pwInfo == NULL) {
@@ -1910,7 +1918,7 @@ WOLFSSHD_AUTH* wolfSSHD_AuthCreateUser(void* heap, const WOLFSSHD_CONFIG* conf)
         }
 
         if (ret == WS_SUCCESS) {
-            ret = SetDefualtUserID(auth);
+            ret = SetDefaultUserID(auth);
             if (ret != WS_SUCCESS) {
                 wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error setting default "
                     "user ID.");
@@ -1942,23 +1950,36 @@ int wolfSSHD_AuthFreeUser(WOLFSSHD_AUTH* auth)
 /* return WS_SUCCESS on success */
 int wolfSSHD_AuthRaisePermissions(WOLFSSHD_AUTH* auth)
 {
-    int ret = 0;
+    int ret = WS_SUCCESS;
 
-    wolfSSH_Log(WS_LOG_INFO, "[SSHD] Attempting to raise permissions level");
 #ifndef WIN32
-    if (auth) {
+    byte flag = 0;
+
+    if (auth == NULL) {
+        return WS_BAD_ARGUMENT;
+    }
+
+    flag = wolfSSHD_ConfigGetPrivilegeSeparation(auth->conf);
+    if (flag == WOLFSSHD_PRIV_SEPARAT || flag == WOLFSSHD_PRIV_SANDBOX) {
+        wolfSSH_Log(WS_LOG_INFO,
+            "[SSHD] Attempting to raise permissions level");
+#ifdef WOLFSSHD_UNIT_TEST
+        if (wsshd_setegid_cb(auth->sGid) != 0) {
+#else
         if (setegid(auth->sGid) != 0) {
+#endif
             wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error raising gid");
             ret = WS_FATAL_ERROR;
         }
 
-        if (seteuid(auth->sUid) != 0) {
+#ifdef WOLFSSHD_UNIT_TEST
+        if (ret == WS_SUCCESS && wsshd_seteuid_cb(auth->sUid) != 0) {
+#else
+        if (ret == WS_SUCCESS && seteuid(auth->sUid) != 0) {
+#endif
             wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Error raising uid");
             ret = WS_FATAL_ERROR;
         }
-    }
-    else {
-        ret = WS_BAD_ARGUMENT;
     }
 #endif
 
--- a/apps/wolfsshd/test/test_configuration.c
+++ b/apps/wolfsshd/test/test_configuration.c
@@ -1559,6 +1559,282 @@ static int test_AuthReducePermissionsUser_uid_fail(void)
     wsshd_setreuid_cb = savedReuid;
     return ret;
 }
+
+static WGID_T s_setegid_arg;
+static WUID_T s_seteuid_arg;
+static int    s_setegid_ret;
+static int    s_seteuid_ret;
+static int    s_setegid_called;
+static int    s_seteuid_called;
+
+static int stub_setegid(WGID_T egid)
+{
+    s_setegid_called = 1;
+    s_setegid_arg    = egid;
+    return s_setegid_ret;
+}
+
+static int stub_seteuid(WUID_T euid)
+{
+    s_seteuid_called = 1;
+    s_seteuid_arg    = euid;
+    return s_seteuid_ret;
+}
+
+static void InstallPrivRaiseStubs(int egidRet, int euidRet,
+    int (**savedEgid)(WGID_T), int (**savedEuid)(WUID_T))
+{
+    *savedEgid       = wsshd_setegid_cb;
+    *savedEuid       = wsshd_seteuid_cb;
+    wsshd_setegid_cb = stub_setegid;
+    wsshd_seteuid_cb = stub_seteuid;
+    s_setegid_ret    = egidRet;
+    s_seteuid_ret    = euidRet;
+    s_setegid_called = 0;
+    s_seteuid_called = 0;
+    s_setegid_arg    = 0;
+    s_seteuid_arg    = 0;
+}
+
+/* Synthetic "sshd" account used so privilege-separation tests don't depend
+ * on the host actually having an sshd system user configured. */
+static struct passwd stub_sshd_pw;
+
+static struct passwd* stub_getpwnam(const char* name)
+{
+    if (name != NULL && WSTRCMP(name, WOLFSSH_USER_STRING(WOLFSSH_SSHD_USER)) == 0) {
+        WMEMSET(&stub_sshd_pw, 0, sizeof(stub_sshd_pw));
+        stub_sshd_pw.pw_name = (char*)WOLFSSH_USER_STRING(WOLFSSH_SSHD_USER);
+        stub_sshd_pw.pw_uid  = 1000;
+        stub_sshd_pw.pw_gid  = 1000;
+        return &stub_sshd_pw;
+    }
+    return NULL;
+}
+
+static void InstallGetpwnamStub(struct passwd* (**savedGetpwnam)(const char*))
+{
+    *savedGetpwnam   = wsshd_getpwnam_cb;
+    wsshd_getpwnam_cb = stub_getpwnam;
+}
+
+/* UsePrivilegeSeparation no must let SetDefaultUserID succeed without a
+ * configured sshd system user, since no uid/gid switching will ever happen. */
+static int test_AuthCreateUser_privSepOff(void)
+{
+    int ret = WS_SUCCESS;
+    WOLFSSHD_CONFIG* conf;
+    WOLFSSHD_AUTH* auth;
+    static const char line[] = "UsePrivilegeSeparation no";
+
+    conf = wolfSSHD_ConfigNew(NULL);
+    if (conf == NULL) {
+        return WS_MEMORY_E;
+    }
+
+    if (ParseConfigLine(&conf, line, (int)WSTRLEN(line), 0) != WS_SUCCESS) {
+        ret = WS_FATAL_ERROR;
+    }
+
+    if (ret == WS_SUCCESS) {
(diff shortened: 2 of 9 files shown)


examples/echoserver/echoserver.c
```diff
@@ -3137,7 +3144,8 @@ THREAD_RETURN WOLFSSH_THREAD echoserver_test(void* args)
         if (tpmHostKeyPath != NULL) {
             if (EchoserverInitTpmHostKey(ctx, tpmHostKeyPath,
                     ECHOSERVER_TPM_KEY_AUTH_DEFAULT) != 0) {
-                ES_ERROR("Couldn't load TPM host key from %s.\n", tpmHostKeyPath);
+                ES_ERROR("Couldn't load TPM host key from %s.\n",
```
stenslae, line 3266: Freed keyLoadBuf on the failure path.

examples/echoserver/echoserver.c
```diff
@@ -3212,17 +3247,26 @@ THREAD_RETURN WOLFSSH_THREAD echoserver_test(void* args)
...
                 ES_ERROR("WMALLOC failed\n");
             }
             load_file(caCert, certBuf, &certBufSz);
             ret = wolfSSH_CTX_AddRootCert_buffer(ctx, certBuf, certBufSz,
                     WOLFSSH_FORMAT_PEM);
             if (ret != 0) {
+                #ifdef WOLFSSH_SMALL_STACK
+                WFREE(keyLoadBuf, NULL, 0);
+                #endif
                 ES_ERROR("Couldn't add root cert\n");
```
stenslae, line 3412: Added a free for certBuf

apps/wolfsshd/auth.c
```diff
@@ -1850,6 +1850,11 @@ static int SetDefualtUserID(WOLFSSHD_AUTH* auth)
     struct passwd* pwInfo;
     int ret = WS_SUCCESS;
 
+    if (wolfSSHD_ConfigGetPrivilegeSeparation(auth->conf) ==
```
aidangarske, line 1852: **🟡 [Medium] Privilege-separation permission behavior lacks regression tests**
*💡 SUGGEST*  `test`

The PR changes daemon privilege behavior by returning success from `SetDefaultUserID` when `UsePrivilegeSeparation no` is configured, and by making `wolfSSHD_AuthRaisePermissions` a no-op unless the config is `WOLFSSHD_PRIV_SEPARAT` or `WOLFSSHD_PRIV_SANDBOX`. Grep found config parsing tests and reduce-permission tests, but no tests that exercise `wolfSSHD_AuthRaisePermissions` or auth creation under the new privilege modes. This is a user-visible daemon startup/auth behavior change, so it should have regression coverage.

**Suggestion:**
```suggestion
Add unit coverage for `UsePrivilegeSeparation no` returning success without requiring the configured sshd user, and for `wolfSSHD_AuthRaisePermissions` calling the uid/gid restore path only for `yes` and `sandbox` modes.
```

examples/echoserver/echoserver.c
```diff
@@ -2676,12 +2675,12 @@ static int wsUserAuth(byte authType,
             authData->type == map->type) {
 
             if (authData->type == WOLFSSH_USERAUTH_PUBLICKEY) {
```
aidangarske, line 2677: **🟡 [Medium] Multiple public-key lookup fix has no regression coverage**
*💡 SUGGEST*  `test`

The PR fixes `wsUserAuth` so a hash mismatch for the first public key registered to a user does not immediately reject later keys. Existing public-key tests use their own `serverPubkeyUserAuth` callback and do not exercise the echoserver `PwMapList` traversal. The same fix is mirrored in the ESP-IDF echoserver, so a regression could silently return if this path is refactored.

**Suggestion:**
```suggestion
Add a regression scenario with two public-key entries for the same username where the first key hash mismatches and the second matches.
```

apps/wolfsshd/auth.c
```diff
@@ -1942,24 +1947,31 @@ int wolfSSHD_AuthFreeUser(WOLFSSHD_AUTH* auth)
 /* return WS_SUCCESS on success */
 int wolfSSHD_AuthRaisePermissions(WOLFSSHD_AUTH* auth)
 {
-    int ret = 0;
+    int ret = WS_SUCCESS;
 
-    wolfSSH_Log(WS_LOG_INFO, "[SSHD] Attempting to raise permissions level");
 #ifndef WIN32
```
aidangarske, line 1955: **🔵 [Low] New bare scope block violates local C convention**
*🔧 NIT*  `convention`

The PR adds a standalone scope block under `#ifndef WIN32` only to limit `flag`. The project coding guidance treats newly added bare scope blocks as a convention issue; this can be written with the declaration inside the existing preprocessor region instead.

**Suggestion:**
```suggestion
#ifndef WIN32
    byte flag = 0;

    if (auth == NULL) {
        return WS_BAD_ARGUMENT;
    }

    flag = wolfSSHD_ConfigGetPrivilegeSeparation(auth->conf);
    if (flag == WOLFSSHD_PRIV_SEPARAT || flag == WOLFSSHD_PRIV_SANDBOX) {
        ...
    }
#endif
```

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Privilege-separation permission behavior lacks regression tests** — `apps/wolfsshd/auth.c:1853-1961`
- [Medium] **Multiple public-key lookup fix has no regression coverage** — `examples/echoserver/echoserver.c:2677-2727`
- [Low] **New bare scope block violates local C convention** — `apps/wolfsshd/auth.c:1952-1974`

*Review generated by Skoll.*

apps/wolfsshd/auth.c
stenslae: Added unit coverage for the UsePrivilegeSeparation no case to ensure it returns success without requiring the configured sshd user. I also added tests verifying that wolfSSHD_AuthRaisePermissions only triggers the uid/gid restore path for the yes and sandbox modes as expected.

examples/echoserver/echoserver.c
stenslae: Added a new regression test to validate that if the first key's hash mismatches, it successfully falls back to and matches against the second key without prematurely failing.

apps/wolfsshd/auth.c
stenslae: Fixed the styling.

aidangarske (commented):
## Skoll Code Review

### Posted findings
- [Low] **Mixed pubKeyFmtId/pubKeyId within a single if/else-if free chain is fragile** — `src/internal.c:14597-14637`
- [Low] **New privilege-separation-ON tests silently skip when no "sshd" system user exists** — `apps/wolfsshd/test/test_configuration.c:1679-1822`
- [Low] **New error paths free host private key material without zeroization** — `examples/echoserver/echoserver.c:3275-3323`
- [Info] **internal.c "memory leak fix" is a semantic no-op; no leak existed or was fixed** — `src/internal.c:14597-14619`

src/internal.c
```diff
@@ -14595,24 +14595,22 @@ int SendKexDhReply(WOLFSSH* ssh)
     }
 
     if (sigKeyBlock_ptr != NULL) {
```
aidangarske, line 14599: **🔵 [Low] Mixed pubKeyFmtId/pubKeyId within a single if/else-if free chain is fragile** · _Logic_

Verified by reading the post-change chain at src/internal.c:14597-14637. The PR converted the RSA and ECDSA branches of the key-free chain to test `pubKeyFmtId`, but left the ED25519 and ML-DSA branches in the same `if / else if` chain testing `pubKeyId`. The chain is currently correct only because `pubKeyFmtId == pubKeyId` for every non-RSA_SHA2 id. This coupling is invisible at the call site: the free chain must mirror the key-init switch in `SendKexGetSigningKey`, which dispatches on `pubKeyId` (line 12838), so the free logic now keys off a different field than the init logic. If a future change adds any further `pubKeyFmtId` remap (for example collapsing `ID_X509V3_MLDSA*` to `ID_MLDSA*`, mirroring what is already done for RSA), the RSA/ECDSA branches would silently start matching different ids than the ED25519/ML-DSA branches, producing either a leaked key or a free of an uninitialized union member. This is a maintainability/latent-risk issue, not a live defect.

**Fix:** Use one field consistently across the whole chain. Since the union member is chosen by the init switch in SendKexGetSigningKey (which dispatches on pubKeyId), keying the free chain on pubKeyId throughout is the lower-risk option; otherwise convert the ED25519/ML-DSA branches to pubKeyFmtId as well and add a comment stating the free chain must mirror the init switch.

apps/wolfsshd/test/test_configuration.c
```diff
@@ -1559,6 +1559,266 @@ static int test_AuthReducePermissionsUser_uid_fail(void)
...
+
+    wolfSSHD_ConfigFree(conf);
+    return ret;
+}
+
+/* With privilege separation on, wolfSSHD_AuthRaisePermissions restores the
+ * uid/gid the process started with (captured at AuthCreateUser time). Skipped
+ * when the environment has no "sshd" system user, since SetDefaultUserID
+ * requires one to succeed for this mode. */
+static int test_AuthRaisePermissions_separateCallsSyscalls(void)
```
aidangarske, line 1701: **🔵 [Low] New privilege-separation-ON tests silently skip when no "sshd" system user exists** · _Missing Tests_

Verified: the guard at test_configuration.c:1687-1691 returns WS_SUCCESS when getpwnam("sshd") is NULL, and the auth.c diff confirms SetDefaultUserID only consults getpwnam() on the privilege-separation-ON path (the PRIV_OFF path now short-circuits to getgid()/getuid()). Three of the six new tests — the only ones that exercise the security-relevant branch where privilege separation is ON and setegid/seteuid are actually invoked — bail out with `return WS_SUCCESS` when `getpwnam("sshd")` returns NULL, because they cannot construct an auth object without that account. Many CI images and minimal containers have no `sshd` system account, so in those environments all three tests report PASS while asserting nothing. The result is that the assertions guarding the changed code (that Raise calls setegid/seteuid with the saved uid/gid, that a setegid failure short-circuits seteuid via the new `ret == WS_SUCCESS &&` guard, and that a seteuid failure is reported) may never actually execute in CI, while the suite still shows green. The tests do emit a Log() skip line, so this is not fully silent, but the pass/fail signal does not distinguish "verified" from "not run". Only `test_AuthRaisePermissions_offSkipsSyscalls` and `test_AuthCreateUser_privSepOff` (the priv-sep-OFF paths) run unconditionally.

**Fix:** Decouple these tests from the host's user database so they always run: either add a `wsshd_getpwnam_cb` stub (matching the existing wsshd_setegid_cb/wsshd_seteuid_cb pattern this PR already adds to auth.c/auth.h) so SetDefaultUserID can be driven with a synthetic passwd entry, or expose a WOLFSSHD_UNIT_TEST-only setter for auth->sUid/sGid. Failing that, have the runner report skipped tests distinctly from passing ones so a green suite does not imply these branches were verified.

examples/echoserver/echoserver.c
```diff
@@ -3264,10 +3272,16 @@ THREAD_RETURN WOLFSSH_THREAD echoserver_test(void* args)
         if (loadDefaultHostKeys) {
             bufSz = load_key(peerEcc, keyLoadBuf, bufSz);
             if (bufSz == 0) {
+                #ifdef WOLFSSH_SMALL_STACK
```
aidangarske, line 3276: **🔵 [Low] New error paths free host private key material without zeroization** · _Zeroization_

Verified against the diff and the ES_ERROR macro definition (echoserver.c:2853-2865), which ends in WOLFSSL_RETURN_FROM_THREAD(0) — so each added WFREE is on a genuine early-return path and does execute. Under WOLFSSH_SMALL_STACK, `keyLoadBuf` is a heap buffer that `load_key()` / `load_key_ed25519()` fill with the server's ASN.1 *private* host key. The PR adds `WFREE(keyLoadBuf, NULL, 0)` on several new error paths reached after the buffer already holds private key bytes — most clearly the "Couldn't use first/second/Ed25519 key buffer" paths, where `load_key()` succeeded and only `wolfSSH_CTX_UsePrivateKey_buffer()` failed. The buffer is released back to the allocator with the key material intact rather than being cleared with `ForceZero()` first. Mitigating context: this is an example/test program rather than library code, the pre-existing success-path free at line 3439 has the same gap, and reaching these paths requires a key-load failure, so this is a defense-in-depth gap rather than an exploitable flaw. It is listed because the diff introduces additional unzeroized frees of key material on paths that previously leaked the allocation instead.

**Fix:** Add `ForceZero(keyLoadBuf, EXAMPLE_KEYLOAD_BUFFER_SZ);` immediately before each `WFREE(keyLoadBuf, ...)` (including the existing block-end free at line 3439, and the ESP-IDF copy), or factor the free into a small helper that zeroizes then frees so future exit paths inherit the behavior.

src/internal.c
aidangarske, line 14599: **⚪ [Info] internal.c "memory leak fix" is a semantic no-op; no leak existed or was fixed** · _Logic_

Verified against the diff (HEAD~1..HEAD) and the assignment site. The PR description states this hunk "Fixed a memory leak where RSA/ECC keys in signature blocks were leaked if they matched X.509 format IDs instead of standard SSH format IDs." The change is strictly behavior-preserving. `pubKeyFmtId` is assigned in exactly one place (src/internal.c:14397-14401): `pubKeyFmtId = pubKeyId`, then remapped to `ID_SSH_RSA` only when `pubKeyId` is `ID_RSA_SHA2_256` or `ID_RSA_SHA2_512`. No other code assigns `pubKeyFmtId`, and `pubKeyId` is never reassigned after 14392. Therefore: BEFORE, the RSA branch matched `pubKeyId` in {ID_SSH_RSA, ID_RSA_SHA2_256, ID_RSA_SHA2_512, ID_X509V3_SSH_RSA}. AFTER, `pubKeyFmtId == ID_SSH_RSA` holds exactly when `pubKeyId` is in {ID_SSH_RSA, ID_RSA_SHA2_256, ID_RSA_SHA2_512}, plus `pubKeyFmtId == ID_X509V3_SSH_RSA` when `pubKeyId == ID_X509V3_SSH_RSA` — an identical set. For ECDSA, `pubKeyFmtId == pubKeyId` for every ECDSA and X509V3_ECDSA id (no remap applies), so that branch is identical too. The X.509 ids were already covered by the BEFORE code under the same `#ifdef WOLFSSH_CERTS` guard. The refactor is harmless and arguably clearer, but it fixes no leak. This is worth flagging so a maintainer does not close a real leak report believing it was addressed here.

**Fix:** Correct the PR description: this hunk is a readability refactor, not a leak fix. If an actual key leak was observed, capture the reproducing configuration — the fix is not in this hunk and the underlying issue may still be open.

stenslae: I didnt see any memory leaks in internal.c related to the pubKeyId... Meant to say the leak in the echoservers. Fixed pr summary.

stenslae: All branches now consistently use pubKeyFmtId.  Switching them to pubKeyId would make the RSA branch miss ID_RSA_SHA2_256/512.

apps/wolfsshd/test/test_configuration.c
stenslae: Fixed as suggested

examples/echoserver/echoserver.c
stenslae: Added inline

tests/testsuite.c
```diff
@@ -150,6 +150,73 @@ static void wolfSSH_EchoTest(void)
...
+        WSNPRINTF(cA[clientArgc++], ARGLEN, "%d", ready.port);
+    #endif
+
+    clientArgs.argc = clientArgc;
+    clientArgs.argv = clientArgv;
+    clientArgs.return_code = EXIT_SUCCESS;
+    clientArgs.signal = &ready;
+    clientArgs.user_auth = NULL;
+
+    client_test(&clientArgs);
```
stenslae, line 209: More test assertions added

apps/wolfsshd/test/test_configuration.c
```diff
@@ -1559,6 +1559,282 @@ static int test_AuthReducePermissionsUser_uid_fail(void)
...
+
+/* UsePrivilegeSeparation no must let SetDefaultUserID succeed without a
+ * configured sshd system user, since no uid/gid switching will ever happen. */
+static int test_AuthCreateUser_privSepOff(void)
+{
+    int ret = WS_SUCCESS;
+    WOLFSSHD_CONFIG* conf;
+    WOLFSSHD_AUTH* auth;
+    static const char line[] = "UsePrivilegeSeparation no";
+
```
stenslae, line 1629: Moved where the macros get declared so i could use them

aidangarske (approved):
Looks good over to @ejohnstown

ejohnstown (approved):

PR #1072, opened 2026-06-29
SSHD: OpenSSH Certificate Auth Support

stenslae:
Fixed memory leaks and verification logic in OpenSSH Certificate CA authentication in SSHD

  - Corrected Authentication Logic: Enforced a  WSSHD_AUTH_FAILURE  return code when the
  verification loop completes without finding a matching trusted CA fingerprint.
  - Resolved Memory Leaks
  - Adopted Secure I/O: Migrated all raw file system operations to WFOPEN ,  WFGETS ,  WFCLOSE.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -835,7 +835,6 @@ static int SearchForPubKey(const char* path, const char* authKeysFile,
     if (f != WBADFILE) {
         WFCLOSE(NULL, f);
     }
-
     if (lineBuf != NULL) {
         WFREE(lineBuf, NULL, DYNTYPE_BUFFER);
     }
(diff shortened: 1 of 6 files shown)


apps/wolfsshd/auth.c
```diff
@@ -2249,4 +2274,19 @@ WOLFSSHD_CONFIG* wolfSSHD_AuthGetUserConf(const WOLFSSHD_AUTH* auth,
...
+int wolfSSHD_TestCheckOsshCertCa(const byte* caKeyHash,
+                                  word32 caKeyHashSz,
+                                  const char* caKeysFile)
+{
+    WS_UserAuthData_PublicKey pk;
+    WMEMSET(&pk, 0, sizeof(pk));
+    pk.isOsshCert = 1;
+    pk.caKeyHash  = caKeyHash;
+    pk.caKeyHashSz = caKeyHashSz;
+    return CheckPublicKeyUnix(NULL, &pk, caKeysFile, NULL, NULL);
```
stenslae: Added !_WIN32 to the header guard around the OSSH cert test check.

PR #1073, opened 2026-06-30
Gate wolfsshd PermitRootLogin on resolved UID 0

yosuke-wolfssl:
## Summary

`wolfsshd` enforced `PermitRootLogin` only when the SSH username was the literal
string `"root"`. The privileged-account decision (`DoCheckUser`) used
`XSTRCMP(usr, "root") == 0`, while the user-existence check only verified
`getpwnam() != NULL` and the post-auth privilege drop used `pw_uid`. So any
UID-0 account with a different name (a BSD-style `toor`, or any admin-created
superuser alias) bypassed `PermitRootLogin no` yet still got a real UID-0
session. The username is attacker-controlled in `SSH_MSG_USERAUTH_REQUEST`, so
the non-`root` UID-0 account can be targeted deliberately. OpenSSH gates on the
resolved `pw->pw_uid == 0` for exactly this reason.

Addressed by f_6516.

## Fix

In `DoCheckUser` (`apps/wolfsshd/auth.c`):

- **Unix:** resolve the account with `getpwnam` and enforce `PermitRootLogin`
  whenever `pw_uid == 0`, **or** the name is literally `root`. The uid check
  closes the alias bypass; the literal `root` term is retained as
  defense-in-depth so a transient `getpwnam` failure cannot fail open for the
  name `root` (the existence check does its own independent `getpwnam`, so the
  two lookups can disagree within one connection).
- **Windows:** no UID-0 concept and no logon token at this pre-auth stage
  (`LogonUserExExW`/`LsaLogonUser` both create the token *as part of*
  authentication, after `DoCheckUser`). The literal-`root` check is kept so the
  prior behavior is not silently dropped. A proper Windows equivalent
  (`CheckTokenMembership` against `BUILTIN\Administrators`) belongs in a
  post-auth gate and is noted as future work.

```c
#ifndef _WIN32
    pwInfo = getpwnam(usr);
    if ((pwInfo != NULL && pwInfo->pw_uid == 0) || XSTRCMP(usr, "root") == 0) {
        isRoot = 1;
    }
#else
    if (XSTRCMP(usr, "root") == 0) {
        isRoot = 1;
    }
#endif

    if (isRoot == 1) {
        usrConf = wolfSSHD_AuthGetUserConf(auth, usr, NULL, NULL, NULL, NULL,
                                           NULL);
        if (usrConf == NULL || wolfSSHD_ConfigGetPermitRoot(usrConf) == 0) {
            wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Login as root not permitted");
            ret = WOLFSSH_USERAUTH_REJECTED;
        }
    }
```

## Tests

New integration test `apps/wolfsshd/test/sshd_permitroot_test.sh` (registered in
`run_all_sshd_tests.sh`). The rejection happens pre-auth, so the test needs no
valid credentials — it connects and inspects the daemon log:

- **root, `PermitRootLogin no`** → expects `Login as root not permitted`.
- **root, `PermitRootLogin yes`** (positive control) → message absent, request
  still reached `DoCheckUser`.
- **UID-0 alias** (`useradd -o -u 0 -M -s nologin`), `PermitRootLogin no` →
  expects rejection, proving the check is uid-based not name-based. Hard-guarded
  to Linux + root + `useradd`/`userdel`, with `trap` cleanup on EXIT and
  INT/TERM/HUP; skipped elsewhere. CI runs the suite under `sudo` on
  `ubuntu-latest`, so this layer executes there.

The coverage note in `apps/wolfsshd/test/test_configuration.c` is updated to
reflect the uid-based check.

## Verification

- `./configure --enable-all && make` — clean.
- `apps/wolfsshd/test/test_configuration` — passes.
- `sshd_permitroot_test.sh` — `bash -n` clean; modeled on the existing
  `sshd_empty_password_test.sh`; exercised by the Linux CI harness.

## Notes / follow-up

- Windows `PermitRootLogin` still only matches the literal name `root`; a real
  admin-membership check (`CheckTokenMembership`) must be a post-auth gate
  because the logon token does not exist until authentication completes.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -1455,21 +1455,34 @@ static int DoCheckUser(const char* usr, WOLFSSHD_AUTH* auth)
 {
     int ret = WOLFSSH_USERAUTH_SUCCESS;
     int rc;
+    int isRoot = 0;
     WOLFSSHD_CONFIG* usrConf;
+#ifndef _WIN32
+    struct passwd* pwInfo;
+#endif
 
     wolfSSH_Log(WS_LOG_INFO, "[SSHD] Checking user name %s", usr);
 
+#ifndef _WIN32
+    /* PermitRootLogin covers every uid 0 account (so an alias like "toor"
+     * cannot bypass it) and the literal name "root", so a transient getpwnam
+     * failure cannot skip the check for root. */
+    pwInfo = getpwnam(usr);
+    if ((pwInfo != NULL && pwInfo->pw_uid == 0) || XSTRCMP(usr, "root") == 0) {
+        isRoot = 1;
+    }
+#else
+    /* No uid 0 on Windows and no logon token yet at this pre-auth stage, so
+     * fall back to the literal name; a token based Administrators membership
+     * check would belong after authentication. */
     if (XSTRCMP(usr, "root") == 0) {
-        /* Resolve the per-user configuration so that a Match block override of
-         * PermitRootLogin is honored, rather than only consulting the global
-         * config node. The lookup is only needed for the root user, so it is
-         * scoped to this branch.
-         *
-         * A NULL return means the root user's configuration could not be
-         * resolved (e.g. group set enumeration failed inside
-         * wolfSSHD_AuthGetUserConf). Fail closed and reject the login in that
-         * case rather than falling back to the global node, since denying root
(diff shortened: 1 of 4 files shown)


aidangarske (dismissed):
Looks good.

ejohnstown (approved):

PR #1076, opened 2026-06-30
wolfterm: bound escBuf saves and clear esc state on OSC resume

yosuke-wolfssl:
## Overview

Hardens the `escBuf` partial-sequence reassembly in `wolfSSH_DoControlSeq` and
fixes an escape-state leak in the `wolfSSH_ConvertConsole` OSC resume path
(`src/wolfterm.c`). Windows-only console handling (`USE_WINDOWS_API`).

## Changes

1. **Tighten both `escBuf` save branches.** When a CSI sequence's arguments run
   to the end of a buffer with no command char, the partial bytes are stashed in
   `ssh->escBuf` for the next call. `escBuf` is `WOLFSSL_MAX_ESCBUF` bytes, so a
   save of exactly that length fills it with no room for continued state. Both
   save branches now reject `bufSz - *idx >= WOLFSSL_MAX_ESCBUF` before the
   `WMEMCPY` and log via `WLOG(WS_LOG_ERROR, ...)`, matching the boundary
   semantics of the resume check and keeping the two branches consistent.

2. **Guard the initial-parse CSI save branch.** The initial-parse path
   (`escState` was `WC_ESC_NONE`, buffer opens with `ESC [`) copied into `escBuf`
   with no capacity bound. In the default build `getArgs` caps the advance at
   `WOLFSSH_MAX_CONSOLE_ARGS` (16) < `WOLFSSL_MAX_ESCBUF` (19), so the copy was
   already safe; the guard is defense-in-depth for a custom
   `WOLFSSH_MAX_CONSOLE_ARGS > WOLFSSL_MAX_ESCBUF` build.

3. **Clear escape state on OSC resume completion.** When a split OSC sequence
   completes via the resume handler in `wolfSSH_ConvertConsole`, reset
   `escState = WC_ESC_NONE` (not just `escBufSz`). Without this, `escState`
   stayed `WS_ESC_OSC` and the next call re-entered OSC parsing on plain bytes
   instead of printing them.

## Testing

Extended `test_wolfSSH_ConvertConsole` in `tests/api.c`:

- **Split CSI across calls** (`ESC [` | `12` | `m`) — the continuation carries
  only argument bytes; verifies reassembly waits (`WS_WANT_READ`) and then
  completes (`WS_SUCCESS`) without reading past the buffer.
- **Trailing plain byte after completion** — a plain argument byte after the
  sequence finishes must be printed (`WS_SUCCESS`), not swallowed back into a
  new parse.
- **Single-buffer args-to-end** (`ESC [ 1 2`) — args run to the end of one
  buffer; verifies the partial save waits and completes on the next byte.

Verified on native Windows (Visual Studio Build Tools 2022, static Debug x64):

```
api-test EXITCODE=0
```

Local GCC preflight sweep (`--enable-all`, Zephyr defines, strncpy lint) clean.

### DIFF
--- a/src/wolfterm.c
+++ b/src/wolfterm.c
@@ -475,7 +475,8 @@ static int wolfSSH_DoControlSeq(WOLFSSH* ssh, WOLFSSH_HANDLE handle, byte* buf, 
             numArgs = getArgs(buf, bufSz, &i, args);
             if (i >= bufSz) {
                 /* save left overs for next call */
-                if (bufSz - *idx > WOLFSSL_MAX_ESCBUF) {
+                if (bufSz - *idx >= WOLFSSL_MAX_ESCBUF) {
+                    WLOG(WS_LOG_ERROR, "escBuf state exceeds capacity");
                     return WS_FATAL_ERROR;
                 }
                 WMEMCPY(ssh->escBuf, buf + *idx, bufSz - *idx);
@@ -491,6 +492,10 @@ static int wolfSSH_DoControlSeq(WOLFSSH* ssh, WOLFSSH_HANDLE handle, byte* buf, 
 
         if (i >= bufSz) {
             /* save left overs for next call */
+            if (bufSz - *idx >= WOLFSSL_MAX_ESCBUF) {
+                WLOG(WS_LOG_ERROR, "escBuf state exceeds capacity");
+                return WS_FATAL_ERROR;
+            }
             WMEMCPY(ssh->escBuf, buf + *idx, bufSz - *idx);
             ssh->escBufSz = bufSz - *idx;
             return WS_WANT_READ;
@@ -687,6 +692,7 @@ int wolfSSH_ConvertConsole(WOLFSSH* ssh, WOLFSSH_HANDLE handle, byte* buf,
             if (ret == WS_WANT_READ) {
                 return ret;
             }
+            ssh->escState = WC_ESC_NONE;
             ssh->escBufSz = 0;
             break;
 
--- a/tests/api.c
+++ b/tests/api.c
@@ -3370,6 +3370,18 @@ static byte osc_trunc_str[] = { /* "ESC ] 0 ; title" with no BEL terminator */
 static byte osc_full[] = { /* well formed "ESC ] 0 ; hi BEL" */
     0x1B, 0x5D, 0x30, 0x3B, 0x68, 0x69, 0x07
 };
+static byte csi_open[] = { /* "ESC [" with no args yet, escBufSz left at 0 */
+    0x1B, 0x5B
+};
+static byte csi_args[] = { /* pure args "12" with no command char */
+    0x31, 0x32
+};
+static byte csi_cmd[] = { /* command char 'm' completes the sequence */
+    0x6D
+};
+static byte csi_inline[] = { /* "ESC [ 1 2" args run to end of one buffer */
+    0x1B, 0x5B, 0x31, 0x32
+};
 #endif /* USE_WINDOWS_API */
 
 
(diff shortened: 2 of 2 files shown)


src/wolfterm.c
```diff
@@ -669,6 +685,9 @@ int wolfSSH_ConvertConsole(WOLFSSH* ssh, WOLFSSH_HANDLE handle, byte* buf,
             if (ret == WS_WANT_READ) {
```
aidangarske, line 683: **🟠 [High] CSI capacity errors are swallowed on resume**
*🚫 BLOCK*  `bug`

The PR adds `WS_FATAL_ERROR` returns in `wolfSSH_DoControlSeq` when saving a CSI continuation would exceed `escBuf`, but the resume path in `wolfSSH_ConvertConsole` treats every non-`WS_WANT_READ` result as a completed sequence. For example, after `ESC [` leaves `ssh->escState == WS_ESC_CSI` and `escBufSz == 0`, a following buffer of `WOLFSSL_MAX_ESCBUF` argument bytes hits the new `bufSz - *idx >= WOLFSSL_MAX_ESCBUF` guard and returns `WS_FATAL_ERROR`. This caller then clears the escape state, falls through with `i` still 0, writes those bytes as plain output, and returns `WS_SUCCESS`. That means the new defensive rejection path introduced by the PR is not actually enforced for the resume case it is meant to harden.

**Suggestion:**
```suggestion
ret = wolfSSH_DoControlSeq(ssh, handle, buf, bufSz, &i);
if (ret == WS_WANT_READ) {
    return ret;
}
if (ret != WS_SUCCESS) {
    return ret;
}
ssh->escState = WC_ESC_NONE;
ssh->escBufSz = 0;
break;
```

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **CSI capacity errors are swallowed on resume** — `src/wolfterm.c:684-691`

*Review generated by Skoll.*

src/wolfterm.c
yosuke-wolfssl: Hello @aidangarske ,
Thank you for review. Digging in, this turns out to be pre-existing behavior rather than something this PR introduces, so I've kept the fix consistent with the existing convention and filed the deeper
cleanup as a follow-up.
Let me share my analysis with you:

**Why the guards stay `WS_FATAL_ERROR`:** `wolfSSH_DoControlSeq` already returns
`WS_FATAL_ERROR` for the identical `escBufSz >= WOLFSSL_MAX_ESCBUF` capacity
check at the top of the function, and `wolfSSH_ConvertConsole` already only
branches on `WS_WANT_READ` — so that existing fatal is swallowed on resume too.
My new guards mirror that existing check. Importantly, both guards return
*before* the `WMEMCPY` into `escBuf`, so the overflow they protect against
cannot happen regardless of how the caller treats the return code — the
memory-safety fix is enforced. (This path is also only reachable with a custom
`WOLFSSH_MAX_CONSOLE_ARGS >= WOLFSSL_MAX_ESCBUF` build; in the default config
`getArgs` caps the advance at 16 < 19.)

**Why not the suggested `if (ret != WS_SUCCESS) return ret;`:** it changes
existing default-build behavior. `wolfSSH_DoControlSeq` returns `WS_FATAL_ERROR`
from its `default:` case for any CSI command not in its switch — including
cursor up/down (`A`/`B`) and save/restore (`s`/`u`) — which are currently
ignored, and the api-test already asserts unknown console codes yield
`WS_SUCCESS` ("should skip over unknown console code"). Propagating every
non-`WS_SUCCESS` would break that test and would abort the session on those
sequences when they land split across a packet boundary.

**The real fix** is architectural: `WS_FATAL_ERROR` is overloaded for both
"unsupported sequence, skip it" and "corrupt state, stop." The clean solution is
to make the unsupported/`default:` path return `WS_SUCCESS` (skip) and reserve
`WS_FATAL_ERROR` for genuine bad state, after which `ConvertConsole` can
propagate fatals safely. That touches the error contract across three entry
paths and needs its own tests, so I'd prefer to do it as a separate PR rather
than widen this security fix.

Current PR remains focused on the OOB read + escape-state leak, verified on
Windows (MSVC, Debug x64) with api-test passing.
Please let me know if you prefer to widen this PR to fix the whole error handling.
I'll rework here.

aidangarske (dismissed):
over to @ejohnstown

ejohnstown (approved):

PR #1079, opened 2026-06-30
wolfsshd: bind FPKI certificate UPN realm to AuthorizedUPNDomains

yosuke-wolfssl:
### Summary

When `WOLFSSL_FPKI` is enabled, `wolfsshd` validated a client certificate's UPN
SAN by comparing only the local part (the text before `@`) to the requested SSH
username and discarding the realm/domain. A client holding a valid certificate
from a *trusted* CA for `alice@other-domain.example` could therefore
authenticate as local user `alice`. This is most dangerous on the CA-only path
(no `AuthorizedKeysFile` set), where the CA-verified chain plus the local-part
match are the only gates.

This PR adds an **opt-in** allowlist of permitted UPN realms, `AuthorizedUPNDomains`.
When set, a certificate's UPN realm must exactly match one of the configured
domains (case-insensitive) in addition to the existing local-part match. When
unset, behavior is unchanged for backward compatibility and a warning is logged
so operators know the domain is not being checked.

Addressed by f_5580.

### Problem

`RequestAuthentication` (`apps/wolfsshd/auth.c`) scanned the UPN SAN up to `@`
and matched only the local part:

```c
for (idx = 0; idx < current->len; idx++) {
    if (current->name[idx] == '@') break;
}
if ((int)XSTRLEN(usr) == idx && XSTRNCMP(usr, current->name, idx) == 0) {
    usrMatch = 1;
}
```

The realm after `@` was explicitly ignored, so any trusted-CA certificate whose
local part equalled the target username was accepted regardless of its domain.

### Solution

- New config option **`AuthorizedUPNDomains`** — a whitespace/comma separated
  list of permitted UPN realms. Plumbed like the other `WOLFSSHD_CONFIG` string
  options (struct field, option table, copy/free, getter) and scoped to `Match`
  blocks via the existing config copy.
- New helper **`MatchUPNToUser()`** that performs the local-part match and, when
  an allowlist is configured, requires the certificate's realm to be present and
  to match a listed domain exactly (case-insensitive). It is length-bounded
  (`nameSz`) because certificate `altName` buffers are not guaranteed to be
  NUL-terminated.
- The multi-value `ForceCommand`/`AuthorizedUPNDomains` storage was de-duplicated
  into a shared `SetListString()` helper instead of a per-option handler.

Example configuration:

```
# /etc/ssh/sshd_config
TrustedUserCAKeys /etc/ssh/ca-cert-ecc.pem
HostCertificate   /etc/ssh/server-cert.pem

# only accept certificate UPNs in these realms
AuthorizedUPNDomains corp.example partner.example
```

Behavior:

| `AuthorizedUPNDomains` | Certificate UPN     | Result                         |
| ---------------------- | ------------------- | ------------------------------ |
| unset                  | `alice@anything`    | accepted (local part only) + warning logged |
| `corp.example`         | `alice@corp.example`| accepted                       |
| `corp.example`         | `alice@other.example`| rejected                      |
| `corp.example`         | `alice` (no realm)  | rejected                       |

### Backward compatibility

Unset is the default and preserves the prior local-part-only behavior, so
existing FPKI deployments are unaffected until they opt in. A `WS_LOG_WARN` is
emitted on each authentication attempt in which a certificate UPN matches while
the option is unset, making the unchecked-domain gap visible in logs.

The warning is deliberately per-attempt rather than one-time-per-process: a
one-time flag would require shared mutable state, which is racy in threaded
builds (Windows), and the log volume is bounded by the rate of successful
certificate logins on a deployment that has chosen not to opt in.

### Security guidance

`AuthorizedUPNDomains` is opt-in, so the default remains fail-open with respect
to the certificate realm. FPKI deployments that rely on the CA-only path (no
`AuthorizedKeysFile`) **should always set `AuthorizedUPNDomains`**, since the
CA-verified chain and the local-part match are otherwise the only gates on
identity. The per-attempt warning exists to make an un-opted-in deployment
visible in the logs.

### Files changed

- `apps/wolfsshd/configuration.c`, `apps/wolfsshd/configuration.h` — new option,
  getter, and shared `SetListString()` helper.
- `apps/wolfsshd/auth.c`, `apps/wolfsshd/auth.h` — `MatchUPNToUser()` and its
  use in `RequestAuthentication`; helper exported under `WOLFSSHD_UNIT_TEST`
  via `WOLFSSHD_STATIC`.
- `apps/wolfsshd/wolfsshd.c` — `wolfsshd -?` advertises a `Build features: FPKI
  certificate UPN domain checking` line so test scripts can detect the feature
  deterministically.
- `apps/wolfsshd/test/test_configuration.c` — unit tests for option parsing,
  `Match`-block copy, and `MatchUPNToUser` (match/mismatch, missing/empty realm,
  multi/comma/tab/CR-LF separators, case-insensitivity, and length-bounded
  buffers).
- `apps/wolfsshd/test/create_sshd_config.sh`,
  `apps/wolfsshd/test/sshd_x509_upn_fail.sh`,
  `apps/wolfsshd/test/run_all_sshd_tests.sh` — end-to-end negative test: a cert
  with realm `example` is rejected by a daemon configured with
  `AuthorizedUPNDomains other.example`.

### Testing

- **Unit tests** (`apps/wolfsshd/test/test_configuration`): all pass. New
  `test_MatchUPNToUser` covers the separator set, case-insensitivity, and the
  length-bounded (non-NUL-terminated) buffer path. Negative controls confirm the
  vectors fail when enforcement is removed or when the helper reads past
  `nameSz`.
- **Integration test**: `sshd_x509_upn_fail.sh` gates on FPKI by grepping the
  daemon's help output, then asserts the daemon-side rejection reason
  (`incorrect user cert sent`) from the log with a before/after match count, so
  neither a non-FPKI build, a stale log line, nor an unrelated client failure
  can produce a false pass. Skips cleanly (exit 77) when FPKI is not compiled in.
- The existing positive test `sshd_config_test_x509` now also sets
  `AuthorizedUPNDomains example`, matching the fixture certificate's UPN realm
  (`keys/renewcerts.cnf`: `otherName = msUPN;UTF8:fred@example`). This gives the
  allowlist-accept path end-to-end coverage, which it otherwise lacked.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -1387,6 +1387,70 @@ static int CAKeysFileDiffers(const char* a, const char* b)
 }
 
 
+/* Returns 1 when the certificate UPN <user>@<domain> in name[0..nameSz)
+ * authorizes login as 'usr'. allowList is a whitespace/comma list of permitted
+ * realms; NULL/empty matches the local part only, else domain must be listed. */
+#if defined(WOLFSSL_FPKI) || defined(WOLFSSHD_UNIT_TEST)
+WOLFSSHD_STATIC int MatchUPNToUser(const char* usr, const char* name,
+                                   int nameSz, const char* allowList)
+{
+    int ret = 0;
+    int idx = 0;
+    int domainSz;
+    int tokSz;
+    const char* domain;
+    const char* p;
+    const char* tok;
+
+    if (usr != NULL && name != NULL && nameSz >= 0) {
+        /* locate the '@' separating the local part from the domain */
+        for (idx = 0; idx < nameSz; idx++) {
+            if (name[idx] == '@') {
+                break;
+            }
+        }
+
+        /* the local part must equal the requested user name exactly */
+        if ((int)XSTRLEN(usr) == idx && XSTRNCMP(usr, name, idx) == 0) {
+            if (allowList == NULL || *allowList == '\0') {
+                /* no allowlist configured: keep local-part-only matching */
+                ret = 1;
+            }
+            else if (idx < nameSz) {
+                /* an allowlist is set: the UPN domain must be present and
+                 * listed as an exact, case-insensitive match */
+                domain = name + idx + 1;
+                domainSz = nameSz - idx - 1;
+
+                p = allowList;
+                while (*p != '\0' && ret == 0) {
+                    /* skip separators preceding the realm token */
+                    while (*p == ' ' || *p == '\t' || *p == '\r' ||
+                            *p == '\n' || *p == ',') {
+                        p++;
+                    }
+
+                    tok = p;
+                    while (*p != '\0' && *p != ' ' && *p != '\t' &&
+                            *p != '\r' && *p != '\n' && *p != ',') {
+                        p++;
+                    }
+                    tokSz = (int)(p - tok);
+
+                    if (tokSz > 0 && tokSz == domainSz &&
+                            WSTRNCASECMP(tok, domain, (size_t)domainSz) == 0) {
+                        ret = 1;
+                    }
+                }
+            }
+        }
+    }
+
+    return ret;
+}
+#endif /* WOLFSSL_FPKI || WOLFSSHD_UNIT_TEST */
+
+
 /*
  * @TODO this will take a pipe or equivalent to talk to a privileged thread
  * rather than having WOLFSSHD_AUTH directly with privilege separation.
@@ -1538,29 +1602,33 @@ static int RequestAuthentication(WS_UserAuthData* authData,
                 }
                 else {
                     int usrMatch = 0;
+                    int upnRealmUnchecked = 0;
                     DNS_entry* current = dCert->altNames;
+                    const char* upnDomains =
+                        wolfSSHD_ConfigGetAuthorizedUPNDomains(usrConf);
 
                     while (current != NULL) {
                         if (current->type == ASN_OTHER_TYPE &&
                                 current->oidSum == UPN_OID) {
-                            /* found UPN oid, check name against user */
-                            int idx;
-
-                            for (idx = 0; idx < current->len; idx++) {
-                                if (current->name[idx] == '@') break;
-                                /* UPN format is <user>@<domain>
-                                 * since currently not doing any checks on
-                                 * domain it is  not treated as an error if only
-                                 * the user name is present without the domain
-                                 */
-                            }
-
-                            if ((int)XSTRLEN(usr) == idx &&
-                                    XSTRNCMP(usr, current->name, idx) == 0) {
+                            /* bind the cert identity to the requested user;
+                             * MatchUPNToUser also enforces the realm allowlist
+                             * when AuthorizedUPNDomains is set */
+                            if (MatchUPNToUser(usr, current->name, current->len,
+                                    upnDomains)) {
                                 usrMatch = 1;
+                                if (upnDomains == NULL || *upnDomains == '\0') {
+                                    upnRealmUnchecked = 1;
+                                }
                             }
                         }
                         current = current->next;
+                    }
+
+                    /* a UPN matched but no realm policy is set; warn per auth
+                     * attempt so the opt-in gap is visible, no shared state */
+                    if (upnRealmUnchecked) {
+                        wolfSSH_Log(WS_LOG_WARN, "[SSHD] AuthorizedUPNDomains "
+                            "not set; certificate UPN domain is not checked");
                     }
 
                     if (usrMatch == 0) {
--- a/apps/wolfsshd/test/create_sshd_config.sh
+++ b/apps/wolfsshd/test/create_sshd_config.sh
@@ -26,9 +26,27 @@ PermitEmptyPasswords no
 UsePrivilegeSeparation no
 UseDNS no
 
-TrustedUserCAKeys $PWD/../../../keys/ca-cert-ecc.pem     
-HostKey $PWD/../../../keys/server-key.pem                
+TrustedUserCAKeys $PWD/../../../keys/ca-cert-ecc.pem
+HostKey $PWD/../../../keys/server-key.pem
 HostCertificate $PWD/../../../keys/server-cert.pem
+AuthorizedUPNDomains example
+
+EOF
+
+cat <<EOF > sshd_config_test_x509_upn_bad
+Port 22222
+Protocol 2
+LoginGraceTime 600
+PermitRootLogin yes
+PasswordAuthentication yes
+PermitEmptyPasswords no
+UsePrivilegeSeparation no
+UseDNS no
+
+TrustedUserCAKeys $PWD/../../../keys/ca-cert-ecc.pem
+HostKey $PWD/../../../keys/server-key.pem
+HostCertificate $PWD/../../../keys/server-cert.pem
+AuthorizedUPNDomains other.example
 
 EOF
 
--- a/apps/wolfsshd/test/sshd_x509_upn_fail.sh
+++ b/apps/wolfsshd/test/sshd_x509_upn_fail.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+
+# Negative test for AuthorizedUPNDomains. run_all_sshd_tests.sh starts the
+# daemon with sshd_config_test_x509_upn_bad, whose AuthorizedUPNDomains is
+# "other.example", while the client certificate carries the UPN realm
+# "example". The wolfSSHd UPN domain check must therefore reject the cert.
+
+PWD=`pwd`
+
+# The UPN domain check is compiled only when wolfSSL is built with FPKI. Probe
+# the daemon binary's help output, which prints an FPKI marker under the same
+# build guard, and skip when the check is not present.
+if ! ../wolfsshd "-?" 2>&1 | grep -q "FPKI"; then
+    echo "wolfSSHd built without FPKI; UPN domain check not compiled in, skipping"
+    exit 77
+fi
+
+# Count existing rejection lines first so a stale match left in the appended
+# log (start_sshd.sh uses 'wolfsshd -E ./log.txt', which never truncates) is
+# not mistaken for this run's rejection.
+BEFORE=`sudo grep -c "incorrect user cert sent" ./log.txt 2>/dev/null`
+BEFORE=${BEFORE:-0}
+
+cd ../../..
+
+if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
+    echo "expecting host, port and user as arguments"
+    echo "$0 127.0.0.1 22222 user"
+    exit 1
+fi
+
+TEST_CLIENT="./examples/client/client"
+PRIVATE_KEY="./keys/$3-key.der"
+PUBLIC_KEY="./keys/$3-cert.der"
+CA_CERT="./keys/ca-cert-ecc.der"
+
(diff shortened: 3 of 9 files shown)


apps/wolfsshd/test/test_configuration.c
```diff
@@ -1618,6 +1625,119 @@ static int test_CAKeysFileDiffers(void)
...
+static int test_MatchUPNToUser(void)
+{
+    int ret = WS_SUCCESS;
+    int i;
+    int sz;
+
+    /* name buffers are not necessarily NUL terminated in a real cert, so the
+     * helper is length bounded. nameSz of 0 derives the length from the string;
+     * a non-zero nameSz bounds a longer buffer to prove bytes past it are
+     * ignored. */
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/auth.c
```diff
@@ -1539,24 +1603,28 @@ static int RequestAuthentication(WS_UserAuthData* authData,
                 else {
                     int usrMatch = 0;
                     DNS_entry* current = dCert->altNames;
+                    const char* upnDomains =
+                        wolfSSHD_ConfigGetAuthorizedUPNDomains(usrConf);
+                    static int upnDomainWarned = 0;
+
+                    /* warn when the UPN domain is unchecked so the opt-in gap
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/auth.c
```diff
@@ -1538,31 +1602,35 @@ static int RequestAuthentication(WS_UserAuthData* authData,
...
-
-                            if ((int)XSTRLEN(usr) == idx &&
-                                    XSTRNCMP(usr, current->name, idx) == 0) {
+                            /* bind the cert identity to the requested user;
+                             * MatchUPNToUser also enforces the realm allowlist
+                             * when AuthorizedUPNDomains is set */
+                            if (MatchUPNToUser(usr, current->name, current->len,
+                                    upnDomains)) {
                                 usrMatch = 1;
+                                if (upnDomains == NULL || *upnDomains == '\0') {
```
aidangarske, line 1619: **🔵 [Low] AuthorizedUPNDomains opt-in default keeps realm-unchecked (fail-open) UPN matching and warns on every cert auth attempt**
*💡 SUGGEST*

When AuthorizedUPNDomains is unset (the default), MatchUPNToUser returns 1 on a local-part-only match (allowList == NULL || *allowList == '\0' branch), so a trusted-CA certificate for alice@any-realm still authenticates as local user alice — the original cross-realm behavior. The PR fixes the cross-domain bypass only when an operator opts in. This is a deliberate, documented backward-compatibility decision (security view: Info, CWE-1188, fail-open but mitigated). It is mitigated by a per-attempt WS_LOG_WARN; the code comment correctly documents this as per-attempt with no shared state, but the PR description claims a one-time (per process) warning — the description is stale relative to the code (review view: Low). On a busy FPKI deployment that has intentionally not opted in, this produces one WARN line per successful cert login. Confirmed against the code at auth.c:1616-1632.

**Recommendation:** Behavior is acceptable for backward compatibility. Document prominently that FPKI CA-only deployments should always set AuthorizedUPNDomains, or provide a build/config option to make realm checking fail-closed for security-sensitive deployments. Separately, reconcile the code and PR description: either confirm per-attempt logging is intended (visibility vs. log volume) or gate the warning with a static flag to make it one-time-per-process as the PR text implies.

apps/wolfsshd/test/create_sshd_config.sh
```diff
@@ -26,9 +26,27 @@ PermitEmptyPasswords no
 UsePrivilegeSeparation no
 UseDNS no
 
-TrustedUserCAKeys $PWD/../../../keys/ca-cert-ecc.pem     
-HostKey $PWD/../../../keys/server-key.pem                
+TrustedUserCAKeys $PWD/../../../keys/ca-cert-ecc.pem
+HostKey $PWD/../../../keys/server-key.pem
+HostCertificate $PWD/../../../keys/server-cert.pem
+AuthorizedUPNDomains example
```
aidangarske, line 32: **🟡 [Medium] Existing positive x509 UPN test now depends on fixture cert realm being exactly 'example'**
*💡 SUGGEST*

The pre-existing positive config sshd_config_test_x509 now sets AuthorizedUPNDomains example. This converts a previously domain-agnostic passing test into one that requires the client fixture certificate's UPN realm to be exactly example (case-insensitive). If the fixture cert generated by renewcerts.sh carries any other realm (or none), the existing positive x509 test will start failing as a side effect of this PR. The new negative test (sshd_config_test_x509_upn_bad with other.example) is internally consistent with the assumption that the cert realm is example, but that assumption is now load-bearing for the positive path too.

**Recommendation:** Confirm the generated client cert's UPN realm is example (grep the cert / renewcerts output). If it is not guaranteed, either align the config value with the actual fixture realm or drop AuthorizedUPNDomains from the positive config to keep that test testing only the local-part path.

apps/wolfsshd/test/sshd_x509_upn_fail.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+# Negative test for AuthorizedUPNDomains. run_all_sshd_tests.sh starts the
+# daemon with sshd_config_test_x509_upn_bad, whose AuthorizedUPNDomains is
+# "other.example", while the client certificate carries the UPN realm
+# "example". The wolfSSHd UPN domain check must therefore reject the cert.
```
aidangarske, line 6: **🔵 [Low] Unquoted variable expansions in new test script reduce robustness**
*🔧 NIT*

PWD=`pwd` and the later cd $PWD are unquoted; a working directory containing spaces would break the cd. This matches the existing sibling scripts' style so it is not new-in-kind, but the new file could quote for safety.

**Recommendation:** Quote the expansions (cd "$PWD"). Non-blocking; consistent with existing scripts otherwise.

apps/wolfsshd/auth.c
```diff
@@ -1387,6 +1387,70 @@ static int CAKeysFileDiffers(const char* a, const char* b)
 }
 
 
+/* Returns 1 when the certificate UPN <user>@<domain> in name[0..nameSz)
```
aidangarske, line 1390: **⚪ [Info] Multi-line function header comment on MatchUPNToUser**
*🔧 NIT*

The three-line block comment above the helper describes parameters/return in the style the project's comment guidance discourages for new functions. The function name and the inline comments inside already make the intent clear. Minor; several existing functions in this file use similar blocks, so this is consistent with the file if not with the stricter guidance.

**Recommendation:** Optionally collapse to a single-line WHY comment. Take it or leave it.

apps/wolfsshd/auth.c
```diff
@@ -1387,6 +1387,70 @@ static int CAKeysFileDiffers(const char* a, const char* b)
...
+
+                    tok = p;
+                    while (*p != '\0' && *p != ' ' && *p != '\t' &&
+                            *p != '\r' && *p != '\n' && *p != ',') {
+                        p++;
+                    }
+                    tokSz = (int)(p - tok);
+
+                    if (tokSz > 0 && tokSz == domainSz &&
+                            WSTRNCASECMP(tok, domain, (size_t)domainSz) == 0) {
```
aidangarske, line 1441: **⚪ [Info] WSTRNCASECMP degrades to case-sensitive strncmp on MICROCHIP targets**
*🔧 NIT*

The realm comparison relies on WSTRNCASECMP for case-insensitive matching. On MICROCHIP_MPLAB_HARMONY/PIC32 builds (wolfssh/port.h:599) this macro maps to plain strncmp, so realm matching would be case-sensitive there. wolfsshd realistically only builds on Unix/Windows where the case-insensitive mapping applies, so this is a theoretical note rather than a real defect for this app.

**Recommendation:** No action needed unless wolfsshd is ever targeted at MICROCHIP builds; noted for completeness.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **AuthorizedUPNDomains opt-in default keeps realm-unchecked (fail-open) UPN matching and warns on every cert auth attempt** — `apps/wolfsshd/auth.c:1619-1632`
- [Medium] **Existing positive x509 UPN test now depends on fixture cert realm being exactly 'example'** — `apps/wolfsshd/test/create_sshd_config.sh:32`
- [Low] **Unquoted variable expansions in new test script reduce robustness** — `apps/wolfsshd/test/sshd_x509_upn_fail.sh:6`
- [Info] **Multi-line function header comment on MatchUPNToUser** — `apps/wolfsshd/auth.c:1390-1392`
- [Info] **WSTRNCASECMP degrades to case-sensitive strncmp on MICROCHIP targets** — `apps/wolfsshd/auth.c:1441`

*Review generated by Skoll.*

apps/wolfsshd/auth.c
yosuke-wolfssl: Thanks — you're right that the description was stale. The per-attempt warning is intentional: the original one-time-per-process version used a function-local static int, which Copilot correctly flagged as a data race in threaded (Windows) builds, so I moved to a stack-local upnRealmUnchecked set only when a UPN actually matches. I've updated the PR description to describe the per-attempt behavior and explain why, and added a "Security guidance" section stating that FPKI CA-only deployments should always set AuthorizedUPNDomains. Keeping the default opt-in for backward compatibility as you suggested; log volume is bounded by the rate of successful cert logins on a deployment that has chosen not to opt in.

apps/wolfsshd/test/create_sshd_config.sh
yosuke-wolfssl: Confirmed, the realm is exactly example and it's pinned rather than incidental. keys/renewcerts.cnf:47 has otherName = msUPN;UTF8:fred@example, and the checked-in keys/fred-cert.der decodes to othername: UPN:fred@example.

Worth noting the positive x509 test was already load-bearing on this fixture before the change: MatchUPNToUser requires the local part to equal the requested username, so the cert already had to be fred@... for that test to pass. Adding AuthorizedUPNDomains example extends an existing dependency rather than introducing a new one, and it gives the allowlist-accept path end-to-end coverage that nothing else exercised. Leaving as-is.

apps/wolfsshd/test/sshd_x509_upn_fail.sh
yosuke-wolfssl: Fixed — quoted the cd "$PWD" on line 41. The PWD=\pwd`` assignment on line 8 doesn't word-split (assignments aren't subject to it), so I left it matching the sibling scripts' style.

apps/wolfsshd/auth.c
yosuke-wolfssl: Agreed, no action. wolfsshd isn't targeted at MICROCHIP_MPLAB_HARMONY/PIC32 builds, so the case-sensitive strncmp mapping in wolfssh/port.h isn't reachable here.

apps/wolfsshd/auth.c
yosuke-wolfssl: Leaving as-is. The block is exactly three lines, but it's consistent with the other function headers in this file as you mentioned.

yosuke-wolfssl:
Hello @aidangarske ,
Could you review this again please ?

aidangarske (approved):
Looks good i would like @ejohnstown to look over this though.

ejohnstown (approved):

PR #1081, opened 2026-07-02
Add TPM-backed X.509 host certificate support and fix TPM-build client auth

aidangarske:
# Description 

- Add tpm backed x509 host cert support for tpm builds 
- minor change and fixes a TPM-build regression where clients could not perform password/keyboard-interactive authentication.
- CI test for matrix x509 added 

support for ZD 22094

### DIFF
--- a/.github/workflows/tpm-ssh.yml
+++ b/.github/workflows/tpm-ssh.yml
@@ -17,6 +17,10 @@ jobs:
       matrix:
         keytype: [ rsa, ecc ]
         sim: [ ibmswtpm2, fwtpm ]
+        # raw: plain TPM host key (verified with OpenSSH).
+        # x509: TPM host key presented as an X.509 certificate, verified by the
+        #       tpmcertserver/tpmcertclient example.
+        hostkey: [ raw, x509 ]
 
     steps:
     - uses: actions/checkout@v6
@@ -47,8 +51,12 @@ jobs:
       run: |
         cd wolfssl
         ./autogen.sh
+        # certgen/certreq/certext: generate the X.509 host certificate from the
+        # TPM key. WC_SIG_MIN_HASH_TYPE=SHA: the RSA x509v3-ssh-rsa host cert is
+        # SHA-1; modern wolfSSL otherwise rejects SHA-1 RSA signatures.
         ./configure --enable-wolftpm --enable-wolfssh --enable-keygen \
-            CFLAGS="-DWC_RSA_NO_PADDING"
+            --enable-certgen --enable-certreq --enable-certext \
+            CFLAGS="-DWC_RSA_NO_PADDING -DWC_SIG_MIN_HASH_TYPE=WC_HASH_TYPE_SHA"
         make
         sudo make install
         sudo ldconfig
@@ -97,13 +105,14 @@ jobs:
       run: |
         cd wolfssh
         ./autogen.sh
-        ./configure --enable-tpm
+        ./configure --enable-tpm --enable-certs
         make
         sudo make install
         sudo ldconfig
 
     # Server host key resident in the TPM: the private key never enters RAM.
     - name: Test TPM host key (${{ matrix.keytype }})
+      if: matrix.hostkey == 'raw'
       run: |
         cd wolftpm
         ./examples/keygen/keygen hostkey.bin -${{ matrix.keytype }} -t -eh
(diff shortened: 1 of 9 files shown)


src/internal.c
```diff
@@ -16422,6 +16417,19 @@ int SendUserAuthRequest(WOLFSSH* ssh, byte authType, int addSig)
         keySig_ptr->sigId = ID_NONE;
         keySig_ptr->heap = ssh->ctx->heap;
 
+#ifdef WOLFSSH_TPM
```
dgarske, line 16420: **🟠 [Medium] TPM publickey preference has no fallback to password/keyboard on failure**

When `ssh->ctx->tpmKey != NULL` and the server offers publickey, the new block strips the `WOLFSSH_USERAUTH_PASSWORD` (and `WOLFSSH_USERAUTH_KEYBOARD`) bits from `authType`. Because `DoUserAuthFailure()` re-invokes `SendUserAuthRequest()` on every server FAILURE, a client with a TPM key configured can never fall back to password/keyboard auth if the TPM publickey attempt is rejected by the server -- it will keep re-sending publickey until the server disconnects. The comment says 'prefer publickey', but the implementation makes it exclusive with no fallback. This matches the pre-PR TPM behavior (which was publickey-only), so it is not a regression, but confirm exclusivity (rather than true 'prefer with fallback') is the intended design. If a client legitimately has both a TPM key and a valid password and the key is not authorized server-side, it will fail to authenticate.

**Fix:** Confirm the intended behavior. If password/keyboard fallback should occur after publickey fails, gate the stripping so it only applies to the first attempt (e.g. track a publickey-attempted flag) rather than on every DoUserAuthFailure retry.

examples/tpmcertserver/tpmcertclient.c
```diff
@@ -0,0 +1,245 @@
...
+static int TpmCcHostKeyCheck(const byte* pubKey, word32 pubKeySz, void* ctx)
+{
+    WOLFSSH_UNUSED(pubKey);
+    WOLFSSH_UNUSED(pubKeySz);
+    WOLFSSH_UNUSED(ctx);
+    return 0;
+}
+
+
+static int TpmCcLoadFile(const char* file, byte* buf, word32* bufSz)
```
dgarske, line 70: **🟠 [Medium] TpmCcLoadFile silently truncates CA files larger than the buffer**

`fread(buf, 1, *bufSz, f)` reads at most `*bufSz` (2048) bytes. If the CA file is larger than the buffer, `n == *bufSz` and the function reports success (`*bufSz = n`) while silently returning a truncated certificate. The truncated DER then fails to parse in `wolfSSH_CTX_AddRootCert_buffer()` with a confusing error, or (worse for a test harness) masks the real cause. There is no check that EOF was actually reached.

**Fix:** Detect the file-larger-than-buffer case and return an error instead of loading a truncated certificate.

examples/tpmcertserver/tpmcertserver.c
```diff
@@ -0,0 +1,398 @@
...
+
+/* Create a signing key inside the TPM and produce a self-signed X.509
+ * certificate over its public key. The private key never leaves the TPM: the
+ * certificate is signed through the wolfTPM crypto callback. */
+static int TpmCsMakeKeyAndCert(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
+        int useEcc, byte* certDer, word32* certDerSz)
+{
+    int rc;
+    int devId = INVALID_DEVID;
+    int sigType;
```
dgarske: **🔵 [Low] sigType/subject may trigger -Wmaybe-uninitialized**

`int sigType;` and `const char* subject;` are only assigned inside the first `if (rc == 0)` block and only read inside the later `if (rc == 0)` block that calls `wolfTPM2_CSR_Generate_ex()`. The uses are logically guarded (they are reachable only when the assigning block ran), but GCC/Clang flow analysis often cannot correlate the separate `rc == 0` blocks across the intervening `wolfTPM2_CreateAndLoadKey()` call and may emit `-Wmaybe-uninitialized`. If the example is ever built with `-Werror` this breaks the build.

**Fix:** Initialize `sigType = 0;` and `subject = NULL;` at declaration to silence potential maybe-uninitialized warnings.

examples/tpmcertserver/tpmcertserver.c
```diff
@@ -0,0 +1,398 @@
...
+{
+    int ret = WOLFSSH_USERAUTH_INVALID_PASSWORD;
+
+    WOLFSSH_UNUSED(ctx);
+
+    if (authType == WOLFSSH_USERAUTH_PASSWORD) {
+        const char* pw = (const char*)authData->sf.password.password;
+        word32 pwSz = authData->sf.password.passwordSz;
+
+        if (pwSz == (word32)XSTRLEN(TPMCS_PASSWORD)
```
dgarske, line 69: **🔵 [Low] Example server compares password with non-constant-time XMEMCMP**

The password check uses `XMEMCMP`, which is not constant-time and violates the wolfSSL convention of comparing secrets with `ConstantCompare()`. This is example/demo code with a hardcoded password, so impact is negligible, but the convention exists and other examples may be copied from this one. Note this is a convention nit, not a security-audit finding.

**Fix:** Use `ConstantCompare()` for the password comparison to match repo convention, or leave as-is given it is demo code with a fixed credential.

.github/workflows/tpm-ssh.yml
```diff
@@ -129,9 +143,55 @@ jobs:
...
+        done
+        ./examples/tpmcertserver/tpmcertclient -A tpm-server-cert.der \
+            -p 22223 -h 127.0.0.1 > tpmcert_client.txt 2>&1
+        echo "----- server -----"; cat tpmcert_server.txt
+        echo "----- client -----"; cat tpmcert_client.txt
+        grep -q "verified server X.509 host certificate" tpmcert_client.txt
+        grep -q "Client connected and verified the TPM-backed host certificate" \
+            tpmcert_server.txt
+
+    # Negative test: a client that trusts an unrelated CA must reject the server.
```
dgarske, line 169: **🔵 [Low] Negative CI test can false-pass on server startup failure**

The negative test asserts rejection purely via the client's non-zero exit code (`if tpmcertclient ...; then ERROR`). If the server never comes up (TPM provisioning failure, port bind failure, etc.), the client also exits non-zero -- from a plain connection failure, not certificate rejection -- and the negative test passes anyway, hiding the real problem. This is largely mitigated because the positive test step (lines 149-167) runs first in the same matrix cell and would fail on a broken server, but the negative step is not self-guarding.

**Fix:** After the client fails, additionally assert the failure was a certificate/verification error (e.g. `grep -q "wolfSSH_connect failed" tpmcert_client_neg.txt` and confirm the server logged that it started listening) so a startup failure cannot masquerade as a correct rejection.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **TPM publickey preference has no fallback to password/keyboard on failure** — `src/internal.c:16420-16431`
- [Medium] **TpmCcLoadFile silently truncates CA files larger than the buffer** — `examples/tpmcertserver/tpmcertclient.c:70-89`
- [Low] **sigType/subject may trigger -Wmaybe-uninitialized** — `examples/tpmcertserver/tpmcertserver.c:87-91`
- [Low] **Example server compares password with non-constant-time XMEMCMP** — `examples/tpmcertserver/tpmcertserver.c:69-72`
- [Low] **Negative CI test can false-pass on server startup failure** — `.github/workflows/tpm-ssh.yml:169-190`

dgarske (changes requested):
https://github.com/wolfSSL/wolfssh/pull/1081#pullrequestreview-4621584250

dgarske (approved):

PR #1083, opened 2026-07-06
Bound SFTP handle length against received payload in GetHandle

yosuke-wolfssl:
## Summary

Fixes a heap over-read in the SFTP **client** when parsing a server's `SSH_FXP_HANDLE` reply.

`wolfSSH_SFTP_GetHandle` allocates `state->buffer` to exactly the server-declared payload size, then reads the handle-string length prefix `sz` and validates it only against `WOLFSSH_MAX_HANDLE` and the caller's `*handleSz` — never against the bytes actually present in the buffer. A malicious or compromised server can declare a 4-byte payload (just the length prefix) while encoding `sz = 256`. Both existing guards pass (`256 > 256` is false; the caller sets `*handleSz = WOLFSSH_MAX_HANDLE = 256`, so `256 < 256` is false), and:

```c
WMEMCPY(handle,
        (wolfSSH_SFTP_buffer_data(&state->buffer) + UINT32_SZ),
        *handleSz);
```

copies 256 bytes out of a 4-byte allocation — reading ~252 bytes past the buffer. Those bytes land in the client's `handle`, which is echoed verbatim to the server on every later `READ`/`WRITE`/`CLOSE`/`READDIR`, disclosing client heap memory; a read crossing an unmapped page crashes the client instead.

Addressed by f_6696.

## Fix

Add a payload bound to the guard in `wolfSSH_SFTP_GetHandle` — reject when the declared handle length exceeds what the buffer actually holds:

```c
if (wolfSSH_SFTP_buffer_ato32(&state->buffer, &sz) != WS_SUCCESS
       || sz > WOLFSSH_MAX_HANDLE || *handleSz < sz
       || UINT32_SZ + sz
              > wolfSSH_SFTP_buffer_size(&state->buffer)) {
```

`sz` is capped at `WOLFSSH_MAX_HANDLE` (256) by the earlier check, so `UINT32_SZ + sz` cannot overflow. This matches the bound `GetStringRef` already enforces on length-prefixed strings elsewhere.

## Tests

- Added a `WOLFSSH_TEST_INTERNAL` hook `wolfSSH_TestSftpGetHandle` to drive the static function.
- Added `test_SftpGetHandle_sizeBound`, which injects a crafted `HANDLE` reply into a channel and drives `wolfSSH_SFTP_GetHandle`:
  - **Reject case** — payload=4, declared length=256 → must return non-success with `ssh->error == WS_BUFFER_E` (no over-read).
  - **Accept case** — payload=8 carrying a 4-byte handle (exact fit) → must return `WS_SUCCESS` with a clean `ssh->error`.

## Verification

- Unit suite passes: `SftpGetHandle_sizeBound: SUCCESS`, exit 0.
- **Negative control**: reverting only the guard makes ASan report `heap-buffer-overflow READ of size 256` at the `WMEMCPY` in `wolfSSH_SFTP_GetHandle`, confirming the test exercises the exact over-read.

## Files changed

- `src/wolfsftp.c` — payload bound in the guard; `wolfSSH_TestSftpGetHandle` hook.
- `wolfssh/wolfsftp.h` — prototype for the new test hook.
- `tests/unit.c` — injection helper, bound test, and registration.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -7275,7 +7275,9 @@ static int wolfSSH_SFTP_GetHandle(WOLFSSH* ssh, byte* handle, word32* handleSz)
                  * max size */
                 wolfSSH_SFTP_buffer_rewind(&state->buffer);
                 if (wolfSSH_SFTP_buffer_ato32(&state->buffer, &sz) != WS_SUCCESS
-                       || sz > WOLFSSH_MAX_HANDLE || *handleSz < sz) {
+                       || sz > WOLFSSH_MAX_HANDLE || *handleSz < sz
+                       || UINT32_SZ + sz
+                              > wolfSSH_SFTP_buffer_size(&state->buffer)) {
                     WLOG(WS_LOG_SFTP, "Handle size found was too big");
                     WLOG(WS_LOG_SFTP, "Check size set in input handleSz");
                     ssh->error = WS_BUFFER_E;
@@ -7306,6 +7308,20 @@ static int wolfSSH_SFTP_GetHandle(WOLFSSH* ssh, byte* handle, word32* handleSz)
         }
     }
 }
+
+
+#ifdef WOLFSSH_TEST_INTERNAL
+/* Drive wolfSSH_SFTP_GetHandle for unit testing so the handle-length bound
+ * against the received payload can be exercised. Returns the function result. */
+int wolfSSH_TestSftpGetHandle(WOLFSSH* ssh, byte* handle, word32* handleSz)
+{
+    if (ssh == NULL || handle == NULL || handleSz == NULL) {
+        return WS_BAD_ARGUMENT;
+    }
+
+    return wolfSSH_SFTP_GetHandle(ssh, handle, handleSz);
+}
+#endif
 
 
 /* Used to get a list of all files and their attributes from a directory.
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -5516,6 +5516,148 @@ static int test_SftpDoName_sizeBound(void)
 
     return 0;
 }
+
+/* Inject an SFTP HANDLE reply declaring 'declHandleLen' with 'payloadSz'
+ * payload bytes, drive wolfSSH_SFTP_GetHandle, and report result, ssh->error,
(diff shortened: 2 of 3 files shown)


tests/unit.c
```diff
@@ -5516,6 +5516,119 @@ static int test_SftpDoName_sizeBound(void)
...
+    if (err != WS_BUFFER_E)
+        return -591;
+
+    /* payload = 8 bytes carrying a 4-byte handle: exact fit, accepted. */
+    ret = WS_FATAL_ERROR;
+    err = WS_FATAL_ERROR;
+    result = sftpGetHandleInject(UINT32_SZ + 4, 4, &ret, &err);
+    if (result != 0)
+        return result;
+    if (ret != WS_SUCCESS)
```
yosuke-wolfssl, line 5648: Fixed

src/wolfsftp.c
```diff
@@ -7308,6 +7310,20 @@ static int wolfSSH_SFTP_GetHandle(WOLFSSH* ssh, byte* handle, word32* handleSz)
 }
 
 
+#ifdef WOLFSSH_TEST_INTERNAL
+/* Drive wolfSSH_SFTP_GetHandle for unit testing so the handle-length bound
```
aidangarske, line 7314: **⚪ [Info] Multi-line header comment on new test hook**
*🔧 NIT*  `style`

The new test hook has a two-line `/* ... */` header comment describing its purpose. Per the wolfSSL comment conventions, function header comments should be one line max and only when a WHY would surprise the reader; a well-named test hook needs none. This is minor and consistent with some surrounding code, so it is take-it-or-leave-it.

**Recommendation:** Optionally collapse to a single-line comment; no functional impact.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Info] **Multi-line header comment on new test hook** — `src/wolfsftp.c:7314-7315`

*Review generated by Skoll.*

src/wolfsftp.c
yosuke-wolfssl: Leave it since it's consistent with some surrounding code.

aidangarske (approved):

ejohnstown (approved):

PR #1084, opened 2026-07-06
Fix multiple reported issues

kareem-wolfssl:


lihnucs (commented):
Hi Kareem,

I have reviewed all 7 commits in this PR and confirmed every fix is correct:

- 1c61fe4 (HIGH-01 / ssh.c): segSz > outSz short-circuit correctly prevents the unsigned underflow before the subtraction.
- 3ae9836 (HIGH-02 / agent.c): Size equality check id->keyBlobSz != keyBlobSz before WMEMCMP closes both the OOB read and the prefix-match auth bypass.
- 8ad669a (HIGH-03 / agent.c): All three defects addressed -- SHOW_SECRETS guard for the log, ForceZero before free, and exact-size heap allocation replacing the 32-byte truncating stack buffer.
- fd36f9a (HIGH-05 / wolfscp.c): strtoull + three-part validation (errno, endptr, > UINT32_MAX) correctly rejects -1 and all out-of-range values.
- 712fe71 (MED-11 / internal.c): skipSz <= len - begin guard on both language-list blocks prevents the begin wraparound.
- 9407efa (HIGH-04 / certman.c): NULL guard added before cm->cm dereference, matching the pattern in every other public CERTMAN API.
- 78390be (MED-01 / port.c): The strnlen-style bounded scan is more robust than a plain strlen -- handles the case where s1 already lacks a null terminator within n bytes.

All 131 CI checks are now passing. The fixes are well-constructed and I have no change requests.

One request before merge: could you please add a Reported-by trailer to each commit (or to the PR description) crediting the original reporter? The full preferred format is:

Reported-by: Asif Nadaf <<postasif@protonmail.com>>
This is the standard git trailer format used in most open-source security fixes and helps ensure the credit is preserved in the git log permanently. I notice a similar commit (989be64 in PR #1080) credited the reporter by name
only -- if that commit is amendable before merge, including the email there as well would be appreciated.

Thank you for the fast turnaround on all nine reports -- this was a thorough and well-executed response.

kareem-wolfssl:
Thanks for the review and feedback @lihnucs.  All of the initial fix commits currently have "Reported-by: Asif Nadaf <[postasif@protonmail.com](mailto:postasif@protonmail.com)>".  This is included on a new line, you will need to expand the commit message to see it.  I also amended #1080 to include this in the commit message before it was merged.

src/internal.c
```diff
@@ -5055,16 +5055,24 @@ static int DoKexInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
-            begin += skipSz;
+        if (ret == WS_SUCCESS) {
+            if (skipSz <= len - begin)
+                begin += skipSz;
+            else
+                ret = WS_BUFFER_E;
+        }
     }
 
     /* First KEX Packet Follows */
```
aidangarske, line 5219: **🟡 [Medium] ML-DSA-only builds do not advertise publickey auth by default**
*💡 SUGGEST*  `bug`

The PR adds ML-DSA public-key authentication paths and compiles `DoUserAuthRequestPublicKey()` when `!defined(WOLFSSH_NO_MLDSA)`, but the default allowed-auth list still adds `publickey` only when RSA or ECDSA is enabled. In an ML-DSA-only build, a server can now process ML-DSA public-key auth, but a normal USERAUTH_FAILURE response will omit `publickey`, so clients may never attempt the newly supported method unless an application overrides `userAuthTypesCb`.

**Recommendation:** Update the compile-time gate to include the public-key algorithms now supported by this file, at least ML-DSA, and add a regression test for an ML-DSA-only default auth list.

src/ssh.c
```diff
@@ -3835,7 +3835,7 @@ int wolfSSH_RealPath(const char* defaultPath, char* in,
         }
```
aidangarske, line 3835: **🟡 [Medium] TPM host key API leaves ctx mutated when registration fails**
*💡 SUGGEST*  `api`

The new API stores `ctx->tpmDev` and `ctx->tpmKey` before checking whether `wolfSSH_SetHostTpmKey()` succeeds. If registration fails, for example because `privateKeyCount` is already at `WOLFSSH_MAX_PVT_KEYS`, the function returns an error but leaves the context partially configured. A later call with a different TPM key then hits the `ctx->tpmKey != NULL && ctx->tpmKey != key` guard and fails even though the original setup did not succeed.

**Recommendation:** Commit TPM device/key pointers only after the host-key slot registration succeeds, or explicitly roll them back on failure.

aidangarske, line 3835: **🟡 [Medium] New ML-DSA ASN.1 public-key path bypasses caller output buffer**
*💡 SUGGEST*  `api`

The ML-DSA public-key branch always allocates a new output buffer and assigns it to `*out`. That means a caller-provided buffer is ignored, undersized caller buffers do not get `WS_BUFFER_E`, and the original `*out` pointer can be overwritten. This is in newly added ML-DSA handling and should either follow the existing `wolfSSH_ReadKey_buffer_ex()` output-buffer contract or be explicitly documented and tested as allocate-only.

**Recommendation:** Compute the required size first, use `*out` when it is non-NULL and large enough, return `WS_BUFFER_E` when it is too small, and allocate only when `*out == NULL`.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **ML-DSA-only builds do not advertise publickey auth by default** — `src/internal.c:17875-17877`
- [Medium] **TPM host key API leaves ctx mutated when registration fails** — `src/ssh.c:540-543`
- [Medium] **New ML-DSA ASN.1 public-key path bypasses caller output buffer** — `src/ssh.c:1878-1892`

*Review generated by Skoll.*

src/internal.c
kareem-wolfssl: I'm not following.  This PR is not touching ML-DSA pubkey auth paths.

src/ssh.c
kareem-wolfssl: Which new API?  This PR is not changing any tpm struct members or APIs.

kareem-wolfssl: Again this PR is not changing any ML-DSA branches/paths, please clarify

src/port.c
```diff
@@ -839,7 +839,18 @@ char* wstrnstr(const char* s1, const char* s2, unsigned int n)
  * end of s1 including a null terminator. */
 char* wstrncat(char* s1, const char* s2, size_t n)
 {
-    size_t freeSpace = n - strlen(s1) - 1;
+    size_t s1_len = 0;
+    size_t freeSpace;
+
+    while (s1_len < n && s1[s1_len] != '\0') {
```
aidangarske, line 845: **🔴 [High] wstrncat change breaks RealPath callers that pass remaining capacity** · _Logic_

The PR changes wstrncat() to return NULL when s1 is not NUL-terminated within the first n bytes. That matches the helper's documented contract and is safer for callers that pass the full destination size and check the return value, but existing changed-file callers do neither. wolfSSH_RealPath() calls WSTRNCAT(out, ..., outSz - curSz) (src/ssh.c:3843 and src/ssh.c:3846) with remaining capacity instead of the full buffer size while out already holds curSz bytes, and ignores the return. After this PR, for a remote SFTP/SCP path with a long first segment, once curSz >= outSz - curSz the existing terminator is beyond the shortened n, so later appends return NULL even when the final path still fits; wolfSSH_RealPath() still increments curSz and returns WS_SUCCESS with a truncated path (separator/segment omitted). The same new failure mode affects other changed-file SCP path construction such as WSTRNCAT((char*)basePath, "/", sizeof("/")) at src/wolfscp.c:2466. This reaches public wolfSSH_RealPath callers used by SFTP/SCP path handling. Severity note: the `review` mode rated this High (BLOCK) as a silent path-truncation bug; `review-security` rated it Medium (CWE-252, unchecked return value); the stricter High rating is kept here.

**Fix:** Keep the bounded scan, but update affected callers to pass the full destination buffer size and check WSTRNCAT() return values. In wolfSSH_RealPath(), include the optional slash in the bounds check and use explicit length-based appends or call WSTRNCAT(out, ..., outSz) after the existing bounds checks, returning WS_INVALID_PATH_E on NULL:

if (curSz + ((curSz != 1) ? 1 : 0) + segSz >= outSz) {
    return WS_INVALID_PATH_E;
}
if (curSz != 1) {
    if (WSTRNCAT(out, "/", outSz) == NULL)
        return WS_INVALID_PATH_E;
    curSz++;
}
if (WSTRNCAT(out, seg, outSz) == NULL)
    return WS_INVALID_PATH_E;

Fix the Nucleus SCP basePath append to use the actual basePath capacity. Add a regression test where a valid resolved path grows beyond half of the output buffer but still fits.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `REQUEST_CHANGES`
**2 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] [review+review-security] **wstrncat change breaks RealPath callers that pass remaining capacity** — `src/port.c:845-850`
- [Medium] [review] **New SCP numeric parsing lacks regression coverage** — `src/wolfscp.c:1106-1303`

src/wolfscp.c
```diff
@@ -1093,19 +1096,35 @@ static int GetScpFileSize(WOLFSSH* ssh, byte* buf, word32 bufSz,
...
-        ssh->scpFileSz = atoi((char *)(buf + idx));
+        /* replace space with newline to terminate the size field, then parse
+         * with strtoull() which parses in 64-bit width, so a negative field
+         * such as "-1" wraps above UINT32_MAX and is rejected by the bound
+         * below instead of becoming a huge word32 size */
+        char* endptr = NULL;
+        word64 fileSz;
 
-        /* restore space, increment idx to space */
+        buf[spaceIdx] = '\n';
```
aidangarske, line 1106: **🟠 [Medium] New SCP numeric parsing lacks regression coverage** · _test_

The PR replaces atoi with strtoull and adds overflow and malformed-field handling for SCP file sizes and timestamps, but the current tests only expose wolfSSH_TestScpGetFileMode; there are no targeted tests for these changed parser paths. This is boundary-heavy protocol parsing, so regressions around negative values, overflow, and partial parses would be easy to miss.

**Fix:** Add WOLFSSH_TEST_INTERNAL wrappers or ReceiveScpMessage-based tests for file sizes -1, UINT32_MAX, UINT32_MAX + 1, empty/non-numeric fields, timestamp overflow, and malformed timestamp fields.

ejohnstown (approved):

PR #1085, opened 2026-07-08
wolfsshd: make the supplementary-group drop reliable and tested

yosuke-wolfssl:
## Summary

`wolfSSHD_AuthSetGroups` performs the daemon's supplementary-group drop — it resolves the target user's full group list and applies it with `setgroups()` before `wolfSSHD_AuthReducePermissionsUser` reduces uid/gid. This PR started as mutation-coverage for that drop (finding f_6701) and, in the process, surfaced and fixes a real privilege-retention bug on macOS/BSD.

## Motivation

**Coverage gap (f_6701).** Deleting the `setgroups()` call, or defeating its `== -1` failure guard, survived the entire test suite. Unlike the `setregid`/`setreuid` path (baseline 5578), there was no `WOLFSSHD_UNIT_TEST` hook for `setgroups`, so no test observed whether it ran. With the call gone, the spawned user session retains the wolfsshd (root) process's supplementary group memberships — a concrete privilege-retention path.

**Bug found while writing the test.** `wolfSSHD_AuthSetGroups` gated the whole `setgroups()` block on a NULL-size `getgrouplist()` probe returning `-1`. That probe is unreliable: on macOS/BSD it returns *success with size 0*, so the block was skipped and **the supplementary-group drop silently never ran** on those platforms. The file already documents this exact unreliability and solves it with grow-and-retry in `wolfSSHD_GetUserGroupNames`.

## Changes

- **Test hooks** — add `WOLFSSHD_UNIT_TEST` callbacks `wsshd_setgroups_cb` and `wsshd_getgrouplist_cb`, alongside the existing `wsshd_setregid_cb`/`wsshd_setreuid_cb`. Each default forwards to the real syscall through a one-line adapter, because `setgroups()`/`getgrouplist()` prototypes differ across platforms (`size_t` vs `int`, `gid_t*` vs `int*`) and a single portable function-pointer type can't bind to them directly.
- **Bug fix / dedup** — extract the grow-and-retry group-list sizing already used by `wolfSSHD_GetUserGroupNames` into a shared `wolfSSHD_GetUserGroupList` helper, and call it from both `GetUserGroupNames` and `AuthSetGroups`. The list is now resolved through one portable path and `setgroups()` runs on every platform.
- **Guard consistency** — standardize the wolfsshd privilege/group functions on the `_WIN32` guard macro (matching the `auth.h` externs and the file's majority convention), so callback definitions and their references share one Windows guard and no `_WIN32 && !WIN32` build can compile a caller without its definition.

## Tests

The tests mock `getgrouplist` **and** `setgroups`, making them deterministic on every platform (the real macOS NULL-size probe would otherwise skip the drop entirely on the dev machine):

- **`test_AuthSetGroups_ok`** — asserts `setgroups()` is invoked with the resolved group count and the exact forwarded gids, after at least one buffer grow-and-retry (the test build forces `-DWOLFSSHD_GROUP_LIST_INIT=1`).
- **`test_AuthSetGroups_setgroups_fail`** — a `setgroups()` failure must propagate `WS_FATAL_ERROR`.
- **`test_AuthSetGroups_getgrouplist_fail`** — a lookup that never succeeds must fail closed and never reach `setgroups()`.

## Verification

**Bug demonstrated & fixed** (real, unmocked, on macOS):

| Path | Result |
| --- | --- |
| Old NULL-size probe | `ret=0 sz=0` → **setgroups SKIPPED** |
| New grow-and-retry | resolves **17 groups** → **setgroups RUNS** |

**Mutation kills** — each defect is caught by exactly the corresponding test (others still pass):

| Mutation | Killed by |
| --- | --- |
| Delete the `setgroups()` call | `test_AuthSetGroups_ok` |
| Defeat the `setgroups()` `== -1` guard | `test_AuthSetGroups_setgroups_fail` |
| Drop the resolve-failure propagation | `test_AuthSetGroups_getgrouplist_fail` |

- Full `wolfsshd` unit suite: **103 passed, 0 failures**; normal build clean, no new warnings.
- Clean under **ASan + UBSan** (`-fsanitize=address,undefined`; LeakSanitizer disabled on Darwin).

## Notes

- `wolfSSHD_GetUserGroupNames` behavior is unchanged (it already used this exact sizing); it now shares the helper instead of duplicating it.
- Windows and QNX paths are preserved (the stub returns `0` on success under QNX to match the `getgrouplist` contract).
- No production control flow changes in non-`WOLFSSHD_UNIT_TEST` builds beyond the sizing fix itself.

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -93,12 +93,20 @@
 #endif
 
 #if defined(WOLFSSHD_UNIT_TEST) && !defined(_WIN32)
+/* Adapts the platform setgroups(2) to a fixed prototype so unit tests can
+ * swap in a stub regardless of the size argument's native type (int on
+ * macOS, size_t on Linux). */
+static int wsshd_setgroups_default(int size, const WGID_T* list)
+{
+    return setgroups(size, list);
+}
 int (*wsshd_setregid_cb)(WGID_T, WGID_T) = setregid;
 int (*wsshd_setreuid_cb)(WUID_T, WUID_T) = setreuid;
 int (*wsshd_setegid_cb)(WGID_T) = setegid;
 int (*wsshd_seteuid_cb)(WUID_T) = seteuid;
 struct passwd* (*wsshd_getpwnam_cb)(const char*) = getpwnam;
 #define getpwnam wsshd_getpwnam_cb
+int (*wsshd_setgroups_cb)(int, const WGID_T*) = wsshd_setgroups_default;
 #endif
 
 struct WOLFSSHD_AUTH {
@@ -1731,7 +1739,7 @@ static int RequestAuthentication(WS_UserAuthData* authData,
                 ret = WOLFSSH_USERAUTH_REJECTED;
             }
             else {
-            #ifdef WIN32
+            #ifdef _WIN32
                 /* Still need to get users token on Windows */
                 wolfSSH_Log(WS_LOG_INFO,
                     "[SSHD] Relying on CA for public key check");
@@ -2092,7 +2100,7 @@ int wolfSSHD_AuthRaisePermissions(WOLFSSHD_AUTH* auth)
 int wolfSSHD_AuthReducePermissionsUser(WOLFSSHD_AUTH* auth, WUID_T uid,
     WGID_T gid)
 {
-#ifndef WIN32
+#ifndef _WIN32
 #ifdef WOLFSSHD_UNIT_TEST
     if (wsshd_setregid_cb(gid, gid) != 0) {
 #else
@@ -2127,7 +2135,7 @@ int wolfSSHD_AuthReducePermissions(WOLFSSHD_AUTH* auth)
     }
 
     flag = wolfSSHD_ConfigGetPrivilegeSeparation(auth->conf);
-#ifndef WIN32
+#ifndef _WIN32
     if (flag == WOLFSSHD_PRIV_SEPARAT || flag == WOLFSSHD_PRIV_SANDBOX) {
         wolfSSH_Log(WS_LOG_INFO, "[SSHD] Lowering permissions level");
 
@@ -2145,11 +2153,23 @@ int wolfSSHD_AuthReducePermissions(WOLFSSHD_AUTH* auth)
     return ret;
 }
 
-#ifndef WIN32
+#ifndef _WIN32
 #if defined(__OSX__) || defined( __APPLE__)
     #define WGETGROUPLIST(x,y,z,w) getgrouplist((x),(y),(int*)(z),(w))
 #else
     #define WGETGROUPLIST(x,y,z,w) getgrouplist((x),(y),(z),(w))
+#endif
+
+#if defined(WOLFSSHD_UNIT_TEST) && !defined(_WIN32)
+/* Adapts getgrouplist(3) to a fixed prototype so unit tests can swap in a
+ * stub, hiding the macOS int/gid_t argument difference behind WGETGROUPLIST. */
+static int wsshd_getgrouplist_default(const char* usr, WGID_T grp,
+        WGID_T* groups, int* ngroups)
+{
+    return WGETGROUPLIST(usr, grp, groups, ngroups);
+}
+int (*wsshd_getgrouplist_cb)(const char*, WGID_T, WGID_T*, int*)
+    = wsshd_getgrouplist_default;
 #endif
 
 /* Initial guess and upper bound for the number of groups a user can be in.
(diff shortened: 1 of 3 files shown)


apps/wolfsshd/auth.c
```diff
@@ -2184,7 +2204,11 @@ int wolfSSHD_AuthSetGroups(const WOLFSSHD_AUTH* auth, const char* usr,
     grpListSz = sysconf( _SC_NGROUPS_MAX );
 #else
     /* should return -1 if grpListSz is smaller than actual groups */
+#ifdef WOLFSSHD_UNIT_TEST
+    if (wsshd_getgrouplist_cb(usr, gid, NULL, &grpListSz) == -1)
+#else
     if (WGETGROUPLIST(usr, gid, NULL, &grpListSz) == -1)
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/test/test_configuration.c
```diff
@@ -1450,6 +1451,79 @@ static void InstallPrivDropStubs(int regidRet, int reuidRet,
...
+    }
+    for (i = 0; i < s_grouplist_count; i++) {
+        groups[i] = s_grouplist_groups[i];
+    }
+    *ngroups = s_grouplist_count;
+    /* simulate the fill call disagreeing with the probed size */
+    if (s_grouplist_fill_mismatch) {
+        return s_grouplist_count + 1;
+    }
+    return s_grouplist_count;
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/auth.c
```diff
@@ -2042,39 +2062,19 @@ int wolfSSHD_AuthReducePermissions(WOLFSSHD_AUTH* auth)
...
- * supplementary, so Match Group can be evaluated against the full set. On
- * success sets *outNames to an owned array of owned names and *outCount to the
- * entry count; the caller frees them with wolfSSHD_FreeUserGroupNames. Returns
- * WS_SUCCESS on success, leaving *outNames NULL on failure. */
-WOLFSSHD_STATIC int wolfSSHD_GetUserGroupNames(void* heap, const char* usr,
-        WGID_T primaryGid, char*** outNames, word32* outCount)
+/* Resolves the user's full gid list into an owned buffer via grow-and-retry
+ * sizing, since a NULL-size probe is unreliable (macOS reports size 0). Sets
+ * *outList (caller frees) and *outCount; returns WS_SUCCESS. */
+static int wolfSSHD_GetUserGroupList(void* heap, const char* usr,
```
aidangarske, line 2190: **🔵 [Low] New group-list helper uses plain `static` while sibling helpers use WOLFSSHD_STATIC**
*💡 SUGGEST*  `convention`

The extracted helper is declared `static int wolfSSHD_GetUserGroupList(...)`, but the adjacent group helpers it was factored out of (`wolfSSHD_GetUserGroupNames`, `wolfSSHD_FreeUserGroupNames`) use the `WOLFSSHD_STATIC` macro so they can be exposed to the unit-test translation unit. The new helper is only reached indirectly (via `AuthSetGroups`/`GetUserGroupNames`), so plain `static` is functionally fine and the tests pass through the public entry points. Flagging only for local consistency — if the intent is that all internal group helpers be reachable/overridable under `WOLFSSHD_UNIT_TEST`, this one diverges.

**Recommendation:** Consider using `WOLFSSHD_STATIC` to match the surrounding helpers, or leave as `static` if direct test access is intentionally not needed. Not blocking.

apps/wolfsshd/auth.c
```diff
@@ -2042,39 +2062,19 @@ int wolfSSHD_AuthReducePermissions(WOLFSSHD_AUTH* auth)
...
-}
-
-/* Builds the list of group names the user belongs to, primary plus
- * supplementary, so Match Group can be evaluated against the full set. On
- * success sets *outNames to an owned array of owned names and *outCount to the
- * entry count; the caller frees them with wolfSSHD_FreeUserGroupNames. Returns
- * WS_SUCCESS on success, leaving *outNames NULL on failure. */
-WOLFSSHD_STATIC int wolfSSHD_GetUserGroupNames(void* heap, const char* usr,
-        WGID_T primaryGid, char*** outNames, word32* outCount)
+/* Resolves the user's full gid list into an owned buffer via grow-and-retry
```
aidangarske, line 2187: **⚪ [Info] Multi-line function-header comments added on new helpers**
*🔧 NIT*  `style`

The new helpers carry 3-line descriptive header comments (`/* Resolves the user's full gid list ... */`, `/* Builds the owned list ... */`, plus the two adapter blocks). This matches the pre-existing style already in this file, so it is consistent with the local convention; noted only because multi-line function headers are generally discouraged and a well-named function often needs none. No change required.

**Recommendation:** Leave as-is for consistency with the existing file, or trim to a single WHY line if touched again.

apps/wolfsshd/test/test_configuration.c
```diff
@@ -1559,6 +1642,134 @@ static int test_AuthReducePermissionsUser_uid_fail(void)
...
+    conf = wolfSSHD_ConfigNew(NULL);
+    if (conf == NULL)
+        ret = WS_FATAL_ERROR;
+    if (ret == WS_SUCCESS) {
+        auth = wolfSSHD_AuthCreateUser(NULL, conf);
+        if (auth == NULL)
+            ret = WS_FATAL_ERROR;
+    }
+
+    InstallGroupStubs(0, &savedGrouplist, &savedSetgroups);
```
aidangarske, line 2031: **⚪ [Info] getgrouplist-fail test drives ~17 malloc/free grow cycles per run**
*🔧 NIT*  `test`

With `s_grouplist_always_fail = 1`, the stub returns -1 on every call, so the helper grows from WOLFSSHD_GROUP_LIST_INIT (1 in the test build) by doubling until it reaches WOLFSSHD_GROUP_LIST_MAX (65536) before returning WS_FATAL_ERROR — ~17 allocate/free iterations. This correctly exercises the MAX-cap fail-closed path and is not incorrect, just mildly wasteful. It is a strength that the cap path is covered; noting only that a test-only lower cap could make intent clearer and the run cheaper.

**Recommendation:** Optional: build the test with a smaller `-DWOLFSSHD_GROUP_LIST_MAX` to make the cap path explicit and faster. No functional issue.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **New group-list helper uses plain `static` while sibling helpers use WOLFSSHD_STATIC** — `apps/wolfsshd/auth.c:2068`
- [Info] **Multi-line function-header comments added on new helpers** — `apps/wolfsshd/auth.c:2065-2067`
- [Info] **getgrouplist-fail test drives ~17 malloc/free grow cycles per run** — `apps/wolfsshd/test/test_configuration.c:1755-1760`

*Review generated by Skoll.*

apps/wolfsshd/auth.c
yosuke-wolfssl: Thanks, but I'm aware of it. It's intentionally kept as static since direct test access is not needed so far.
Some pre-existing small helpers are also kept static as well.

apps/wolfsshd/test/test_configuration.c
yosuke-wolfssl: Thanks — the description is accurate (INIT=1, doubling to MAX=65536 → ~17 iterations, 256KB peak), and agreed that the MAX-cap fail-closed path being covered is the point of that test.

On the cost: the full 103-test suite runs in 0.01–0.02s, so the grow cycles aren't measurable.

On the suggested fix, I tried it and it doesn't work: `WOLFSSHD_GROUP_LIST_MAX` is a compile-time constant shared by `wolfSSHD_GetUserGroupList`, which `test_GetUserGroupNames` also goes through — and that test resolves the *real running user's* groups. Building with a lowered cap:

```
$ make apps/wolfsshd/test/test_configuration CPPFLAGS="-DWOLFSSHD_GROUP_LIST_MAX=8"
$ ./apps/wolfsshd/test/test_configuration
test_GetUserGroupNames FAILED (ret=-1001)     # real user has 17 groups > cap
exit=23

# default build
103 PASSED, exit=0
```

So a test-only lower cap makes `test_GetUserGroupNames` fail closed on any machine whose user belongs to more groups than the cap — environment-dependent CI failures in exchange for a few microseconds. Making it safe would mean turning `MAX` into a runtime-injectable value, which adds production complexity for what's noted as no functional issue.

Happy to revisit if you see a way to scope the cap to just the failure test.

apps/wolfsshd/auth.c
yosuke-wolfssl: Thanks again, but please let me leave it as-is for consistency.

yosuke-wolfssl:
Hi @aidangarske ,
Could you review this again please ?

aidangarske (dismissed):
Looks good.

ejohnstown (approved):

PR #1089, opened 2026-07-09
fix: validate DH GEX request bounds

MarkAtwood:
## Summary

`DoKexDhGexRequest()` in `src/internal.c` reads three client-controlled values from the SSH_MSG_KEX_DH_GEX_REQUEST message — `dhGexMinSz`, `dhGexPreferredSz`, and `dhGexMaxSz` (via three `GetUint32` calls) — and then proceeds to `SendKexDhGexGroup()` without validating them. There is no check that the values satisfy `min <= preferred <= max` and no bounds check against the library's supported DH group sizes.

Per RFC 4419, these requested sizes are folded into the key-exchange hash `H` (`dhGexMinSz` is used when computing the exchange hash). Accepting nonsensical or out-of-range values invites protocol confusion and divergence of `H` between peers.

## Fix

Add a validation block after the three sizes are parsed and before proceeding. It rejects the request (aborts the handshake with `WS_BAD_ARGUMENT`) when:
- `dhGexMinSz < WOLFSSH_DEFAULT_GEXDH_MIN` (1024)
- `dhGexMaxSz > WOLFSSH_DEFAULT_GEXDH_MAX` (8192)
- `dhGexMinSz > dhGexPreferredSz`
- `dhGexPreferredSz > dhGexMaxSz`

The bound macros are the project's own existing defaults from `wolfssh/internal.h`. The request is **rejected** rather than clamped on purpose: because the requested values feed the RFC 4419 exchange hash `H`, silently clamping would make our `H` diverge from the client's and break the handshake in a harder-to-diagnose way.

## Verification

Built with `--enable-all` (`make`, exit 0) against a local wolfSSL install; the modified translation unit compiles cleanly. The affected path is live under `WOLFSSH_NO_DH_GEX_SHA256` not being defined.

Reported by static analysis (Fenrir finding 60).

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -7254,6 +7254,16 @@ static int DoKexDhGexRequest(WOLFSSH* ssh,
     }
 
     if (ret == WS_SUCCESS) {
+        if (ssh->handshake->dhGexMinSz < WOLFSSH_DEFAULT_GEXDH_MIN ||
+            ssh->handshake->dhGexMaxSz > WOLFSSH_DEFAULT_GEXDH_MAX ||
+            ssh->handshake->dhGexMinSz > ssh->handshake->dhGexPreferredSz ||
+            ssh->handshake->dhGexPreferredSz > ssh->handshake->dhGexMaxSz) {
+            WLOG(WS_LOG_DEBUG, "Invalid DH GEX request parameters");
+            ret = WS_BAD_ARGUMENT;
+        }
+    }
+
+    if (ret == WS_SUCCESS) {
         WLOG(WS_LOG_INFO, "  min = %u, preferred = %u, max = %u",
                 ssh->handshake->dhGexMinSz,
                 ssh->handshake->dhGexPreferredSz,


src/internal.c
```diff
@@ -7253,6 +7253,16 @@ static int DoKexDhGexRequest(WOLFSSH* ssh,
         ret = GetUint32(&ssh->handshake->dhGexMaxSz, buf, len, &begin);
     }
 
+    if (ret == WS_SUCCESS) {
```
aidangarske, line 7256: **🟠 [High] DH GEX validation accepts ranges the server cannot satisfy**
*🚫 BLOCK*  `bug`

The new validation accepts any ordered request inside `WOLFSSH_DEFAULT_GEXDH_MIN..WOLFSSH_DEFAULT_GEXDH_MAX`, then immediately calls `SendKexDhGexGroup()`. That send path always emits `dhPrimeGroup14`, a 2048-bit group. A request such as `min=3072, preferred=3072, max=4096` passes the new check but cannot be satisfied by the current server implementation, so the server still sends a group outside the client's requested range. The PR-added validation gate should match what this path can actually produce, or select a supported group within the requested range.

**Recommendation:** Reject unsatisfiable ranges such as `dhGexMinSz > 2048` or `dhGexMaxSz < 2048` while the GEX server path is hard-coded to group14, or implement group selection before sending.

aidangarske, line 7256: **🟡 [Medium] New DH GEX request validation has no regression coverage**
*💡 SUGGEST*  `test`

This PR changes externally visible server behavior by rejecting malformed DH GEX request sizes, but no tests were added. The only existing `wolfSSH_TestDoKexDhGexRequest` coverage found is the first-packet-follows skip case, which intentionally avoids parsing the payload and does not exercise the new validation block.

**Recommendation:** Cover the new rejection paths through `wolfSSH_TestDoKexDhGexRequest()` or an integration-style server handshake test for `min < 1024`, `max > 8192`, `min > preferred`, `preferred > max`, and at least one accepted boundary case.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **DH GEX validation accepts ranges the server cannot satisfy** — `src/internal.c:7256-7263`
- [Medium] **New DH GEX request validation has no regression coverage** — `src/internal.c:7256-7263`

*Review generated by Skoll.*

ejohnstown:
I think I have a pending PR that fixes this already.

ejohnstown:
Closing -- master addresses this under F-55 (`fa21e6e3` "Select DH GEX group from client size window", `2e6b37b2` "Enforce DH GEX floor on client and server"), by a different design than this PR proposes.

`SelectKexDhGexGroup()` (`src/internal.c:13144`) validates the client's window at group-selection time: it clamps `min` up to the 2048-bit `WOLFSSH_DH_GEX_MIN_BITS` floor, caps `max` to the `f` and shared-secret buffers, and returns `WS_DH_SIZE_E` when no built-in group fits. It does that on local copies, so `DoKexDhGexRequest()` still stores the client's values unmodified and the RFC 4419 exchange hash is computed over exactly what the client sent -- the `H`-divergence problem this PR describes, solved without failing the handshake.

Merging this check on top would regress interop. Measured against current master, `[1024, 2048, 8192]` -- what OpenSSH older than 8.2 sends -- goes from "accepted, group 14" to `WS_BAD_ARGUMENT`, as do `[1024, 1024, 8192]`, `[4096, 2048, 8192]`, and `[2048, 2048, 16384]`. The unit tests all pass with the patch applied, so nothing would flag it.

Also worth noting: the description cites a 1024-bit floor, but `WOLFSSH_DEFAULT_GEXDH_MIN` is 2048 on master now; and that macro is what our *client* requests, not server policy -- master keeps the two separate via `WOLFSSH_DH_GEX_MIN_BITS`.

Thanks for the report -- valid finding, it just landed via the DH-GEX rework.

PR #1090, opened 2026-07-09
fix: correct DoChannelData dataSz bounds check

MarkAtwood:
## Bug

In `DoChannelData()` (src/internal.c), the "Validate dataSz" block checks `if (len < begin)` after `GetSize()` has already returned `WS_SUCCESS`. That condition is provably always false: `GetSize`/`GetUint32` only advance `begin` when the requested bytes fit, guaranteeing `begin <= len` on success. So the check is dead code and never validates the payload length against the remaining buffer.

## Fix

Replace the dead check with the meaningful one that the surrounding code already relies on:

```c
    /* Validate dataSz */
    if (ret == WS_SUCCESS) {
        if (dataSz > len - begin) {
            ret = WS_RECV_OVERFLOW_E;
        }
    }
```

`len - begin` cannot underflow because `GetSize` guarantees `begin <= len` on success. This makes the actual payload-length safety invariant explicit for auditors, matching the existing combined pattern used at src/internal.c:7554 (`if ((len < begin) || (strSz > len - begin))`). No behavioral change on valid input — pure defense-in-depth / readability.

This is a LOW severity finding: the following `*idx = begin + dataSz` is already safe on valid input (no word32 overflow), so this is not an exploitable overflow.

## Build verification

Recompiled the full wolfSSH library (including src/internal.c) against a prebuilt wolfSSL (`--enable-all --enable-ssh`) on Ubuntu 24.04, `make` exit 0.

Reported by static analysis (Fenrir finding 247).

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -11090,7 +11090,7 @@ static int DoChannelData(WOLFSSH* ssh,
 
     /* Validate dataSz */
     if (ret == WS_SUCCESS) {
-        if (len < begin) {
+        if (dataSz > len - begin) {
             ret = WS_RECV_OVERFLOW_E;
         }
     }


src/internal.c
```diff
@@ -11090,7 +11090,7 @@ static int DoChannelData(WOLFSSH* ssh,
 
     /* Validate dataSz */
     if (ret == WS_SUCCESS) {
-        if (len < begin) {
+        if (dataSz > len - begin) {
```
aidangarske, line 11093: **🟡 [Medium] New dataSz guard remains unreachable after GetSize**
*💡 SUGGEST*  `bug`

The PR replaces the dead `len < begin` check with `dataSz > len - begin`, but `dataSz` was just read by `GetSize(&dataSz, buf, len, &begin)`. `GetSize()` already returns non-success when `*v > len - *idx`, and `begin` is that same `idx`. Therefore, whenever execution reaches this new `ret == WS_SUCCESS` block, `dataSz <= len - begin` is already guaranteed. Short channel-data payloads still return `WS_BUFFER_E` from `GetSize()` before this branch can run, so the changed code does not add the stated `WS_RECV_OVERFLOW_E` validation path and remains dead validation logic.

**Recommendation:** Clarify the intended error behavior and add a regression test for a truncated `MSGID_CHANNEL_DATA` payload before treating this as a completed bounds-check fix. If the intent is only to rely on existing bounds checking, remove the redundant validation block and point the comment at `GetSize()`. If the intent is to return `WS_RECV_OVERFLOW_E` specifically for an over-declared channel-data payload, parse the length with `GetUint32()` here and keep the explicit `dataSz > len - begin` check, plus a unit case where `dataSz` exceeds the remaining payload length.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **New dataSz guard remains unreachable after GetSize** — `src/internal.c:11093`

*Review generated by Skoll.*

ejohnstown:
Closing -- this is a no-op as written.

`DoChannelData()` reads the length with `GetSize()`, and `GetSize()` (`src/internal.c:4278`) already applies exactly the proposed condition:

```c
result = GetUint32(v, buf, len, idx);
if (result == WS_SUCCESS) {
    if (*v > len - *idx) {
        result = WS_BUFFER_E;
    }
}
```

So on `WS_SUCCESS`, `dataSz > len - begin` is provably false and the edited block is unreachable in both its old and new form. That bound has been in `GetSize` since `4a518018` (Feb 2021), predating this PR's base. Driving `wolfSSH_TestDoChannelData` with truncated payloads on master confirms it: `dataSz=32/len=12`, `dataSz=0xFFFFFFFF/len=12`, and `dataSz=33/len=40` all return `WS_BUFFER_E` from `GetSize` without ever reaching the check.

The cited pattern at `internal.c:7554` is a different situation -- that site reads its length with a raw `ato32()`, so the `strSz > len - begin` test there is load-bearing.

The observation that the block is dead is correct, and we'll clean it up on master separately -- but replacing it with a second dead check reads as if `GetSize` doesn't validate, which is the opposite of what we want an auditor to conclude. Thanks for the report.

PR #1093, opened 2026-07-10
fix: correctness hardening across wolfSSH

MarkAtwood:
Three low-risk correctness fixes flagged by Fenrir static analysis. Build-verified together with `./configure --with-wolfssl=... --enable-all && make -j8` (exit 0).

- **#4582** `src/log.c:52` — the default-logging-cb guard used `WOLFSSL_NO_DEFAULT_LOGGING_CB`, but the body gate (line 104) and the `#else`/`#endif` comments use `WOLFSSH_NO_DEFAULT_LOGGING_CB`. Defining the documented `WOLFSSH_` macro left the guard active, referencing a compiled-out `DefaultLoggingCb`. Aligned the guard to `WOLFSSH_NO_DEFAULT_LOGGING_CB` so it sets `logFunction = NULL`.
- **#4583** `examples/echoserver/echoserver.c:322` (`global_req`) — the return of `wolfSSH_stream_read` was discarded, so the following `if (ret != WS_SUCCESS)` tested the stale `ret` from the earlier `wolfSSH_global_request`. Captured the read's result into `ret`.
- **#6268** `src/wolfsftp.c:9655` (Windows `STATE_GET_OPEN_LOCAL` branch) — `gOfst` is a 2-word array, so `state->gOfst > 0` is always true. Changed to `state->gOfst[0] > 0 || state->gOfst[1] > 0`, matching the MPLAB and POSIX branches.

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -319,7 +319,7 @@ static void *global_req(void *ctx)
             return NULL;
         }
 
-        wolfSSH_stream_read(threadCtx->ssh, buf, 0);
+        ret = wolfSSH_stream_read(threadCtx->ssh, buf, 0);
         if (ret != WS_SUCCESS)
         {
             printf("wolfSSH_stream_read Failed.\n");
--- a/src/log.c
+++ b/src/log.c
@@ -49,7 +49,7 @@
 #endif
 
 
-#ifndef WOLFSSL_NO_DEFAULT_LOGGING_CB
+#ifndef WOLFSSH_NO_DEFAULT_LOGGING_CB
     static void DefaultLoggingCb(enum wolfSSH_LogLevel level,
             const char *const msgStr);
     static wolfSSH_LoggingCb logFunction = DefaultLoggingCb;
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -9652,7 +9652,7 @@ int wolfSSH_SFTP_Get(WOLFSSH* ssh, char* from,
             #elif defined(USE_WINDOWS_API)
                     {
                         DWORD desiredAccess = GENERIC_WRITE;
-                        if (state->gOfst > 0)
+                        if (state->gOfst[0] > 0 || state->gOfst[1] > 0)
                             desiredAccess |= FILE_APPEND_DATA;
                         state->fileHandle = WS_CreateFileA(to, desiredAccess,
                             (FILE_SHARE_DELETE | FILE_SHARE_READ |


examples/echoserver/echoserver.c
```diff
@@ -319,7 +319,7 @@ static void *global_req(void *ctx)
             return NULL;
         }
 
-        wolfSSH_stream_read(threadCtx->ssh, buf, 0);
+        ret = wolfSSH_stream_read(threadCtx->ssh, buf, 0);
```
aidangarske, line 322: **🟠 [High] Captured zero-length stream read now always shuts down global request sessions**
*🚫 BLOCK*  `bug`

The PR changes `wolfSSH_stream_read(threadCtx->ssh, buf, 0)` from a discarded call into the value tested by `if (ret != WS_SUCCESS)`. That call cannot return `WS_SUCCESS`: `wolfSSH_stream_read` explicitly returns `WS_BAD_ARGUMENT` when `bufSz == 0` in `src/ssh.c:1230-1231`. As a result, after every successful `wolfSSH_global_request`, the global-request thread prints `wolfSSH_stream_read Failed.`, calls `wolfSSH_shutdown(threadCtx->ssh)`, and exits. This is introduced by the PR because the previously stale `ret` still held the successful `wolfSSH_global_request` result, while the new assignment makes the invalid zero-length read control the error path.

**Recommendation:** Do not treat the zero-length `wolfSSH_stream_read` result as a successful keepalive/read result. Either remove this invalid read from the checked path, or replace it with a receive/worker path that is valid for processing the global-request reply and handles expected nonfatal statuses such as `WS_CHAN_RXD`, `WS_WANT_READ`, and `WS_WANT_WRITE` according to this thread's blocking mode. Add a `WOLFSSL_TEST_GLOBAL_REQ` regression path that verifies a successful global request does not immediately shut down the SSH session.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Captured zero-length stream read now always shuts down global request sessions** — `examples/echoserver/echoserver.c:322-323`

*Review generated by Skoll.*

ejohnstown:
Closing as redundant -- all three fixes have since landed on master independently, under the same Fenrir finding numbers:

- **F-4582** (`src/log.c` guard `WOLFSSL_` -> `WOLFSSH_`) -- `5dc6c2dc`
- **F-6268** (`src/wolfsftp.c` Windows GET resume check `gOfst > 0` -> `gOfst[0]/[1]`) -- `945f127f`
- **F-4583** (`echoserver.c` stale `ret` after the zero-length read) -- `d18e2824`, which removed the vestigial read entirely.

master already matches this PR's log.c and wolfsftp.c diffs, so nothing here needs to be carried forward. Thanks for the reports.

PR #1094, opened 2026-07-10
fix: buffer write hardening across wolfSSH

MarkAtwood:
Buffer-write hardening across wolfSSH. Three independent fixes reported by Fenrir static analysis, batched into one PR. Build-verified on Ubuntu 24.04 against wolfSSL `--enable-all --enable-ssh` (`./configure --enable-all && make -j8`, exit 0).

- **#3446** — `examples/portfwd/portfwd.c` and `examples/client/common.c`: clamp `passwordSz` to `sizeof(userPassword)` (256) before the `memcpy`. Without the clamp, a `defaultPassword` longer than 256 bytes overflows the fixed `userPassword` buffer. `userPassword` is a raw password buffer whose length travels in `password.passwordSz`, so clamping without a NUL terminator is correct.
- **#4105** — `examples/client/client.c` and `apps/wolfssh/wolfssh.c`: in `readPeer()`'s Windows branch, `WriteFile` now writes `(DWORD)ret` (bytes actually read) instead of the full `bufSz`, matching the POSIX `write(STDOUT_FILENO, buf, ret)` sibling. Prevents emitting stale buffer tail on short reads.
- **#4106** — `src/wolfterm.c` cases `'D'` and `'E'`: `WS_WRITECONSOLE` now writes `sizeof("\n") - 1` (1 byte) instead of `sizeof("\n")` (2 bytes), dropping the stray trailing NUL from the console output.

The #4105 and #4106 sites are under `USE_WINDOWS_API` (not compiled on Linux); they are source-verified against the working POSIX path. The #3446 sites are compiled and covered by the successful Linux build.

### DIFF
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -563,7 +563,7 @@ static THREAD_RET readPeer(void* in)
                 buf[bufSz - 1] = '\0';
 
             #ifdef USE_WINDOWS_API
-                if (WriteFile(stdoutHandle, buf, bufSz, &writtn, NULL) == FALSE) {
+                if (WriteFile(stdoutHandle, buf, (DWORD)ret, &writtn, NULL) == FALSE) {
                     err_sys("Failed to write to stdout handle");
                 }
             #else
--- a/examples/client/client.c
+++ b/examples/client/client.c
@@ -546,7 +546,7 @@ static THREAD_RET readPeer(void* in)
                 buf[bufSz - 1] = '\0';
 
             #ifdef USE_WINDOWS_API
-                if (WriteFile(stdoutHandle, buf, bufSz, &writtn, NULL) == FALSE) {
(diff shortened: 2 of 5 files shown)


ejohnstown:
Closing as redundant -- all three fixes have since landed on master independently, under the same Fenrir finding numbers:

- **F-3446** (clamp `passwordSz` to `sizeof(userPassword)` before the copy, in `examples/client/common.c` and `examples/portfwd/portfwd.c`) -- `ca856589`
- **F-4105** (Windows `WriteFile` writes `(DWORD)ret` not `bufSz`, in `apps/wolfssh/wolfssh.c` and `examples/client/client.c`) -- `f9fc2986`
- **F-4106** (`src/wolfterm.c` cases `'D'`/`'E'` write only the newline byte, not the trailing NUL) -- `7915772f`

master already matches this PR's diffs -- the wolfterm.c sites write a literal `1`, functionally identical to `sizeof("\n") - 1` -- so nothing here needs to be carried forward. Thanks for the reports.

PR #1097, opened 2026-07-10
fix: protocol logic hardening in wolfSSH

MarkAtwood:
Consolidated protocol-logic hardening for wolfSSH (reported by Fenrir static analysis):

- **#4579** — `DoUserAuthRequest` now gates on the advertised auth-type set before running the application `userAuthCb`, so a method the server never advertised is rejected with `USERAUTH_FAILURE` instead of being silently handled.
- **#6271** — `DoChannelRequest` only honors `shell`/`exec`/`subsystem` while `sessionType` is still `UNKNOWN`, enforcing "at most one succeeds".
- **#6515** — a `USERAUTH_REQUEST` received *after* authentication completes is now silently discarded per **RFC 4252 §5.1**, instead of tearing down the connection.
- **#6520** — `DoChannelExtendedData`/`PutBuffer` preserve unread buffered data instead of overwriting it.

Build-verified (`--enable-all`) against wolfSSL; all four applied together compile and link clean.

> Note: the two crypto-**minimum** changes originally bundled here (DH-GEX 1024→2048 and the RSA ≥2048 auth-key floor) are policy/interop decisions and have been **split into #1101** so they can be reviewed on their own.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -766,7 +766,13 @@ INLINE static int IsMessageAllowedServer(WOLFSSH *ssh, byte msg)
         }
     }
     else {
-        if (msg >= MSGID_USERAUTH_REQUEST && msg < MSGID_GLOBAL_REQUEST) {
+        /* RFC 4252 section 5.1: after SSH_MSG_USERAUTH_SUCCESS, a further
+         * SSH_MSG_USERAUTH_REQUEST is allowed past the gate so DoPacket can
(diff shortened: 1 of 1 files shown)


ejohnstown:
Closing: all four findings are wont_fix in Fenrir as of 2026-07-10, #6520's target no longer exists after 50ee1b61, and #6271 fails test_DoChannelRequest on this head.

PR #1099, opened 2026-07-10
fix: force-zero secret buffers before free

MarkAtwood:
Force-zero hardening across wolfSSH: scrub secret-bearing buffers before they are freed or leave scope, and fix one private-key leak on an error path. All changes build-verified together on Ubuntu 24.04 with `./configure --with-wolfssl=<all> --enable-all && make` (exit 0). Reported by Fenrir static analysis.

- **#1278** `src/internal.c` `KeyAgreeEcdhMlKem_client` — ForceZero the ML-KEM private key in `handshake->x` before `wc_MlKemKey_Free`, on both success and error, mirroring the DH path.
- **#2082** `src/agent.c` `wolfSSH_AGENT_ID_free` — replace `WMEMSET` with `ForceZero` for the key buffer and the id struct so the compiler cannot elide the scrub.
- **#2496** `src/internal.c` `SignHRsa`/`SignHEcdsa` — ForceZero `digest` (and `encSig` in SignHRsa) before return.
- **#2497** `src/internal.c` `BuildUserAuthRequest{Rsa,RsaCert,Ecc,EccCert}` — ForceZero `digest`; also `encDigest` (in-scope) for the two RSA paths.
- **#2498** `src/internal.c` `DoUserAuthRequestPublicKey` — ForceZero `digest` before it leaves scope on all exits of the else block.
- **#2500** `src/internal.c` `DoUserAuthRequestRsa` — ForceZero `encDigest` before free (SMALL_STACK) / before scope exit (stack array).
- **#2501** `src/internal.c` `DoUserAuthRequestRsaCert` — same `encDigest` scrub.
- **#2886** `src/internal.c` `SshResourceFree` — ForceZero `ssh->h` and `ssh->sessionId` (and reset their sizes) alongside the already-zeroed KDF inputs before free.
- **#3453** `src/internal.c` `SetHostPrivateKey` — on the CTX-full reject branch, take ownership and `ForceZero`+`WFREE` `der` instead of leaking the private key (no double-free: the store path is unaffected).
- **#3680** `src/internal.c` `ChannelDelete` — ForceZero `inputBuffer.buffer` (NULL-guarded) before free to scrub residual plaintext.
- **#6277** `src/wolfsftp.c` `wolfSSH_SFTP_ClearState` GET/PUT and the GET/PUT `_CLEANUP` cases — ForceZero the SFTP get/put state structs before free (all four sites).
- **#6278** `apps/wolfsshd/wolfsshd.c` POSIX `SHELL_Subsystem` — ForceZero `channelBuffer`/`shellBuffer` on the data-carrying exit.

`ForceZero` is already used throughout these translation units; no new includes.

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -1916,6 +1916,8 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         }
     }
 
+    WS_FORCEZERO(channelBuffer, sizeof channelBuffer);
+    WS_FORCEZERO(shellBuffer, sizeof shellBuffer);
     (void)conn;
     return WS_SUCCESS;
 }
--- a/src/agent.c
+++ b/src/agent.c
@@ -1607,7 +1607,7 @@ void wolfSSH_AGENT_ID_free(WOLFSSH_AGENT_ID* id, void* heap)
             WS_FORCEZERO(id->keyBuffer, id->keyBufferSz);
             WFREE(id->keyBuffer, heap, DYNTYPE_STRING);
         }
-        WMEMSET(id, 0, sizeof(WOLFSSH_AGENT_ID));
+        WS_FORCEZERO(id, sizeof(WOLFSSH_AGENT_ID));
         WFREE(id, heap, DYNTYPE_AGENT_ID);
     }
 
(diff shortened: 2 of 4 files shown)


apps/wolfsshd/wolfsshd.c
```diff
@@ -1916,6 +1916,8 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         }
     }
 
+    wc_ForceZero(channelBuffer, sizeof channelBuffer);
```
JacobBarthelmeh: Can we use the wolfSSH macro for force zero here?

JacobBarthelmeh (changes requested):

apps/wolfsshd/wolfsshd.c
ejohnstown: Updated.

ejohnstown (approved):
The original was good except for rebasing and changing everything to the new WS_FORCEZERO macro.

JacobBarthelmeh (approved):

PR #1101, opened 2026-07-10
fix: raise wolfSSH crypto minimum

MarkAtwood:
This change raises a cryptographic minimum and **rejects peers below it by design** — good hardening, but calling it out explicitly since it affects interop.

- **RSA user-authentication keys must be ≥ 2048 bits** (Fenrir F-6517)
  Per **NIST SP 800-131A**; matches OpenSSH's floor. The `DoUserAuthRequestRsa` path previously enforced *no* minimum (the cert path `DoUserAuthRequestRsaCert` already did). Added the missing check, and introduced a **`WOLFSSH_RSA_MIN_KEY_BITS`** define (default 2048) that **both** RSA auth paths now use, so the minimum is uniform and configurable. Both paths count the modulus bits with `mp_count_bits()` rather than scaling the encrypt size, which rounds up to a byte and would let a 2041-2047 bit key through. Smaller keys fail with `WS_CERT_KEY_SIZE_E`.

**Compatibility:** deployments still using sub-2048-bit RSA keys will be rejected and must regenerate.

Split out from #1097 so the uncontroversial protocol fixes there can merge independently of this policy decision. Build-verified (`--enable-all`) against wolfSSL. Reported by static analysis (Fenrir F-6517).

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8458,6 +8458,28 @@ static int DoUserAuthRequestPassword(WOLFSSH* ssh, WS_UserAuthData* authData,
 }
 
 #ifndef WOLFSSH_NO_RSA
+/* Utility for the RSA user auth paths. */
+/* returns WS_SUCCESS when the key clears WOLFSSH_RSA_MIN_KEY_BITS. */
+static int CheckRsaKeyBits(RsaKey* key)
+{
+    int ret = WS_SUCCESS;
+    int keyBits;
+
+    /* The encrypt size rounds up to a byte; offload builds leave n empty. */
+    keyBits = mp_count_bits(&key->n);
+    if (keyBits == 0) {
+        keyBits = wc_RsaEncryptSize(key) * 8;
+    }
+
+    if (keyBits < WOLFSSH_RSA_MIN_KEY_BITS) {
+        WLOG(WS_LOG_DEBUG, "RSA auth key too small (%d bits)", keyBits);
+        ret = WS_CERT_KEY_SIZE_E;
+    }
+
+    return ret;
+}
+
+
 /* Utility for DoUserAuthRequestPublicKey() */
 /* returns negative for error, positive is size of digest. */
 static int DoUserAuthRequestRsa(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
@@ -8544,6 +8566,10 @@ static int DoUserAuthRequestRsa(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
                 ret = WS_CRYPTO_FAILED;
             }
         }
+    }
+
+    if (ret == WS_SUCCESS) {
+        ret = CheckRsaKeyBits(key);
     }
 
     if (ret == WS_SUCCESS) {
@@ -8705,11 +8731,7 @@ static int DoUserAuthRequestRsaCert(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
     }
 
     if (ret == WS_SUCCESS) {
-        int keySz = wc_RsaEncryptSize(key) * 8;
-        if (keySz < 2048) {
-            WLOG(WS_LOG_DEBUG, "Key size too small (%d)", keySz);
-            ret = WS_CERT_KEY_SIZE_E;
-        }
+        ret = CheckRsaKeyBits(key);
     }
 
     if (ret == WS_SUCCESS) {
--- a/wolfssh/internal.h
+++ b/wolfssh/internal.h
@@ -566,6 +566,13 @@ enum NameIdType {
 #ifndef WOLFSSH_DEFAULT_GEXDH_MIN
     #define WOLFSSH_DEFAULT_GEXDH_MIN 2048
 #endif
+
+#ifndef WOLFSSH_RSA_MIN_KEY_BITS
+    /* Minimum accepted RSA public-key size (bits) for user authentication.
+     * Per NIST SP 800-131A; override at build time if a smaller key must be
+     * accepted. */
+    #define WOLFSSH_RSA_MIN_KEY_BITS 2048
+#endif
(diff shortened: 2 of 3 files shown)


src/internal.c
```diff
@@ -8214,6 +8214,14 @@ static int DoUserAuthRequestRsa(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
         }
     }
 
+    if (ret == WS_SUCCESS) {
```
aidangarske, line 8571: **🔵 [Low] RSA user-auth minimum accepts 2041-2047 bit keys**
*💡 SUGGEST*

The PR adds the missing RSA user-auth key-size check, but computes `keySz` as `wc_RsaEncryptSize(key) * 8`. `wc_RsaEncryptSize()` is a byte-sized RSA block length, so multiplying by 8 rounds the actual modulus bit length up to the next whole byte. A peer-controlled SSH public-key blob with an RSA modulus of 2041 through 2047 bits occupies 256 bytes, is reported as 2048 bits by this check, and bypasses the new `WOLFSSH_RSA_MIN_KEY_BITS` floor. Data Flow: SSH_MSG_USERAUTH_REQUEST public key blob -> `GetMpint()` extracts attacker-supplied `e` and `n` -> `wc_RsaPublicKeyDecodeRaw()` decodes the key -> new byte-rounded size check accepts it -> if the server authorization callback accepts that key and the signature verifies, authentication succeeds with a sub-2048-bit RSA key. This does not bypass possession of the private key, but it does bypass the security minimum this PR is intended to enforce.

**Recommendation:** Measure the exact modulus bit length before comparing against `WOLFSSH_RSA_MIN_KEY_BITS`, using the decoded mpint bytes rather than the RSA block byte size. For example:

static word32 RsaMpintBitLen(const byte* n, word32 nSz)
{
    word32 bits = 0;

    while (nSz > 0 && *n == 0) {
        n++;
        nSz--;
    }
    if (nSz > 0) {
        byte msb = n[0];
        while (msb != 0) {
            bits++;
            msb >>= 1;
        }
        bits += (nSz - 1) * 8;
    }
    return bits;
}

if (ret == WS_SUCCESS) {
    word32 keyBits = RsaMpintBitLen(n, nSz);
    if (keyBits < WOLFSSH_RSA_MIN_KEY_BITS) {
        WLOG(WS_LOG_DEBUG, "RSA auth key too small (%u bits)", keyBits);
        ret = WS_CERT_KEY_SIZE_E;
    }
}

aidangarske, line 8571: **🟡 [Medium] Add regression coverage for the new crypto minimums**
*💡 SUGGEST*  `test`

The PR introduces new user-visible rejection behavior: raw RSA user-auth keys below `WOLFSSH_RSA_MIN_KEY_BITS` now fail with `WS_CERT_KEY_SIZE_E`, and the default DH GEX request minimum changes from 1024 to 2048. Existing tests exercise 2048-bit RSA vectors and explicit DH GEX validation bounds, but no tests assert a sub-2048 RSA auth key is rejected or that the default GEX request now serializes/uses a 2048-bit minimum. That leaves the new policy boundaries easy to regress later.

**Recommendation:** Add focused unit coverage for the new boundaries: one raw RSA public-key auth vector below 2048 bits expecting `WS_CERT_KEY_SIZE_E`, and a DH GEX request/default initialization test that verifies the default minimum is 2048. Add these regression tests before or shortly after merge so the hardening behavior is pinned.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Low] **RSA user-auth minimum accepts 2041-2047 bit keys** — `src/internal.c:8217-8221`
- [Medium] **Add regression coverage for the new crypto minimums** — `src/internal.c:8217-8222, wolfssh/internal.h:513-521`

*Review generated by Skoll.*

src/internal.c
ejohnstown: Fixed.

wolfssh/internal.h
```diff
@@ -511,7 +511,14 @@ enum NameIdType {
...
 #ifndef WOLFSSH_DEFAULT_GEXDH_MIN
-    #define WOLFSSH_DEFAULT_GEXDH_MIN 1024
+    #define WOLFSSH_DEFAULT_GEXDH_MIN 2048
+#endif
+
+#ifndef WOLFSSH_RSA_MIN_KEY_BITS
+    /* Minimum accepted RSA public-key size (bits) for user authentication.
+     * Per NIST SP 800-131A; override at build time if a smaller key must be
+     * accepted. */
+    #define WOLFSSH_RSA_MIN_KEY_BITS 2048
```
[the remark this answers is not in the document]
ejohnstown: Updated. Both paths depend on the same check function.

src/internal.c
```diff
@@ -8214,6 +8214,14 @@ static int DoUserAuthRequestRsa(WOLFSSH* ssh, WS_UserAuthData_PublicKey* pk,
...
+        int keySz = wc_RsaEncryptSize(key) * 8;
+        if (keySz < WOLFSSH_RSA_MIN_KEY_BITS) {
+            WLOG(WS_LOG_DEBUG, "RSA auth key too small (%d bits)", keySz);
+            ret = WS_CERT_KEY_SIZE_E;
+        }
```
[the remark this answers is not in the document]
ejohnstown: Added. test_DoUserAuthRequestRsa now runs the 1024- and 2047-bit vectors, and the mutation check proved they bite.

src/internal.c
ejohnstown: Added. test_DoUserAuthRequestRsa now runs the 1024- and 2047-bit vectors, and the mutation check proved they bite.

The DH-GEX side of this comment is redundant.

holtrop-wolfssl (approved):

PR #1115, opened 2026-07-21
Harden wolfsshd auth tests

stenslae:
test_AuthSetGroups_ relied on the host having a real sshd system user. Now mocks wsshd_getpwnam_cb instead.


apps/wolfsshd/test/test_configuration.c
```diff
@@ -2098,6 +2103,7 @@ static int test_AuthSetGroups_ok(void)
 
     wsshd_getgrouplist_cb = savedGrouplist;
     wsshd_setgroups_cb    = savedSetgroups;
+    wsshd_getpwnam_cb     = savedGetpwnam;
```
stenslae, line 2128: made it so InstallGetpwnamStub is called unconditionally before wolfSSHD_ConfigNew

ejohnstown (approved):

PR #1117, opened 2026-07-22
SSH hardening: buffers, algo lists, auth attempts, parsing

ejohnstown:
- Zeroize input/output buffer contents before freeing -- decrypted traffic (incl. passwords) no longer left in freed heap.
- Algo-list setters now validate each name and return WS_INVALID_ALGO_ID instead of storing it blindly; "none" cipher/MAC only under WOLFSSH_ALLOW_NONE_CIPHER. Builds whose default list would come out empty now fail at compile time naming the macro to define.
- Server caps failed userauth attempts (CTX maxAuthAttempts, default 6), then disconnects.
- DoProtoId compares the peer's identification against a fixed "SSH-2.0-" including the trailing '-', so "SSH-2.01-x" is no longer accepted as an SSH 2.0 speaker.
- GetString rejects a zero-size destination, which wrapped the null-char subtraction to 0xFFFFFFFF and wrote past the buffer. No in-tree caller reached it.

  Issues: F-2473, F-2474, F-2495, F-6819, F-6975

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -3197,22 +3197,23 @@ static int GenerateKeys(WOLFSSH* ssh, byte hashId, byte doKeyPad)
         }
     }
 
-    if (ret == WS_SUCCESS)
+    /* The "none" cipher has no key or IV; skip zero-length derivations. */
+    if (ret == WS_SUCCESS && cK->ivSz > 0)
         ret = GenerateKey(hashId, 'A',
                           cK->iv, cK->ivSz,
                           ssh->k, ssh->kSz, ssh->h, ssh->hSz,
                           ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS && sK->ivSz > 0)
         ret = GenerateKey(hashId, 'B',
                           sK->iv, sK->ivSz,
                           ssh->k, ssh->kSz, ssh->h, ssh->hSz,
                           ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS && cK->encKeySz > 0)
         ret = GenerateKey(hashId, 'C',
                           cK->encKey, cK->encKeySz,
                           ssh->k, ssh->kSz, ssh->h, ssh->hSz,
                           ssh->sessionId, ssh->sessionIdSz, doKeyPad);
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS && sK->encKeySz > 0)
         ret = GenerateKey(hashId, 'D',
                           sK->encKey, sK->encKeySz,
                           ssh->k, ssh->kSz, ssh->h, ssh->hSz,
@@ -3436,7 +3437,9 @@ static const NameIdPair NameIdMap[] = {
 };
 
 
-byte NameToId(const char* name, word32 nameSz)
+/* Look up a name and, on a match, return its category via 'type'. Returns the
+ * ID, or ID_UNKNOWN if the name is not in the table. 'type' may be NULL. */
+static byte NameToIdType(const char* name, word32 nameSz, byte* type)
 {
     byte id = ID_UNKNOWN;
     word32 i;
@@ -3449,11 +3452,19 @@ byte NameToId(const char* name, word32 nameSz)
             XMEMCMP(name, NameIdMap[i].name, nameSz) == 0) {
 
             id = NameIdMap[i].id;
+            if (type != NULL)
+                *type = NameIdMap[i].type;
             break;
         }
     }
 
     return id;
+}
+
+
+byte NameToId(const char* name, word32 nameSz)
+{
+    return NameToIdType(name, nameSz, NULL);
 }
 
 
@@ -3493,6 +3504,79 @@ const char* NameByIndexType(byte type, word32* idx)
     }
 
     return name;
+}
+
+
+int CheckAlgoList(const char* list, byte type)
+{
+    const char* name;
+    word32 nameSz, listSz, i;
+    int usableCount = 0;
+    int ret = WS_SUCCESS;
+
+    if (list == NULL)
+        return WS_INVALID_ALGO_ID;
+
+    listSz = (word32)WSTRLEN(list);
+    name = list;
+    nameSz = 0;
+
+    for (i = 0; i <= listSz; i++) {
+        if (i == listSz || list[i] == ',') {
+            if (nameSz == 0) {
+                /* Only the empty element a single trailing comma leaves is
+                 * legal; AlgoListSz() strips that one. Others would reach
+                 * KEXINIT as a zero-length name, banned by RFC 4251. */
+                if (i != listSz || i == 0) {
+                    ret = WS_INVALID_ALGO_ID;
+                    break;
+                }
+            }
+            else {
+                byte tokenType = TYPE_OTHER;
+                byte id = NameToIdType(name, nameSz, &tokenType);
+
+                if (id == ID_NONE) {
+                    int noneOk = 0;
+
+                    /* "none" names the null cipher/MAC, not a host key. It
+                     * disables the transport, so it needs the build flag. */
+#ifdef WOLFSSH_ALLOW_NONE_CIPHER
+                    if (type == TYPE_CIPHER || type == TYPE_MAC) {
+                        noneOk = 1;
+                    }
+#endif
+                    if (!noneOk) {
+                        ret = WS_INVALID_ALGO_ID;
+                        break;
+                    }
+                    usableCount++;
+                }
+                else if (id != ID_UNKNOWN) {
+                    /* Wrong category is a caller mistake, not a build
+                     * difference. */
+                    if (tokenType != type) {
+                        ret = WS_INVALID_ALGO_ID;
+                        break;
+                    }
+                    usableCount++;
+                }
+                /* Unknown names are skipped, not rejected: callers pass
+                 * portable supersets, and negotiation ignores them too. */
+            }
+            name = list + i + 1;
+            nameSz = 0;
+        }
+        else {
+            nameSz++;
+        }
+    }
+
+    /* Nothing usable left; fail here, not as a later KEX failure. */
+    if (ret == WS_SUCCESS && usableCount == 0)
+        ret = WS_INVALID_ALGO_ID;
+
+    return ret;
 }
 
 
@@ -4542,6 +4626,10 @@ static byte MatchIdLists(int side, const byte* left, word32 leftSz,
 static INLINE byte BlockSzForId(byte id)
 {
     switch (id) {
+        case ID_NONE:
+            /* RFC 4253 still pads to 8 under the null cipher, and a 0 here
+             * divides by zero in BundlePacket() once NEWKEYS installs it. */
+            return MIN_BLOCK_SZ;
 #ifndef WOLFSSH_NO_AES_CBC
         case ID_AES128_CBC:
         case ID_AES192_CBC:
@@ -7774,11 +7862,16 @@ static int DoExtInfoServerSigAlgs(WOLFSSH* ssh,
     byte algoId;
 
     peerSigIdSz = CountNameList(names, namesSz);
-    if (peerSigIdSz > 0) {
-        peerSigId = (byte*)WMALLOC(peerSigIdSz, ssh->ctx->heap, DYNTYPE_ID);
-        if (peerSigId == NULL) {
-            ret = WS_MEMORY_E;
-        }
+    if (peerSigIdSz == 0) {
+        /* Treat an empty list as no extension. GetNameListRaw() would reject
+         * the NULL idList and drop the connection. */
+        WLOG(WS_LOG_DEBUG, "DEISSA: peer sent an empty server-sig-algs");
+        return WS_SUCCESS;
+    }
+
+    peerSigId = (byte*)WMALLOC(peerSigIdSz, ssh->ctx->heap, DYNTYPE_ID);
+    if (peerSigId == NULL) {
+        ret = WS_MEMORY_E;
     }
 
     if (ret == WS_SUCCESS) {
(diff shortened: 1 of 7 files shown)


src/internal.c
```diff
@@ -3494,6 +3507,79 @@ const char* NameByIndexType(byte type, word32* idx)
...
+    int usableCount = 0;
+    int ret = WS_SUCCESS;
+
+    if (list == NULL)
+        return WS_INVALID_ALGO_ID;
+
+    listSz = (word32)WSTRLEN(list);
+    name = list;
+    nameSz = 0;
+
```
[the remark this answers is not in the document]
ejohnstown: Declined. The build will error if all MACs are disabled.

wolfssh/ssh.h
```diff
@@ -104,6 +104,13 @@ WOLFSSH_API int wolfSSH_ReadKey_file(const char* name,
         byte** out, word32* outSz, const byte** outType, word32* outTypeSz,
         byte* isPrivate, void* heap);
 
+/* SetAlgoList* validate the list, returning WS_SUCCESS or WS_INVALID_ALGO_ID.
+ * Kex/Cipher/Mac reject NULL. Key accepts NULL only on a server, restoring the
+ * default of deriving the host key list from the loaded private keys; a client
+ * has no such fallback. KeyAccepted accepts NULL on either side, but that
+ * empties the list rather than restoring a default: the server then advertises
+ * an empty RFC 8308 "server-sig-algs", telling clients it accepts no signature
+ * algorithms, which stops OpenSSH clients offering rsa-sha2-256/512. */
```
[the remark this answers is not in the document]
ejohnstown: fixed

tests/unit.c
```diff
@@ -3457,6 +3457,112 @@ static int test_DoChannelRequest(void)
...
+        idx += UINT32_SZ;
+        WMEMCPY(buf + idx, fields[i], fieldSz);
+        idx += fieldSz;
+    }
+
+    buf[idx++] = 0;               /* no password change */
+    fieldSz = (word32)WSTRLEN("badpass");
+    PutU32BE(buf + idx, fieldSz);
+    idx += UINT32_SZ;
+    WMEMCPY(buf + idx, "badpass", fieldSz);
```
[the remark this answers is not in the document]
ejohnstown: Correct that the check was missing though not reachable today (the only caller passes a 128-byte buffer for a fixed 50-byte payload); added the check anyway so the helper can't overflow instead of returning 0.

src/internal.c
```diff
@@ -3494,6 +3507,79 @@ const char* NameByIndexType(byte type, word32* idx)
...
+                        ret = WS_INVALID_ALGO_ID;
+                        break;
+                    }
+                    usableCount++;
+                }
+                else if (id != ID_UNKNOWN) {
+                    /* Wrong category is a caller mistake, not a build
+                     * difference. */
+                    if (tokenType != type) {
+                        ret = WS_INVALID_ALGO_ID;
```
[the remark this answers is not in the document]
ejohnstown: Accurate but intentional and pre-existing: the lists are deliberately permissive so callers can pass one portable superset across builds with different algorithm sets, and rejecting or silently filtering unknown tokens would be a behavior change beyond this PR's scope.

src/internal.c
```diff
@@ -7837,6 +7947,46 @@ static int DoExtInfo(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+    ret = SendUserAuthFailure(ssh, partialSuccess);
+    if (count) {
+        int cntRet = CountUserAuthFailure(ssh);
+
+        if (cntRet != WS_SUCCESS)
+            ret = cntRet;
+    }
+
+    return ret;
+}
```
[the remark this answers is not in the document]
ejohnstown: Not a problem here, no change made. `PreparePacket()` appends at
`outputBuffer.length`, past the unsent region `[idx, length)`, and
`GrowBuffer()` compacts by copying from `buffer + idx`. The pending
USERAUTH_FAILURE is preserved and the DISCONNECT is appended after it, so both
flush in order. That's the normal queued-send path, not a special case, so no
deferred-disconnect flag is needed.

tests/api.c
```diff
@@ -3677,6 +3722,22 @@ static void test_wolfSSH_SetAlgoList(void)
...
+    queryIdx = 0; newCipherList = wolfSSH_QueryCipher(&queryIdx);
+    queryIdx = 0; newMacList = wolfSSH_QueryMac(&queryIdx);
+    newKeyAccList = newKeyList;
+    AssertNotNull(newKexList);
+    AssertNotNull(newKeyList);
+    AssertNotNull(newCipherList);
+    /* An AEAD-only build has no MAC names to query. */
+    if (newMacList == NULL) {
+        return;
+    }
```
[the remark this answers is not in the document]
ejohnstown: Removed, though as dead code rather than for the reason given. An
AEAD-only build doesn't exist: `wolfssh/internal.h` already makes zero MAC
algorithms a hard `#error`, and `NameByIndexType()` scans the same table those
macros gate, so `wolfSSH_QueryMac()` can't return NULL in any build that
compiles. The guard was skipping nothing while hiding the KEX/KEY/CIPHER
coverage behind it. Replaced with `AssertNotNull(newMacList)`.

tests/api.c
```diff
@@ -3834,17 +3896,153 @@ static void test_wolfSSH_SetAlgoList(void)
...
+    AssertNotNull(validCipher);
+    AssertNotNull(aKexName);
+
+    /* Every built-in default must pass its own setter, in any build config. */
+    AssertIntEQ(wolfSSH_CTX_SetAlgoListKex(ctx,
+            wolfSSH_CTX_GetAlgoListKex(ctx)), WS_SUCCESS);
+    AssertIntEQ(wolfSSH_CTX_SetAlgoListCipher(ctx,
+            wolfSSH_CTX_GetAlgoListCipher(ctx)), WS_SUCCESS);
+    AssertIntEQ(wolfSSH_CTX_SetAlgoListMac(ctx,
+            wolfSSH_CTX_GetAlgoListMac(ctx)), WS_SUCCESS);
```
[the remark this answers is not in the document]
ejohnstown: Confirmed, but the mechanism is the empty string, not NULL, and the root cause
is in the library. `cannedMacAlgoNames` only admits the SHA-1 MACs under
`WOLFSSH_NO_SHA1_SOFT_DISABLE`, off by default, so
`-DWOLFSSH_NO_HMAC_SHA2_256 -DWOLFSSH_NO_HMAC_SHA2_512` leaves the default list
`""`, which `CheckAlgoList()` rejects with `WS_INVALID_ALGO_ID`. Same gap on
the cipher side: `-DWOLFSSH_NO_AES_GCM -DWOLFSSH_NO_AES_CTR` empties
`cannedEncAlgoNames` via the AES-CBC soft disable and fails line 3941.

An empty default list advertises nothing in KEXINIT, so negotiation can never
succeed -- the new validation was correctly surfacing a broken build config
that previously failed later as `WS_MATCH_MAC_ALGO_E`. The "you need at least
one X" guards checked the implementation macros but not the soft-disable macros
that gate list membership; extends all four (KEX, signing, cipher,
MAC) to fail the build and name the macro to define. Both configs now stop at
compile time, and both pass with the soft disable lifted.

Left the assertion as-is rather than gating it, so the "any build config"
invariant stays true.

src/internal.c
```diff
@@ -7837,6 +7957,46 @@ static int DoExtInfo(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+
+
 #ifdef WOLFSSH_ALLOW_USERAUTH_NONE
```
aidangarske, line 8000: **🟠 [Medium] Opening none auth probe is counted when userauth-none support is compiled in** · _Logic_

The new max-auth-attempts API documents that the opening `none` probe is exempt, and the unsupported-method branch in DoUserAuthRequest implements that exemption via `countIt = !(authNameId == ID_NONE && ssh->authRequests == 1)`. However, when built with `WOLFSSH_ALLOW_USERAUTH_NONE`, `ID_NONE` is routed earlier to `DoUserAuthRequestNone()`, bypassing the exemption; every rejected/failed/missing-callback none attempt there calls `SendUserAuthFailureCount(..., 1)`. A server compiled with `WOLFSSH_ALLOW_USERAUTH_NONE` and `maxAuthAttempts=1` can disconnect a standard client after its initial method-discovery `none` request, before a real password/public-key attempt; with the default limit, users get one fewer real attempt than configured. Both scans agree on this off-by-one; the bugs scan rated it Medium and the security scan rated it Low — the stricter Medium is kept.

**Fix:** Apply the same first-`none` exemption inside the userauth-none handler, e.g. compute `int countIt = !(ssh->authRequests == 1);` for rejected/failure paths and pass it to `SendUserAuthFailureCount()`, or thread a count flag into `DoUserAuthRequestNone()`. Add a `WOLFSSH_ALLOW_USERAUTH_NONE` unit test with `maxAuthAttempts=1` proving a rejected opening none probe neither disconnects nor consumes a configured real attempt.

src/internal.c
```diff
@@ -9611,12 +9779,27 @@ static int DoUserAuthRequest(WOLFSSH* ssh,
...
         ssh->authId = authNameId;
+        ssh->authRequests++;
 
+        /* Each method branch counts its own failures; it can't be centralized
+         * here since handlers return WS_SUCCESS for both success and
+         * failure-sent. */
         if (authNameId == ID_USERAUTH_PASSWORD)
             ret = DoUserAuthRequestPassword(ssh, &authData, buf, len, &begin);
 #ifdef WOLFSSH_KEYBOARD_INTERACTIVE
         else if (authNameId == ID_USERAUTH_KEYBOARD) {
```
aidangarske, line 9790: **🟠 [Medium] Keyboard-interactive restarts only count when the next request is also keyboard** · _State Management_

The PR adds `kbSetupPending` so abandoning an outstanding keyboard-interactive `INFO_REQUEST` is charged as a failed auth attempt, but the check lives only inside the `ID_USERAUTH_KEYBOARD` branch. If a client starts keyboard-interactive, receives an `INFO_REQUEST`, then sends a different `USERAUTH_REQUEST` (password or unsupported `none`), the abandoned keyboard attempt is not counted. Even for a second keyboard request, the `counted` flag is passed into `SendUserAuthKeyboardRequest`; if the new setup callback rejects, the current failed request is not counted because the prior pending exchange was counted instead. This lets peers exceed `maxAuthAttempts` by interleaving abandoned keyboard exchanges with other requests or rejected keyboard setups.

**Fix:** Charge a pending keyboard exchange before method dispatch rather than only inside the keyboard branch, and do not let that prior charge suppress counting a new setup rejection. For example, clear/count `kbSetupPending` immediately after parsing `authNameId`; if the limit is not reached, dispatch normally and call `SendUserAuthKeyboardRequest(..., 0)` so a declined setup callback still counts the current request.

tests/api.c
```diff
@@ -3834,17 +3893,153 @@ static void test_wolfSSH_SetAlgoList(void)
...
+                validCipher, secondCipher);
+        AssertIntEQ(wolfSSH_CTX_SetAlgoListCipher(ctx, listBuf),
+                WS_INVALID_ALGO_ID);
+    }
+
+    /* "none" names no host key, and needs the build flag for cipher/MAC. */
+    AssertIntEQ(wolfSSH_CTX_SetAlgoListKey(ctx, "none"), WS_INVALID_ALGO_ID);
+    AssertIntEQ(wolfSSH_CTX_SetAlgoListKeyAccepted(ctx, "none"),
+            WS_INVALID_ALGO_ID);
+#ifndef WOLFSSH_ALLOW_NONE_CIPHER
```
aidangarske, line 4007: **🟠 [Medium] Add positive coverage for --enable-none-cipher** · _test_

The PR adds `WOLFSSH_ALLOW_NONE_CIPHER` support in `CheckAlgoList()`, `BlockSzForId(ID_NONE)`, and key generation, but the test only asserts rejection when the flag is disabled. When `WOLFSSH_ALLOW_NONE_CIPHER` is enabled, these assertions compile out and there is no positive check that cipher/MAC `none` is accepted or that the negotiated packet/key path works.

**Fix:** Add an `#else` branch asserting `wolfSSH_CTX_SetAlgoListCipher(ctx, "none")` and `wolfSSH_CTX_SetAlgoListMac(ctx, "none")` return `WS_SUCCESS`, plus a small regression case that negotiates or initializes the `ID_NONE` cipher/MAC path, so the new optional feature cannot silently break.

tests/unit.c
```diff
@@ -3627,6 +3769,114 @@ static int test_DoChannelRequest(void)
...
+
+/* Verify the failed-userauth cap actually fires, and that it fires whether or
+ * not the USERAUTH_FAILURE send completes. The want-write case is the
+ * regression: WS_WANT_WRITE is non-fatal in DoReceive(), so a peer that stops
+ * reading its socket must not be able to pipeline uncounted guesses.
+ *
+ * Asserts, with the limit set to 3: attempts 1 and 2 leave the connection up
+ * (WS_SUCCESS, or WS_WANT_WRITE when the send blocks) and attempt 3 returns
+ * WS_USER_AUTH_E, ending the accept loop. */
+static int test_MaxAuthAttempts(void)
```
aidangarske, line 3823: **🟠 [Medium] Auth-attempt cap tests only cover password failures** · _test_

The new failure-counting code modifies password, publickey, keyboard-interactive, invalid-method, and `none` handling, including the new `kbSetupPending` branch. The regression test drives only a password request, so the newly added counting paths in `DoUserAuthRequestPublicKey()`, `DoUserAuthInfoResponse()`, and `SendUserAuthKeyboardRequest()` can regress without this test failing.

**Fix:** Add focused cases for keyboard-interactive restart before `INFO_RESPONSE`, keyboard setup rejection, publickey bad-signature or partial-success failure, and the opening `none` exemption, broadening the max-auth regression coverage beyond password failures.

aidangarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review-security+bugs] **Opening none auth probe is counted when userauth-none support is compiled in** — `src/internal.c:8000-8043`
- [Medium] [bugs] **Keyboard-interactive restarts only count when the next request is also keyboard** — `src/internal.c:9790-9802`
- [Medium] [review] **Add positive coverage for --enable-none-cipher** — `tests/api.c:4007-4010`
- [Medium] [review] **Auth-attempt cap tests only cover password failures** — `tests/unit.c:3823-3864`

JacobBarthelmeh (approved):

src/internal.c
ejohnstown: Fixed in #1127.

src/internal.c
ejohnstown: Fixed in #1127.

tests/api.c
ejohnstown: Fixed in #1127.

tests/unit.c
ejohnstown: Fixed in #1127.

PR #1118, opened 2026-07-22
wolfsshd: load PKCS#8 PEM host keys (e.g. ML-DSA)

shaunchokshi:
`wolfsshd` could not load a host key stored as a PKCS#8 `-----BEGIN PRIVATE KEY-----` PEM file (the form emitted for ML-DSA keys, and by `openssl genpkey`).

`SetupCTX()` pre-converted the file with `wc_PemToDer(..., PRIVATEKEY_TYPE, ...)`, which only recognizes the classic `RSA/EC PRIVATE KEY` headers. On a PKCS#8 body that call returns success but produces a malformed DER (leading `0x04` rather than a `0x30` SEQUENCE); the bytes were then passed to `wolfSSH_CTX_UsePrivateKey_buffer()` as `WOLFSSH_FORMAT_ASN1`, which rejected them with `WS_BAD_FILETYPE_E` (-1016).

Detect PEM vs DER by content and decode PEM with `wc_KeyPemToDer()`, which handles PKCS#1, SEC1 and PKCS#8 private-key bodies. DER keys are unchanged.

### Testing

Built against wolfSSL master; same `wolfsshd` binary run against three host keys with the patch toggled on and off, reading the `wolfSSH_CTX_UsePrivateKey_buffer()` return from `-D -d` output:

| Host key | Before | After |
|---|---|---|
| ML-DSA-65 PEM (PKCS#8) | `ret = -1016`, "Failed to use host private key" | `ret = 0` |
| ML-DSA-65 DER | `ret = 0` | `ret = 0` (no regression) |
| ECDSA SEC1 PEM | `ret = 0` | `ret = 0` (no regression) |

The ML-DSA PEM key was generated with `wolfssl genkey ml-dsa -level 3 -output priv -outform PEM`.

No `ChangeLog.md` entry is included — the top section is the released v1.5.0; happy to add one wherever you'd like it. Suggested wording: "Fixed loading PKCS#8 PEM host keys in wolfSSHd (e.g. ML-DSA)."

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -349,9 +349,8 @@ static int SetupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
         byte** banner)
 {
     int ret = WS_SUCCESS;
-    DerBuffer* der = NULL;
-    byte* privBuf;
-    word32 privBufSz;
+    byte* privBuf = NULL;
+    word32 privBufSz = 0;
     void* heap = NULL;
 
     if (ctx == NULL) {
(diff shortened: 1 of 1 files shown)


dgarske:
See ZD 22185

apps/wolfsshd/wolfsshd.c
```diff
@@ -412,28 +411,57 @@ static int SetupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
...
+                if (ret == WS_SUCCESS
+                        && wolfSSH_CTX_UsePrivateKey_buffer(*ctx, privBuf,
+                                privBufSz, WOLFSSH_FORMAT_ASN1) < 0) {
                     wolfSSH_Log(WS_LOG_ERROR,
                         "[SSHD] Failed to use host private key.");
                     ret = WS_BAD_ARGUMENT;
                 }
 
+                if (keyDer != NULL) {
+                    wc_ForceZero(keyDer, dataSz);
```
ejohnstown: use WS_FORCEZERO()

apps/wolfsshd/wolfsshd.c
```diff
@@ -349,7 +349,6 @@ static int SetupCTX(WOLFSSHD_CONFIG* conf, WOLFSSH_CTX** ctx,
         byte** banner)
 {
     int ret = WS_SUCCESS;
-    DerBuffer* der = NULL;
     byte* privBuf;
```
ejohnstown: Initialize these to NULL/0. There's a path where they can end up used wihtout a value.

ejohnstown (changes requested):

apps/wolfsshd/wolfsshd.c
ejohnstown: The Windows build tests are catching this.

apps/wolfsshd/wolfsshd.c
shaunchokshi: Done — switched to WS_FORCEZERO() in 6d6373e.

apps/wolfsshd/wolfsshd.c
shaunchokshi: Done — privBuf/privBufSz now initialized to NULL/0 in 6d6373e, which also clears the uninitialized-use path the Windows build was catching.

ejohnstown (changes requested):
I can't accept that commit from Claude. Can you squash it into the other commit and force-push it again, please?

shaunchokshi:
Squashed to a single commit, and addressed both Copilot findings in it: empty (0-byte) host-key files now return a file error before the PEM path (no WMALLOC(0)), and the raw key buffer is zeroized before free alongside the decoded DER.  Thanks for the patience!

ejohnstown (approved):

anhu:
Contributor agreement has been accepted.

Issue #1120, opened 2026-07-22
wolfsshd: ML-DSA host key cannot complete KEX — wc_MlDsaKey_ExportPubRaw returns BAD_FUNC_ARG in SendKexDhReply


shaunchokshi:
An ML-DSA-65 host key loads into `wolfsshd` and `ssh-mldsa-65` is negotiated by both peers, but the handshake then fails server-side while building the KEXDH reply. The result is that an ML-DSA host key can be loaded and advertised but cannot sign the KEX, so no client can complete a handshake against it.

I'm reporting rather than patching — the fix looks like it belongs either in wolfCrypt (derive/retain the public key) or in how wolfSSH obtains the host public key, and that seemed like a maintainer call rather than something to guess at in a signing path.

### Environment

- wolfSSH `master` @ `53047bd`
- wolfSSL `master` @ `793608e` (5.9.2)
- Linux 6.x (CachyOS), gcc 16.1.1
- wolfSSH configured with `--enable-sshd --enable-shell --enable-scp --enable-sftp --enable-fwd`

### Reproduction

1. Generate an ML-DSA-65 host key with wolfCLU: `wolfssl genkey ml-dsa -level 3 -out h -output priv`
2. Point a `wolfsshd` config's `HostKey` at it and run `wolfsshd -D -d -f <config>`
3. Connect with the wolfSSH example client: `examples/client/client -h 127.0.0.1 -p <port> -u <user> -P <pass>`

### Observed

```
[SSHD] Adding name : ssh-mldsa-65
DKI: Server Host Key Algorithms
GNL: name ID ssh-mldsa-65 matches ssh-mldsa-65        # client offered it too
Using ML-DSA Host key
Leaving SendKexDhReply(), ret = -173                  # BAD_FUNC_ARG
[SSHD] Failed to accept WOLFSSH connection ... error -1001
```

### Root cause

The failure is in the ML-DSA block of `SendKexDhReply()` (`src/internal.c`, ~L13264-13279). `wc_MlDsaKey_PrivateKeyDecode()` succeeds, but the immediately following `wc_MlDsaKey_ExportPubRaw()` returns `BAD_FUNC_ARG`: the key decoded from a private-key file has no public-key component available to export, and wolfCrypt does not derive it here.

[...]

PR #1121, opened 2026-07-22
Added Public Key Validation for Decoded Private Host Keys

stenslae:
https://github.com/wolfSSL/wolfssh/issues/1120

ML-DA & Ed25519: If a private key is decoded, but public key is not exported, an error is logged.
ECDSA: If a private key lacks a derivable public key, a warning is output.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -1690,12 +1690,18 @@ void wolfSSH_KEY_clean(WS_KeySignature* key)
  * fails try to load it as if ECDSA. Both public and private keys can be
  * decoded. For RSA keys, the key format is described as "ssh-rsa".
  *
+ * Private-only ML-DSA keys are rejected (WS_CRYPTO_FAILED) as public keys
+ * cannot be derived. ECDSA derives and validates the public key here.
+ * Ed25519 allows missing public keys if HAVE_ED25519_MAKE_KEY is defined
+ * (derived later at KEX); otherwise rejected like ML-DSA.
+ *
  * @param in        key to identify
  * @param inSz      size of key
  * @param isPrivate indicates private or public key
  * @param heap      heap to use for memory allocation
  * @param pkey      optionally return populated WS_KeySignature
- * @return          keyId as int, WS_MEMORY_E, WS_UNIMPLEMENTED_E
+ * @return          keyId as int, WS_MEMORY_E, WS_UNIMPLEMENTED_E,
+ *                  WS_CRYPTO_FAILED
  */
 int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
     WS_KeySignature **pkey)
@@ -1704,6 +1710,9 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
     word32 idx;
     int ret;
     int dynType = isPrivate ? DYNTYPE_PRIVKEY : DYNTYPE_PUBKEY;
+    /* Set to WS_CRYPTO_FAILED if ML-DSA key lacks derivable public key.
+     * Prevents Ed25519 fallback decode. */
+    int noPubKeyRet = 0;
 #ifndef WOLFSSH_NO_MLDSA
     byte mlDsaLevel = 0;
     int mlDsaInit = 0;
@@ -1765,6 +1774,8 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
 
                 /* If decode was successful, this is an ECDSA key. */
                 if (ret == 0) {
+                    int curveSupported = 1;
+
                     switch (wc_ecc_get_curve_id(key->ks.ecc.key.idx)) {
                         case ECC_SECP256R1:
                             key->keyId = ID_ECDSA_SHA2_NISTP256;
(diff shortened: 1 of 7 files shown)


ejohnstown (commented):
Tested this on Ubuntu 26.04 against wolfSSL master (ac01707f5) built with
`--enable-all --enable-experimental --enable-mldsa`, wolfSSH configured
`--enable-all --enable-sshd`. Clean build (no new warnings), `make check`
passes 10/10 with 1 skip, and valgrind is clean on the new reject paths
(all heap blocks freed, 0 errors) as well as on the ECDSA warn path and a
successful ML-DSA handshake.

Confirmed the fix does what it says. Using a private-only ML-DSA-65 DER
(4060 bytes, from `wc_MlDsaKey_PrivateKeyToDer`), master starts the daemon
and then fails with `SendKexDhReply(), ret = -173` and accept error -1001,
exactly as reported in #1120. With this PR, wolfsshd refuses at startup
with -1029 and the new message. Same story for a 48-byte private-only
Ed25519 key: -134 at KEX on master, clean load-time rejection here.

Worth recording that neither rejection is a regression. A private-only
Ed25519 key cannot sign at all (`wc_ed25519_sign_msg` returns -173), so it
was never a usable host key, and the ML-DSA advice in the message is
accurate: `wc_MlDsaKey_PrivateKeyDecode` handles a seed-encoded key via
`MakeKeyFromSeed`, which produces the public half. The repo's own
`keys/server-key-mldsa*.der` and `server-key-ed25519.der` all carry public
keys, so nothing existing is affected.

Two things I would like addressed before this goes in.

1. The `MLDSA_MAX_PUB_KEY_SIZE` scratch buffer on the stack. It takes
   `IdentifyAsn1Key`'s frame from about 224 bytes to about 2736 (measured:
   `sub sp, sp, #0xa50`; the compiler folds the ECC and Ed25519 buffers into
   the ML-DSA one). Everywhere else in internal.c that needs a key block
   this size heap-allocates it (see the sigKeyBlock allocations around
   internal.c:6717 and internal.c:14653), which keeps `WOLFSSH_SMALL_STACK`
   builds viable. `WS_KeySignature.ks.mldsa` holds only the `MlDsaKey` and
   has no `q` field to reuse, so this wants a WMALLOC/WFREE pair around the
   export.

2. The ECDSA path is warn-only, so it does not actually meet the stated
   goal. I built a SEC1 key with the optional public point omitted: behavior
   is identical to master, still `SendKexDhReply(), ret = -246` at handshake
   time, just with a log line beforehand. The X.509 pairing rationale in the
   comment is plausible but is not tested here and I did not verify it
   either. Either confirm that claim and note the asymmetry, or reject
   ECDSA the same way. Separately, the warning text ("this only works if
   paired with an X.509 certificate") is misleading on the client user-auth
   path, which goes through this same function.

Minor: no test covers the new behavior. A private-only ML-DSA DER is a few
lines to generate with `wc_MlDsaKey_PrivateKeyToDer`, and would lock in the
load-time rejection.

stenslae:
Fixed suggestion 1 and test coverage. For suggestion 2, when ECDSA is paired with a cert, we don't need to derive the public key. I did update the comment for suggestion 2.

src/internal.c
```diff
@@ -1865,6 +1923,23 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
...
                     ret = wc_Ed25519PrivateKeyDecode(in, &idx,
                             &key->ks.ed25519.key, inSz);
+                    if (ret == 0) {
+                        /* Same gap as ML-DSA/ECDSA: priv-only decode can
+                         * succeed with no derivable public key; reject here
+                         * instead of at first handshake. */
+                        byte pubScratch[ED25519_PUB_KEY_SIZE];
+                        word32 pubScratchSz = sizeof(pubScratch);
+
+                        if (wc_ed25519_export_public(&key->ks.ed25519.key,
```
stenslae: Added additonal tests

tests/unit.c
```diff
@@ -6116,6 +6251,77 @@ static int test_IdentifyAsn1Key(void)
             result = -611; goto done;
         }
     }
+
+    /* Private-only DER (no embedded public key, unlike unitTestMlDsaPrivKey
```
stenslae: Added coverage

tests/auth.c
```diff
@@ -1338,6 +1364,180 @@ static void test_pubkey_auth_wrong_key(void)
...
+    mlDer = (byte*)WMALLOC(WC_MLDSA_44_PRV_KEY_DER_SIZE, NULL, 0);
+    AssertNotNull(mlDer);
+    mlDerSz = wc_MlDsaKey_PrivateKeyToDer(&mlKey, mlDer,
+            WC_MLDSA_44_PRV_KEY_DER_SIZE);
+    wc_MlDsaKey_Free(&mlKey);
+    AssertIntGT(mlDerSz, 0);
+
+    ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL);
+    AssertNotNull(ctx);
+    AssertIntLT(wolfSSH_CTX_UsePrivateKey_buffer(ctx, mlDer, (word32)mlDerSz,
```
stenslae: Fixed assertions

src/internal.c
```diff
@@ -1857,14 +1885,36 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
...
+         * as Ed25519. */
+        if (key->keyId == ID_UNKNOWN && noPubKeyRet == 0) {
             idx = 0;
             ret = wc_ed25519_init_ex(&key->ks.ed25519.key, heap, INVALID_DEVID);
 
             if (ret == 0) {
                 if (isPrivate) {
                     ret = wc_Ed25519PrivateKeyDecode(in, &idx,
                             &key->ks.ed25519.key, inSz);
+#ifdef DEBUG_WOLFSSH
```
stenslae: Fixed the rejection logic

src/internal.c
```diff
@@ -1704,6 +1710,9 @@ int IdentifyAsn1Key(const byte* in, word32 inSz, int isPrivate, void* heap,
     word32 idx;
     int ret;
     int dynType = isPrivate ? DYNTYPE_PRIVKEY : DYNTYPE_PUBKEY;
+    /* Set to WS_CRYPTO_FAILED if ML-DSA key lacks derivable public key.
+     * Prevents Ed25519 fallback decode. */
+    int noPubKeyRet = 0;
 #ifndef WOLFSSH_NO_MLDSA
```
stenslae, line 1716: Added suggested to guard

ejohnstown (approved):

PR #1130, opened 2026-07-27
Fix pty-req terminal mode size handling when stdin is not a tty

yosuke-wolfssl:
### Problem
`CreateMode()` returned `-1` when `tcgetattr(STDIN_FILENO)` failed, and
`SendChannelTerminalRequest()` stored that in `word32 modeSz`, making it
`0xFFFFFFFF`. The `PreparePacket()` size arithmetic is `word32`, so it wrapped and
allocated a *small* output buffer; the guarded copy below then ran
`WMEMCPY(output + idx, mode, 0xFFFFFFFF)` — a 4 GiB read out of a 4096-byte stack
buffer into a ~60-byte heap buffer. High severity, CWE-681 leading to CWE-787.

Triggered by any client that requests a pty while stdin is not a terminal:
`wolfssh -u user host < /dev/null`, cron/CI runs, or an embedder with no
controlling terminal. Both in-tree clients set `WOLFSSH_SESSION_TERMINAL`
unconditionally in interactive mode. Closes f-7202.

### Fix (`src/internal.c`)
- **`CreateModeTermios()`** is a new helper holding the termios body; it returns
  `WS_FATAL_ERROR` instead of printing and returning `-1`. **`CreateMode()`** falls
  back to the default mode list (38400 + `TTY_OP_END`) when there is no terminal —
  the path the no-termios build already used, so the `#else` branch is gone.
- **The mode writes are bounded at the source.** `TTYWordSet()`/`TTYSet()` took a
  bare `byte*` and advanced `idx` blindly, so the caller could only detect damage
  already done. Both now refuse to write past the buffer:

```c
    if (*idx + TERMINAL_MODE_SZ + 1 > TERMINAL_MODES_MAX_SZ)
        return;
```

  The reserved byte leaves room for the terminator, whose write is guarded the same
  way — `CreateMode()` returns `WS_FATAL_ERROR` if it will not fit. A full buffer
  truncates to a well-formed `TTY_OP_END`-terminated list. `TERMINAL_MODES_MAX_SZ`
  replaces the bare `4096` literals so the buffer and its bound cannot drift.
- **`GetTerminalInfo()`** had the same gap for the *window size*: the
  `ioctl(TIOCGWINSZ)` return was ignored, so a client with no terminal negotiated a
  0x0 pty, which wolfsshd passes straight to `TIOCSWINSZ`. The no-terminal defaults
  are now assigned once up front and each platform arm overwrites only what it
  actually read, which also removed the duplicated defaults and the `#else` arm.

Behavior change: a non-tty client completes the `pty-req` with default modes and an
80x24 window and runs to EOF instead of crashing, matching `ssh -tt host </dev/null`.

### Tests (`tests/unit.c`)
`test_SendChannelTerminalRequestNoTty()` points stdin *and* stdout at `/dev/null`
(`TIOCGWINSZ` reads stdout), pins `TERM`, sends the request, restores the
environment, then asserts every field of the captured packet by value: `"pty-req"`,
the term string, an 80x24 window with zero pixel dimensions, and the exact 11-byte
fallback mode list.

### Verification
- `make check`: 10 passed, 1 skipped, 0 failed. Clean under gcc-13 `-Werror` across
  6 configs. ASan + UBSan clean.
- Negative controls: reverting `src/internal.c` makes the test die under ASan with
  `stack-buffer-overflow ... SendChannelTerminalRequest`; deleting the setter guards
  with the buffer shrunk to 8 bytes gives `stack-buffer-overflow ... in TTYWordSet`;
  restoring the unconditional `ioctl()` yields `window 0x0 (0x0 px)`; changing the
  fallback baud to 9600 fails the mode-content assertion.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -19617,8 +19617,12 @@ int SendChannelRequest(WOLFSSH* ssh, byte* name, word32 nameSz)
 
 #if defined(WOLFSSH_TERM) && !defined(NO_FILESYSTEM)
 
+/* out is always a TERMINAL_MODES_MAX_SZ buffer, and one byte is held back
+ * for the terminator CreateMode() appends */
 static void TTYWordSet(word32 flag, int type, byte* out, word32* idx)
 {
+    if (*idx + TERMINAL_MODE_SZ + 1 > TERMINAL_MODES_MAX_SZ)
+        return;
     out[*idx] = type; *idx += 1;
     c32toa(flag, out + *idx); *idx += UINT32_SZ;
 }
@@ -19629,6 +19633,8 @@ static void TTYWordSet(word32 flag, int type, byte* out, word32* idx)
     /* sets terminal mode in buffer and advances idx */
     static void TTYSet(word32 isSet, int type, byte* out, word32* idx)
     {
+        if (*idx + TERMINAL_MODE_SZ + 1 > TERMINAL_MODES_MAX_SZ)
+            return;
         if (isSet) isSet = 1;
         out[*idx] = type; *idx += 1;
         c32toa(isSet, out + *idx); *idx += UINT32_SZ;
@@ -19638,25 +19644,17 @@ static void TTYWordSet(word32 flag, int type, byte* out, word32* idx)
     {
         TTYWordSet((flag & 0xFF), type, out, idx);
     }
-#endif /* !USE_WINDOWS_API && !MICROCHIP_PIC32 && !NO_TERMIOS*/
 
-
-/* create terminal mode string for pseudo-terminal request
- * returns size of buffer */
-static int CreateMode(WOLFSSH* ssh, byte* mode)
-{
-    word32 idx = 0;
-
-    #if !defined(USE_WINDOWS_API) && !defined(MICROCHIP_PIC32) && \
-        !defined(NO_TERMIOS)
+    /* fills mode with the terminal settings of stdin
+     * returns size of buffer, or WS_FATAL_ERROR when there is no terminal */
+    static int CreateModeTermios(byte* mode)
     {
         WOLFSSH_TERMIOS term;
         int baud;
+        word32 idx = 0;
 
-        if (tcgetattr(STDIN_FILENO, &term) != 0) {
-            printf("Couldn't get the original terminal settings.\n");
-            return -1;
-        }
+        if (tcgetattr(STDIN_FILENO, &term) != 0)
+            return WS_FATAL_ERROR;
 
         /* get baud rate */
         baud = (int)cfgetospeed(&term);
@@ -19770,18 +19768,44 @@ static int CreateMode(WOLFSSH* ssh, byte* mode)
         TTYSet((term.c_cflag &  CS8), WOLFSSH_CS8, mode, &idx);
         TTYSet((term.c_cflag &  PARENB), WOLFSSH_PARENB, mode, &idx);
         TTYSet((term.c_cflag &  PARODD), WOLFSSH_PARODD, mode, &idx);
+
+        return (int)idx;
     }
-    #else
-    {
-        /* No termios. Just set the bitrate to 38400. */
+#endif /* !USE_WINDOWS_API && !MICROCHIP_PIC32 && !NO_TERMIOS*/
+
+
+/* create terminal mode string for pseudo-terminal request
+ * mode must be a TERMINAL_MODES_MAX_SZ buffer
+ * returns size of buffer, or WS_FATAL_ERROR when it would not fit */
+static int CreateMode(WOLFSSH* ssh, byte* mode)
+{
+    word32 idx = 0;
+    int termRet = WS_FATAL_ERROR;
+    int ret = WS_FATAL_ERROR;
+
+    #if !defined(USE_WINDOWS_API) && !defined(MICROCHIP_PIC32) && \
+        !defined(NO_TERMIOS)
+        termRet = CreateModeTermios(mode);
+        if (termRet < 0)
+            WLOG(WS_LOG_INFO, "No terminal settings, using default modes");
+    #endif
+
+    if (termRet < 0) {
+        /* No terminal settings. Just set the bitrate to 38400. */
         TTYWordSet(38400, WOLFSSH_TTY_OP_ISPEED, mode, &idx);
         TTYWordSet(38400, WOLFSSH_TTY_OP_OSPEED, mode, &idx);
     }
-    #endif /* !USE_WINDOWS_API && !MICROCHIP_PIC32 && !NO_TERMIOS */
+    else {
+        idx = (word32)termRet;
+    }
+
+    if (idx + 1 <= TERMINAL_MODES_MAX_SZ) {
+        mode[idx++] = WOLFSSH_TTY_OP_END;
+        ret = (int)idx;
+    }
 
     WOLFSSH_UNUSED(ssh);
-    mode[idx++] = WOLFSSH_TTY_OP_END;
-    return idx;
+    return ret;
 }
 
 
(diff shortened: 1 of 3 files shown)


src/internal.c
```diff
@@ -20050,21 +20062,30 @@ int SendChannelTerminalRequest(WOLFSSH* ssh)
...
     const char* term = NULL;
     byte mode[4096];
-    word32 termSz, typeSz, modeSz;
+    word32 termSz = 0, typeSz = 0, modeSz = 0;
     word32 w = 0, h = 0, pxW = 0, pxH = 0;
+    int modeRet;
 
     WLOG(WS_LOG_DEBUG, "Entering SendChannelTerminalRequest()");
 
     if (ssh == NULL)
```
ejohnstown, line 19930: **🟠 [Medium] No-tty pty-req now negotiates a 0x0 terminal window**

The PR turns the no-terminal case from a fatal (crashing) path into a completed `pty-req`. The mode string now falls back to sane defaults (38400 baud), but the *window size* has no matching fallback. `GetTerminalInfo()` calls `ioctl(STDOUT_FILENO, TIOCGWINSZ, &windowSize)` and never checks the return value; `windowSize` is zero-initialized (`struct winsize windowSize = { 0,0,0,0 };`), so when there is no controlling terminal the client sends `widthChar = 0, heightRows = 0, widthPixels = 0, heightPixels = 0`. The server takes those verbatim: `apps/wolfsshd/wolfsshd.c:1646` does `s.ws_col = ssh->widthChar; ... ioctl(childFd, TIOCSWINSZ, &s);`, so the allocated pty really is 0 columns by 0 rows. Note the `#else` arm of the same function already documents 80x24 as the 'sane defaults for terminal size if not yet supported' -- the ioctl arm silently produces 0x0 instead. The 0x0 case was technically reachable before the PR (tty on stdin, pipe on stdout), but the PR is what makes the fully non-interactive client complete the request rather than die, so this is now the normal outcome of the newly supported use case. Since the whole point of the change is 'complete the pty-req with defaults instead of crashing', the size deserves the same treatment as the modes.

**Fix:** Give the window size the same no-terminal fallback the modes just got (80x24), or state explicitly in the commit message that a 0x0 pty is the intended result for non-tty clients. Worth a test assertion on the width/height fields either way, since the new test currently skips over them.

src/internal.c
```diff
@@ -20050,21 +20062,30 @@ int SendChannelTerminalRequest(WOLFSSH* ssh)
...
+    if (ret == WS_SUCCESS) {
+        GetTerminalInfo(&w, &h, &pxW, &pxH, &term);
+        if (term == NULL) {
+            term = "xterm";
+        }
+        termSz = (word32)WSTRLEN(term);
+        typeSz = (word32)WSTRLEN(cType);
+
+        /* a negative or oversized result would wrap the packet size */
+        modeRet = CreateMode(ssh, mode);
```
ejohnstown, line 19942: **🟠 [Medium] CreateMode() bound check happens after the buffer has already been written**

The new guard `if (modeRet < 0 || (word32)modeRet > (word32)sizeof(mode)) ret = WS_FATAL_ERROR;` runs *after* `CreateMode()` has already written into `mode`. `CreateMode()` still takes only `byte* mode` with no size parameter, and neither it nor `CreateModeTermios()` bounds its writes -- `TTYWordSet`/`TTYSet`/`TTYCharSet` advance `idx` blindly and the final `mode[idx++] = WOLFSSH_TTY_OP_END;` is unchecked. Today the worst case is roughly 55 entries * 5 bytes plus the terminator (~280 bytes) against a 4096-byte buffer, so nothing overflows, but the caller's check can only ever detect damage that has already been done. The check also encodes the buffer size in two places (the caller's `sizeof(mode)` and the callee's implicit assumption). Since the PR is specifically hardening this call, passing the size through would make the invariant enforceable rather than merely observable.

**Fix:** Add a `modeMaxSz` parameter to `CreateMode()`/`CreateModeTermios()` and bound the writes inside, keeping the caller's check as a belt-and-braces assertion. If that is considered out of scope for this fix, add a comment at the top of `CreateMode()` stating the maximum number of bytes it can emit so the 4096 buffer size is provably sufficient.

tests/unit.c
```diff
@@ -5632,6 +5644,113 @@ static int test_DoChannelRequest(void)
...
+    modeSz = CaptureUint32(s_chanReqCapture + idx);
+    idx += UINT32_SZ;
+
+    if (modeSz == 0 || modeSz > 4096 || idx + modeSz > s_chanReqCaptureSz) {
+        printf("SendChannelTerminalRequest: modeSz=%u\n", modeSz);
+        result = -812;
+        goto done;
+    }
+    if (s_chanReqCapture[idx + modeSz - 1] != WOLFSSH_TTY_OP_END) {
+        result = -813;
```
ejohnstown: **🟠 [Medium] New test asserts packet shape but never that the default modes were used**

The test's assertions on the encoded modes are purely structural: `modeSz != 0`, `modeSz <= 4096`, the modes fit inside the capture, and the last byte is `WOLFSSH_TTY_OP_END`. Nothing checks that the *fallback* content was produced. The fallback is fully deterministic -- `TTYWordSet(38400, WOLFSSH_TTY_OP_ISPEED)` + `TTYWordSet(38400, WOLFSSH_TTY_OP_OSPEED)` + terminator = exactly 11 bytes -- so it can be asserted exactly. As written, the only thing that would catch the termios path being taken by accident is the 256-byte cap on `s_chanReqCapture` (a real-tty packet is ~330 bytes and would trip the `-812` bound check), which is an incidental property of the capture harness rather than a deliberate assertion. A future change that made `CreateMode()` emit different or partial default content would still pass.

**Fix:** Assert the exact fallback encoding (length 11, ISPEED/OSPEED opcodes, 38400 values) so the test pins the behavior the fix introduces, not just that the packet is well-formed.

ejohnstown (commented):
## Skoll Code Review

**Findings:** 7 total — 3 posted, 4 skipped

### Posted findings
- [Medium] **No-tty pty-req now negotiates a 0x0 terminal window** — `src/internal.c:20071-20087`
- [Medium] **CreateMode() bound check happens after the buffer has already been written** — `src/internal.c:20083-20087`
- [Medium] **New test asserts packet shape but never that the default modes were used** — `tests/unit.c:5735-5745`

### Skipped findings
- [Low] ~~Test packet walk can wrap word32 and read past the 256-byte capture buffer~~
- [Low] ~~Test hardcodes 4096 instead of tracking the library's mode buffer size~~
- [Low] ~~Test failure codes -800..-814 collide with three existing tests~~
- [Low] ~~modeRet failure branch in SendChannelTerminalRequest is now unreachable and untested~~

tests/unit.c
ejohnstown: Concretely, the fallback is 11 deterministic bytes, so the assertion can be exact:

```C
static const byte expectedModes[] = {
    WOLFSSH_TTY_OP_ISPEED, 0x00, 0x00, 0x96, 0x00,
    WOLFSSH_TTY_OP_OSPEED, 0x00, 0x00, 0x96, 0x00,
    WOLFSSH_TTY_OP_END
};

if (modeSz != (word32)sizeof(expectedModes) ||
        idx + modeSz > s_chanReqCaptureSz) { result = -812; goto done; }
if (WMEMCMP(s_chanReqCapture + idx, expectedModes, modeSz) != 0) {
    result = -813; goto done;
}
```

Three regressions the current assertions accept: (1) if CreateModeTermios() ever returned 0 rather than a negative, CreateMode() tests if (termRet < 0), so 0 takes the else, idx stays 0, both TTYWordSet() calls are skipped, and mode becomes just { WOLFSSH_TTY_OP_END } -- modeSz == 1 passes every check while the pty-req carries no modes at all; (2) a wrong opcode or baud (9600 for 38400) changes neither size nor terminator; (3) both paths writing leaves ~290 bytes, still under 4096. Also worth asserting typeSz == 7 and the "pty-req" bytes before trusting the offset walk.

src/internal.c
yosuke-wolfssl: Fixed. `GetTerminalInfo()` now assigns the no-terminal defaults (80x24, zero
pixels) once up front, and each platform arm overwrites only what it actually
read — the `ioctl()` return is checked, and a zero `ws_col`/`ws_row` counts as
"no terminal". That also collapsed the duplicated defaults, so the `#else` arm
is gone and 80/24 became `TERMINAL_WIDTH_DEFAULT` / `TERMINAL_HEIGHT_DEFAULT`.

src/internal.c
yosuke-wolfssl: Fixed at the source. Rather than add a size parameter, the limit is a shared
constant — all three setters are `static` and every call site writes the one
`mode[]` buffer:

```c
    if (*idx + TERMINAL_MODE_SZ + 1 > TERMINAL_MODES_MAX_SZ)
        return;
```

In `TTYWordSet()` and `TTYSet()`, that covers all 58 writes (`TTYCharSet()`
delegates). The reserved byte leaves room for the terminator, whose write is
also guarded — `CreateMode()` returns `WS_FATAL_ERROR` if it won't fit, covering
the path where `termRet` was used directly as a write offset. A full buffer
truncates to a well-formed `TTY_OP_END`-terminated list. `TERMINAL_MODES_MAX_SZ`
replaces the bare `4096` in both the declaration and the test.

Negative control: with the guards deleted and the buffer shrunk to 8 bytes, ASan
reports `stack-buffer-overflow ... in TTYWordSet`; restored at the same size it
runs clean.

tests/unit.c
yosuke-wolfssl: Adopted, including the exact 11-byte `expectedModes[]` compare and the
`typeSz == 7` / `"pty-req"` check. Regression (1) from your follow-up —
`CreateModeTermios()` returning `0` and leaving a bare `{ TTY_OP_END }` — is now
caught by the content compare rather than passing as `modeSz == 1`.

Two adjustments: the window assertion needed stdout redirected too (`TIOCGWINSZ`
reads `STDOUT_FILENO`), and bounds are written `len > captureSz - idx` instead of
`idx + len > captureSz`, closing the skipped `word32` wrap finding.

yosuke-wolfssl:
Hello @ejohnstown ,
Thank you for reviewing. I fixed the issue you mentioned.
Could you please review it again ?

tests/unit.c
```diff
@@ -5672,6 +5687,218 @@ static int test_DoChannelRequest(void)
...
+        result = -1476;
+        goto done;
+    }
+    if (dup2(devNull, STDIN_FILENO) < 0) { result = -1477; goto done; }
+    if (dup2(devNull, STDOUT_FILENO) < 0) { result = -1478; goto done; }
+
+    ret = SendChannelTerminalRequest(ssh);
+
+    /* put the streams back before anything is printed */
+    if (dup2(stdinCopy, STDIN_FILENO) < 0 ||
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

ejohnstown (approved):

PR #1132, opened 2026-07-28
Match OpenSSH on empty name-list elements

ejohnstown:
CountNameList() and GetNameListRaw() disagreed about leading and trailing commas in a peer name-list, so ,ssh-rsa was counted as one name, parsed as two, and rejected with WS_BUFFER_E.

GetNameListRaw() now stops at the first zero-length name and drops the rest of the list, which is what OpenSSH's match_list() does with a peer proposal. CountNameList() is deleted
— DoExtInfoServerSigAlgs() parses into a capped array instead, so nothing predicts the name count ahead of the parse.

Two adjacent bugs fell out: a doubled comma folded into the preceding name (ssh-rsa,ssh-dss,, silently lost ssh-dss), and an empty name-list returned WS_BAD_ARGUMENT instead of parsing to zero names.

Regression tests cover the KEXINIT and server-sig-algs paths, asserting which names survive. Supersedes #1123.

Issue: F-6817

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -4439,47 +4439,19 @@ int GetStringRef(word32* strSz, const byte** str,
 }
 
 
-static word32 CountNameList(const byte* buf, word32 len)
-{
-    word32 count = 0;
-
-    if (buf != NULL && len > 0) {
-        word32 i;
-
-        count = 1;
-        for (i = 0; i < len; i++) {
-            if (buf[i] == ',') {
-                count++;
-            }
-        }
-        /* remove leading comma */
-        if (count > 0 && buf[0] == ',') {
-            count--;
-        }
-        /* remove trailing comma */
-        if (count > 0 && buf[len-1] == ',') {
-            count--;
-        }
-    }
-
-    return count;
-}
-
-
+/* Name-list to IDs. *idListSz is capacity in, count out, and bounds every
+ * store. One ID per name at most, and names are capped at
+ * WOLFSSH_MAX_NAMELIST_CNT, so an idList that size never trips the bound. */
 static int GetNameListRaw(byte* idList, word32* idListSz,
         const byte* nameList, word32 nameListSz)
 {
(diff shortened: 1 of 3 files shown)


tests/regress.c
```diff
@@ -3539,57 +3565,154 @@ static void TestKexInitNameListCaps(void)
...
+    const char* description;
+} EmptyNameCase;
+
+/* RFC 4251 section 5: a name-list holds zero or more names separated by
+ * commas, and every name has a non-zero length. So an empty element -- what a
+ * leading, doubled, or trailing comma leaves behind -- can't be a name, and
+ * the list ends there. Names past it are dropped, not rejected: that is what
+ * OpenSSH's match_list() does with a peer proposal, and its ciphers_valid()
+ * with a local one. CountNameList() sizes peer lists and GetNameListRaw()
+ * parses them, so both have to stop in the same place. */
```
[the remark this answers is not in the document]
ejohnstown: Updated to remove the function name.

JacobBarthelmeh (approved):

src/internal.c
JacobBarthelmeh, line 8045: The PR removes the only `ret = WS_MATCH_UA_KEY_ID_E;` assignment in the tree. Verified by repo-wide grep: the sole producer is src/internal.c:8022 (the line this PR deletes); after the change only the enum definition (wolfssh/error.h:130, -1089) and the GetErrorString case (src/internal.c:486) remain. The public error code becomes unreachable dead API surface, and any application that switch()es on wolfSSH_get_error() for it silently stops seeing it.

JacobBarthelmeh (changes requested):

src/internal.c
ejohnstown: Defer. I am keeping it for now so as not change the ABI unnecessarily right now. I have this issue marked on my to-do list to revisit.

JacobBarthelmeh (approved):

PR #1136, opened 2026-07-30
Fix memory-safety, zeroization, and error-handling findings

ejohnstown:
- **F-7204** - echoserver's `LoadPubKeyList()` now checks the `wolfSSH_ReadKey_buffer()`, `wc_CertPemToDer()`, and `WMALLOC()` results so a malformed or missing key file skips the entry instead of handing `PwMapNew()` a NULL buffer and an indeterminate length.
- **F-7205** - `DoSshPubKey()` wrote its null terminator at `c[inSz-1]`, clobbering the last byte of the copied key; it now writes at `c[inSz]`.
- **F-7206** - `_CertMan_init()` never stored `heap` on the struct, so every `WMALLOC`/`WFREE` in certman.c ran with a NULL heap; it is now set during init.
- **F-7207** - `portfwd_worker()` dropped the unsent tail after a short `wolfSSH_ChannelSend()`, and mistook a full buffer's zero-length `recv()` for the peer closing; both are fixed.
- **F-7208** - `CheckProfile()` ignored the `wc_GetDateAsCalendarTime()` return and compared an indeterminate `struct tm`; it now zero-initializes, checks the return, and fails the profile on a parse error.
- **F-7209** - `DoPemKey()` hardcoded `isPrivate = 1` when identifying the decoded key; it now passes the caller's flag through.
- **F-7210, F-7211** - the example client tracks public-key allocations with a `userPublicKeyAlloc` flag rather than `pubKeyName` (which leaked the cert buffer for `-J` without `-j`), and frees the CA cert and the TPM public key from the heap they were allocated from.
- **F-7212** - `wolfSSH_MakeRsaKey()`, `MakeEcdsaKey()`, and `MakeEd25519Key()` freed their key structs unconditionally, so a failed init freed an indeterminate stack struct; frees are now gated on an init flag.
- **F-7213** - `GetConfigInt()` ran `WSTRCMP()` on the caller's non-terminated slice, rejecting valid zero values like `LoginGraceTime 0m`; it now compares the NUL-terminated copy that `atol()` was given.
- **F-7214** - `wolfSSH_AGENT_worker()` dereferenced `ssh->agent` without a null check; it now returns `WS_AGENT_NULL_E`.
- **F-7215** - `wolfSSH_SFTP_DoName()` left `ret` holding the read length, so a NAME response with zero entries took the error path; `ret` is cleared before the loop.
- **F-7216** - `AeadIncrementExpIv()` returned early on the first non-zero byte, making the nonce increment data-dependent; the carry now runs the full width with no early exit.
- **F-7219** - `SshResourceFree()` wiped only `kSz` bytes of `ssh->k`, which can under-cover after a rekey; it now zeroes `sizeof(ssh->k)` and resets `kSz`.
- **F-7220** - `CheckPasswordHashUnix()` left the hashed password in `crypt()`'s static buffer; it is wiped once the comparison is done.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -1154,6 +1154,105 @@ static void test_wolfSSH_ReadKey(void)
 }
 
 
+#ifndef WOLFSSH_NO_RSA
+/* SubjectPublicKeyInfo for the same RSA key as serverKeyRsaDer. */
+static const char serverKeyRsaPubPem[] =
+    "-----BEGIN PUBLIC KEY-----\n"
+    "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA2l2tJRR2FVnzQP08uGIw\n"
+    "s23A+ezsi4MenkKcykFq04rhUjTgDRNiftQPrlxNBPGN+sWtd6paBcrv+I2r/4op\n"
+    "CUwEwvUZy+0fsbQp08NsqSPfo6DlCN6tjHH5NIhs7Tvwb6UPrFn/azPxcPuMpLNF\n"
+    "Io2dd3rlKV+EFNmZ6urOLVHz41j6WwIPybUqvLJe08Iwuzyxw+9Y81CUKIvEZUr3\n"
+    "ANmX2WtNjZWhimIGtFARIoO06irn0KggR0//Rq7FE+E4i/hUrzpNL/gf14SQ2JMF\n"
+    "BsJ9kNvjnNDEZVoDrQCsWqLN2j+JWDdTvytGeqyJQStaLuh2517jKYWjY+rmhmB8\n"
+    "LQIDAQAB\n"
+    "-----END PUBLIC KEY-----\n";
+/* The same SubjectPublicKeyInfo in DER, what the PEM above decodes to. */
+static const char serverKeyRsaPubDer[] =
+    "30820122300d06092a864886f70d01010105000382010f003082010a02820101"
+    "00da5dad2514761559f340fd3cb86230b36dc0f9ecec8b831e9e429cca416ad3"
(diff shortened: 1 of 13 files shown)


tests/api.c
```diff
@@ -1290,6 +1389,60 @@ static void test_wolfSSH_ReadKey_noTrailingNewline(void)
...
+            WOLFSSH_FORMAT_SSH, &keyTrim, &keyTrimSz,
+            &keyType, &keyTypeSz, NULL);
+    AssertIntEQ(ret, WS_SUCCESS);
+    AssertNotNull(keyTrim);
+    /* Must match the canonical decode exactly, not be truncated by a byte. */
+    AssertIntEQ(keyTrimSz, keyRefSz);
+    AssertIntEQ(WMEMCMP(keyTrim, keyRef, keyRefSz), 0);
+    AssertStrEQ(keyType, "ssh-rsa");
+
+    WFREE(keyRef, NULL, DYNTYPE_FILE);
```
philljj: Doesn't `wolfSSH_ReadKey_buffer()` allocate them as `DYNTYPE_PRIVKEY`?
```
newKey = (byte*)WMALLOC(newKeySz, heap, DYNTYPE_PRIVKEY);
```

philljj (changes requested):
Looks good, just 1 question.

Also, copilot suppressed this comment, but I can't figure out if actually valid:
> When appBufferUsed == appBufferSz, appFd remains in the read fdset, so select() will return immediately as long as the socket stays readable. Because the recv() is skipped in that state, the loop can busy-spin at 100% CPU while waiting for wolfSSH_ChannelSend() to drain the buffer. Consider clearing appFd from rxFds while the buffer is full and still running the flush logic on select() timeout.

tests/api.c
ejohnstown: Updated the mem type appropriately.

philljj (approved):

PR #1137, opened 2026-08-03
Dispatch DoAsn1Key public keys on key type

yosuke-wolfssl:
### Problem

`DoAsn1Key()`'s public-key branch had exactly two arms: an ML-DSA arm gated on an explicit key-ID test, and an unconditional `else` that assumed RSA. But `IdentifyAsn1Key(in, inSz, 0, ...)` also returns `ID_ECDSA_SHA2_NISTP256/384/521` and `ID_ED25519`, which decode into the `ecc_key` and `ed25519_key` members of the `WS_KeySignature.ks` union.

Those IDs fell into the RSA arm and reached `wc_RsaFlattenPublicKey(&key->ks.rsa.key, ...)`, walking `RsaKey.n`/`RsaKey.e` `mp_int` fields that overlay unrelated members of a different key type. Observed behavior: the call returned `WS_SUCCESS` with `*outType` set to `ecdsa-sha2-nistp256` but a blob whose first string was `ssh-rsa`, built from unrelated memory. The encode/decode roundtrip cannot hold, and the read is out of the caller's control.

Reachable through the public API `wolfSSH_ReadPublicKey_buffer()`. The in-tree caller `wolfSSH_TPM_InitKey()` (`examples/client/common.c`) feeds a TPM public key exported as ASN.1, so a TPM-resident ECC key trips this.

### Fix (`src/ssh.c`)

Hoisted the `*outType`/`*outTypeSz` assignment above the chain, gated the RSA arm on its own ID, and added the missing arms. Unknown types now fail closed with `WS_UNIMPLEMENTED_E` instead of falling through to RSA.

| Key ID | Blob layout |
|---|---|
| `ID_SSH_RSA` | `ssh-rsa`, `e`, `n` (unchanged) |
| `ID_ECDSA_SHA2_NISTP256/384/521` | `ecdsa-sha2-nistpXXX`, `nistpXXX`, `Q` via `wc_ecc_export_x963()` |
| `ID_ED25519` | `ssh-ed25519`, `A` via `wc_ed25519_export_public()` |
| `ID_MLDSA44/65/87` | unchanged |
| anything else | `WS_UNIMPLEMENTED_E` |

The fallback also closes a second instance of the same defect: `IdentifyAsn1Key` assigns `ID_MLDSA44/65/87` without per-level guards, while the ML-DSA arm is per-level guarded, so in a `WOLFSSH_NO_MLDSA44` build such a key previously hit the RSA arm too.

Closes f-7504.

### Tests (`tests/unit.c`)

`test_ReadPublicKeyAsn1()` covers RSA, ECDSA P-256/384/521, Ed25519 and ML-DSA 44/65/87. Each case reads a public key back through `wolfSSH_ReadPublicKey_buffer()` and checks `outType` plus every blob string against the key's own exported value, requiring the blob to be exactly consumed — content, not just framing. The RSA case reproduces the leading zero byte `DoAsn1Key` prepends when the modulus MSB is set.

Mutation-tested rather than assumed: corrupting a modulus byte, forcing `nMsb = 0`, and corrupting the ML-DSA raw key are each caught by the corresponding case. The whole test fails before the fix.

### Verification

- `make check`: 10 pass, 1 skip (`external.test`, needs network), 0 fail, in both a standard build and one against wolfSSL `--enable-mldsa`.
- Linked and run against a lean wolfSSL matching the CI configuration, default and `--enable-smallstack`.
- Preflight sweep clean: lint plus 6 configs under `gcc-13 -Werror`.
- `src/ssh.c` compiles `-Werror` clean across all eight on/off combinations of `WOLFSSH_NO_RSA`/`ECDSA`/`ED25519`, and with ML-DSA compiled in. `tests/unit.c` was checked over the same matrix and is clean for every configuration it supports, including an ML-DSA-only build with the other three key types disabled.

### Reviewer notes

Two points have been raised repeatedly by automated review. The rationale is recorded here so it does not need re-deriving.

**The `WS_UNIMPLEMENTED_E` fallback cannot be reached by any input.** `DoAsn1Key` guards the entire dispatch chain on `ret > 0`. Every `keyId` that `IdentifyAsn1Key` can assign (`ID_SSH_RSA`, the three ECDSA curves, the three ML-DSA levels, `ID_ED25519`) has a matching arm, and anything unrecognised returns a negative error, so the block including the fallback is skipped.

This was checked, not assumed. Feeding an ECC SPKI whose curve OID is an unsupported curve returns `WS_UNIMPLEMENTED_E` with `out == NULL`; changing the fallback to return a different code and rebuilding yields the *same* result, proving the line never executes. Any test written against it would pass identically with the `else` deleted. The only configuration that can execute it is `WOLFSSH_NO_MLDSA44` (or the 65/87 equivalents) with ML-DSA otherwise enabled, reaching it through the unguarded level mapping described below, and CI does not build that combination. The branch is deliberate defence in depth against future drift between the two functions.

**Public-key blobs are allocated with `DYNTYPE_PRIVKEY`.** That reads oddly for a public key, but it matches the callers: `examples/client/common.c` and `apps/wolfssh/common.c` both free `wolfSSH_ReadPublicKey_buffer()` output with `DYNTYPE_PRIVKEY`, as do `DoSshPubKey`, `DoPemKey`, `DoOpenSshKey` and the RSA arm. The tag is advisory rather than functional: without `WOLFSSL_STATIC_MEMORY` wolfSSL discards it before any custom allocator callback sees it, and with static memory the free path selects its bucket by size, consulting the type only for the IO-pool types, which wolfSSH's `500`-based enum can never collide with. Aligning the tag is a consistency fix, not a routing fix.

### Not in this PR

The unguarded `ID_MLDSA44/65/87` assignment in `IdentifyAsn1Key` reports a key type the build has disabled. This change makes it fail closed rather than confuse types, but the ID arguably should not be produced at all. Left for a separate finding.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1907,6 +1907,9 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
     }
 
     if (ret > 0 && !isPrivate) {
+        *outType = (const byte*)IdToName(ret);
+        *outTypeSz = (word32)WSTRLEN((const char*)*outType);
+
 #ifndef WOLFSSH_NO_MLDSA
         if (
 #ifndef WOLFSSH_NO_MLDSA44
@@ -1928,12 +1931,9 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
 #else
                 WC_MLDSA_44_PUB_KEY_SIZE;
 #endif
-            const char* name = IdToName(ret);
-            word32 nameLen = (word32)WSTRLEN(name);
+            const char* name = (const char*)*outType;
+            word32 nameLen = *outTypeSz;
             word32 localIdx = 0;
-
-            *outType = (const byte*)name;
-            *outTypeSz = nameLen;
 
             rawPub = (byte*)WMALLOC(rawPubSz, heap, DYNTYPE_PUBKEY);
             if (rawPub == NULL) {
@@ -1947,7 +1947,7 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
                 }
                 else {
                     *outSz = LENGTH_SZ + nameLen + LENGTH_SZ + rawPubSz;
-                    newKey = (byte*)WMALLOC(*outSz, heap, DYNTYPE_PUBKEY);
+                    newKey = (byte*)WMALLOC(*outSz, heap, DYNTYPE_PRIVKEY);
                     if (newKey == NULL) {
                         ret = WS_MEMORY_E;
                     }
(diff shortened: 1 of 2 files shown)


tests/unit.c
```diff
@@ -9272,6 +9272,358 @@ static int test_IdentifyAsn1Key_MlDsaPrivOnlyDer(byte level,
...
+                CheckPubBlobStr(blob, blobSz, &idx, q, qSz) != 0 ||
+                idx != blobSz) {
+            printf("ReadPublicKeyAsn1: ECC %s blob malformed\n", curveName);
+            result = -6949;
+        }
+    }
+
+    if (blob != NULL) {
+        WFREE(blob, NULL, DYNTYPE_PUBKEY);
+    }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

src/ssh.c
```diff
@@ -1928,13 +1931,10 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
...
+            const char* name = (const char*)*outType;
+            word32 nameLen = *outTypeSz;
             word32 localIdx = 0;
 
-            *outType = (const byte*)name;
-            *outTypeSz = nameLen;
-
             rawPub = (byte*)WMALLOC(rawPubSz, heap, DYNTYPE_PUBKEY);
             if (rawPub == NULL) {
                 ret = WS_MEMORY_E;
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

tests/unit.c
```diff
@@ -9272,6 +9305,351 @@ static int test_IdentifyAsn1Key_MlDsaPrivOnlyDer(byte level,
...
+        }
+    }
+    if (result == 0) {
+        if (blobTypeSz != (word32)WSTRLEN(keyName) ||
+                WMEMCMP(blobType, keyName, blobTypeSz) != 0) {
+            printf("ReadPublicKeyAsn1: RSA expected type %s\n", keyName);
+            result = -6964;
+        }
+    }
+    if (result == 0) {
```
yosuke-wolfssl, line 9678: Fixed

tests/unit.c
```diff
@@ -9272,6 +9305,351 @@ static int test_IdentifyAsn1Key_MlDsaPrivOnlyDer(byte level,
...
+        WFREE(blob, NULL, DYNTYPE_PRIVKEY);
+    }
+
+    return result;
+}
+#endif /* WOLFSSH_NO_RSA */
+
+/* wolfSSH_ReadPublicKey_buffer() with WOLFSSH_FORMAT_ASN1 must dispatch on the
+ * identified key type instead of assuming every DER public key is RSA. */
+static int test_ReadPublicKeyAsn1(void)
```
yosuke-wolfssl, line 9711: Fixed

tests/unit.c
```diff
@@ -9272,6 +9305,469 @@ static int test_IdentifyAsn1Key_MlDsaPrivOnlyDer(byte level,
...
+    }
+    wc_FreeRsaKey(&rsaKey);
+
+    return result;
+}
+#endif /* WOLFSSH_NO_RSA */
+
+/* wolfSSH_ReadPublicKey_buffer() with WOLFSSH_FORMAT_ASN1 must dispatch on the
+ * identified key type instead of assuming every DER public key is RSA. */
+static int test_ReadPublicKeyAsn1(void)
```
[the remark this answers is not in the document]
yosuke-wolfssl: Won't do, see the PR body

src/ssh.c
```diff
@@ -1966,68 +1966,162 @@ static int DoAsn1Key(const byte* in, word32 inSz, byte** out,
...
+#ifndef WOLFSSH_NO_RSA
+        if (ret == ID_SSH_RSA) {
+            long e;
+            byte n[RSA_MAX_SIZE]; /* TODO: Handle small stack */
+            word32 nSz = (word32)sizeof(n), eSz = (word32)sizeof(e);
+            const char* keyFormat = "ssh-rsa";
+            word32 idx = 0;
+            int nMsb = 0;
+
+            ret = wc_RsaFlattenPublicKey(&key->ks.rsa.key, (byte*)&e, &eSz,
```
ejohnstown: The three new arms map crypto failures to WS_CRYPTO_FAILED, but this one assigns the wolfCrypt return straight to ret, so a flatten failure propagates a raw wolfCrypt code out of the public API. That's the same -131 signature that both pre-fix instances of this defect produced (Ed25519 SPKI, and an ML-DSA-44 SPKI in a WOLFSSH_NO_MLDSA44 build). Worth normalizing to WS_CRYPTO_FAILED for consistency with the arms added here.

ejohnstown (changes requested):

src/ssh.c
yosuke-wolfssl: You are correct, I fixed it. The RSA arm now takes the flatten result into a local
`wcRet` and maps a failure to `WS_CRYPTO_FAILED`, so the arms are
consistent and no raw wolfCrypt code escapes the public API. The two
following guards became `ret == WS_SUCCESS` instead of `ret == 0`
(same value) to match.

yosuke-wolfssl:
Hello @ejohnstown ,
Thank you for reviewing. I fixed it.
Could you please check this again ?

ejohnstown (approved):

PR #1142, opened 2026-08-04
tests: fix api.test SFTP read target race

ejohnstown:
- The test opened the listing's first regular file, which raced with tests/testsuite.test creating and removing files in the same directory under make -j check — the macOS -1062 flake. It now stages its own target over SFTP, selects it by name, and removes it afterward.
- Staging retries on WS_WANT_READ/WS_WANT_WRITE/WS_REKEYING and clamps writes to land exactly on WOLFSSH_MAX_SFTP_RW for any value of that macro; the listing assert, read block, and cleanup all key on Open success, so reads never target a stale file.
- Reads now verify the returned payload is the staged fill byte, not just a plausible length.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -2697,6 +2697,71 @@ static void sftp_client_connect(WOLFSSH_CTX** ctx, WOLFSSH** ssh, int port)
 }
 
 
+/* Upper bound on retry iterations for the SFTP helpers below. */
+#define SFTP_MAX_RETRY_TRIES 1000
+
+/* Which errors are safe to retry depends on the callee: re-calling one that
+ * has already torn down its state reissues a request that may be partly on
+ * the wire, which is the desync these loops exist to avoid.
+ *
+ * wolfSSH_SFTP_Open() and wolfSSH_SFTP_SendWritePacket() keep their state
+ * only for these two (src/wolfsftp.c STATE_OPEN_SEND, STATE_OPEN_GETHANDLE,
+ * STATE_SEND_WRITE_READ_STATUS).  wolfSSH_SFTP_LS() also keeps it on
+ * WS_REKEYING. */
+static int sftp_error_keeps_state(int err)
+{
+    return err == WS_WANT_READ || err == WS_WANT_WRITE;
+}
+
+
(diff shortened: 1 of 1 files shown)


tests/api.c
```diff
@@ -2749,18 +2774,77 @@ static void test_wolfSSH_SFTP_SendReadPacket(void)
...
         while (tmp != NULL) {
-            if ((tmp->atrb.sz[0] > 0) &&
-                    (tmp->atrb.flags & WOLFSSH_FILEATRB_PERM) &&
-                    !(tmp->atrb.per & 040000)) {
+            if (tmp->fName != NULL && WSTRCMP(tmp->fName, rdName) == 0) {
                 break;
             }
             tmp = tmp->next;
         }
+        if (rdCreated) {
```
JacobBarthelmeh: Having the compare of the sftp results is good. Am I reading this right though that these if{} statements are silently ignored in the case the creating the test file failed? If so it seems like it could be opening up a blind spot with testing coverage.

tests/api.c
```diff
@@ -2749,18 +2774,77 @@ static void test_wolfSSH_SFTP_SendReadPacket(void)
...
+                rdCreated = 1;
+                break;
+            }
+            rdErr = wolfSSH_get_error(ssh);
+            if (rdErr != WS_WANT_READ && rdErr != WS_WANT_WRITE &&
+                    rdErr != WS_REKEYING) {
+                break; /* create denied */
+            }
+        }
+        if (rdCreated) {
```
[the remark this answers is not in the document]
ejohnstown: Fixed. The loop now records the Open return and the test asserts it, so timing out on WANT/REKEYING fails rather than skipping. Same for the LS loop.

tests/api.c
ejohnstown: Good catch, that was a blind spot. Dropped the rdCreated guard entirely: the staging Open, the LS, and the listing match are now asserted, so the read block always runs. Also dropped the atrb.sz >= 18 guard on the first read.

JacobBarthelmeh (approved):

PR #1151, opened 2026-08-07
Report a failed PEM decode as WS_PARSE_E, not WS_BAD_FILE_E

yosuke-wolfssl:
### Problem

`WS_BAD_FILE_E` (-1019) meant two unrelated things: "I could not read the file"
(`ReadFileIntoBuffer()`) and "the PEM body was malformed" (`DoPemCert()` in
`src/ssh.c`, `wolfSSH_ProcessBuffer()` in `src/internal.c`).

So `wolfSSH_ReadCert_buffer()` — a public buffer entry point that opens no file
— answered -1019 for a PEM whose base64 body will not decode. A caller could
not tell "the bytes you gave me are not a certificate" from "your file would
not read" without knowing which form it handed in. `WS_PARSE_E` (-1005) is
already the answer for malformed input everywhere else in this reader,
including the DER arm two functions away.

Inherited style rather than a new mistake: `wolfSSH_ProcessBuffer()` has done
this for years and `DoPemCert()` matched its neighbour when PR #1140 added it.

### Fix (`src/ssh.c`)

"The PEM body would not decode" now answers `WS_PARSE_E` from every entry
point, for certificate, CA and private-key buffers alike. `WS_BAD_FILE_E` is
left meaning only file I/O.

| Site | Was | Is |
|---|---|---|
| `DoPemCert()`, `src/ssh.c` | `WS_BAD_FILE_E` | `WS_PARSE_E` |
| `wolfSSH_ProcessBuffer()` PEM arm, `src/internal.c` | `WS_BAD_FILE_E` | `WS_PARSE_E` |

Measured against the built library, before → after:

```
ReadCert_buffer  [bad body]     -1019 -> -1005   ReadCert_file    [NULL name]    -1019 -> -1019
ReadCert_buffer  [header only]  -1019 -> -1005   ReadCert_file    [missing file] -1019 -> -1019
UseCert_buffer   [bad body]     -1019 -> -1005   UseCert_file     [missing file] -1019 -> -1019
UseCert_buffer   [header only]  -1019 -> -1005   AddRootCert_file [missing file] -1019 -> -1019
AddRootCert_buf  [bad body]     -1019 -> -1005
UsePrivateKey_bf [bad body]     -1019 -> -1005
```

`wolfssh/ssh.h` gained a note naming the split. No caller in the tree
discriminates on these codes; the only `== WS_BAD_FILE_E` comparison is in
`src/wolfsftp.c` and never reaches these paths.

### Tests (`tests/api.c`)

This path had **no coverage at all**, so all six assertions are new pins rather
than moved ones — added to `test_wolfSSH_ReadCert_buffer`,
`test_wolfSSH_CTX_UseCert_buffer`, `test_wolfSSH_CTX_AddRootCert_file` (the
`BUFTYPE_CA` case, as no `_buffer` test exists) and
`test_wolfSSH_CTX_UsePrivateKey_buffer_pem`. The existing half-PEM case stays
pinned at `WS_BAD_FILETYPE_E` so sniff-rejection and decode-rejection remain
distinguishable.

### Verification

- `make check`: 11 passed, 0 failed, 1 skipped (network-dependent `external.test`).
- Negative control: the new assertions fail at -1019 before the fix.
- gcc-13 `-Werror` clean across 6 configs, certs-on and certs-off.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -2607,14 +2607,14 @@ int wolfSSH_ProcessBuffer(WOLFSSH_CTX* ctx,
             return WS_UNIMPLEMENTED_E;
         }
     #endif /* WOLFSSH_CERTS */
-        if (ret < 0) {
+        if (ret <= 0) {
             if (type == BUFTYPE_PRIVKEY) {
                 /* wc_KeyPemToDer may have written partial key material;
                  * zeroize before free on the private-key path. */
                 WS_FORCEZERO(der, inSz);
             }
             WFREE(der, heap, dynamicType);
-            return WS_BAD_FILE_E;
+            return WS_PARSE_E;
         }
         derSz = (word32)ret;
     }
--- a/tests/api.c
+++ b/tests/api.c
@@ -670,6 +670,43 @@ static int load_file(const char* filename, byte** buf, word32* bufSz)
 #endif
 
 
+#ifdef WOLFSSH_CERTS
+
+/* PEM shapes that carry a header the sniff accepts but a body no decoder
+ * will take, so the failure lands in the decoder rather than the sniff. */
+static const char badPemCert[] =
+    "-----BEGIN CERTIFICATE-----\n"
+    "!!!! this is not base64 !!!!\n"
+    "-----END CERTIFICATE-----\n";
+static const char noBodyPemCert[] = "-----BEGIN CERTIFICATE-----\n";
+/* Under one full base64 group, so the body decodes to nothing rather than
+ * failing, and wolfSSL answers 0 for it instead of a negative code. */
+static const char zeroLenPemCert[] =
+    "-----BEGIN CERTIFICATE-----\n"
+    "MI\n"
+    "-----END CERTIFICATE-----\n";
+
+#endif /* WOLFSSH_CERTS */
+
+
+#ifndef WOLFSSH_NO_SERVER
+
+/* The same shapes for a private key, which decodes without certificate
+ * support and so is pinned outside WOLFSSH_CERTS. */
+static const char badPemKey[] =
+    "-----BEGIN PRIVATE KEY-----\n"
+    "!!!! this is not base64 !!!!\n"
+    "-----END PRIVATE KEY-----\n";
+/* This one goes through wc_KeyPemToDer, which answers 0 rather than a
+ * negative code, so it needs its own fixture. */
+static const char zeroLenPemKey[] =
+    "-----BEGIN PRIVATE KEY-----\n"
+    "MI\n"
+    "-----END PRIVATE KEY-----\n";
+
+#endif /* WOLFSSH_NO_SERVER */
+
+
 static void test_wolfSSH_CTX_UseCert_buffer(void)
 {
 #ifdef WOLFSSH_CERTS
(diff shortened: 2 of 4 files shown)


src/internal.c
```diff
@@ -2614,7 +2614,7 @@ int wolfSSH_ProcessBuffer(WOLFSSH_CTX* ctx,
                 WS_FORCEZERO(der, inSz);
             }
             WFREE(der, heap, dynamicType);
-            return WS_BAD_FILE_E;
+            return WS_PARSE_E;
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

tests/api.c
```diff
@@ -834,6 +872,24 @@ static void test_wolfSSH_ReadCert_buffer(void)
     free(cert);
     cert = NULL;
 
+    /* Keeping the header sends these past the sniff and into the decoder,
+     * where a body that will not decode is a parse failure. No file was
+     * opened on this path, so it must not be reported as a file error. */
+    AssertIntEQ(WS_PARSE_E, wolfSSH_ReadCert_buffer((const byte*)badPemCert,
+                (word32)WSTRLEN(badPemCert),
+                &out, &outSz, &outType, &outTypeSz, &flavor, NULL));
+    AssertNull(out);
```
yosuke-wolfssl, line 898: Fixed

wolfssh/ssh.h
```diff
@@ -106,7 +106,10 @@ WOLFSSH_API int wolfSSH_ReadKey_file(const char* name,
 
 #if defined(WOLFSSH_CERTS) || defined(WOLFSSH_OSSH_CERTS)
 /* Decodes a PEM/DER X.509 cert or OpenSSH cert line, detected from content.
- * Caller frees out via heap; on failure every out param is cleared. */
+ * Caller frees out via heap; on failure every out param is cleared. Returns
+ * WS_BAD_FILETYPE_E if the content is not a certificate form, and WS_PARSE_E
+ * if it is but will not decode. WS_BAD_FILE_E comes only from the _file entry
```
yosuke-wolfssl: Fixed

tests/api.c
```diff
@@ -1118,6 +1183,12 @@ static void test_wolfSSH_CTX_UsePrivateKey_buffer_pem(void)
     key = NULL;
 #endif /* WOLFSSH_NO_ECDSA */
 
+    /* A key PEM shares the certificate PEM's decoder, and answers the same
+     * way when its body will not decode. */
+    AssertIntEQ(WS_PARSE_E,
```
yosuke-wolfssl, line 1217: Fixed

wolfssh/ssh.h
yosuke-wolfssl: Fixed

tests/api.c
```diff
@@ -1118,6 +1203,15 @@ static void test_wolfSSH_CTX_UsePrivateKey_buffer_pem(void)
     key = NULL;
 #endif /* WOLFSSH_NO_ECDSA */
 
+    /* A key PEM answers the same way as a certificate one for a body that
+     * will not decode, and for a body that decodes to nothing. */
+    AssertIntEQ(WS_PARSE_E,
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed.

Confirmed, and the guard turned out to be stale rather than deliberate: it dates to `0bb1d1b5b` (2026-04-06), when PEM private-key decoding genuinely was behind `WOLFSSH_CERTS`. `722678df6` (2026-07-22) made the `BUFTYPE_PRIVKEY` arm certs-independent so a PKCS#8 ML-DSA host key would load without certificate support, but left the test guard behind — so the certs-off build, the one that fix exists to serve, had no coverage.

Changes:

- `test_wolfSSH_CTX_UsePrivateKey_buffer_pem` relaxed to `#if !defined(WOLFSSH_NO_SERVER)`.
- `badPemKey` / `zeroLenPemKey` moved out of the `WOLFSSH_CERTS` block into their own `#ifndef WOLFSSH_NO_SERVER` one.
- The RSA and ECDSA success cases inside the test now carry `defined(WOLFSSH_CERTS)` themselves, since they read their keys through `load_file()`, which certificate support carries.

yosuke-wolfssl:
Hi @ejohnstown ,
This is one of the follow-up. Please check this when you have a chance

ejohnstown (approved):

PR #1152, opened 2026-08-10
Replace the matching certificate slot instead of appending a duplicate

yosuke-wolfssl:
### Problem
`SetHostCertificate()` never replaced an existing certificate. Its search loop
records a matching slot in `certIdx`, but `destIdx` is the loop counter and the
loop has no `break`, so on exit it always equals `ctx->privateKeyCount` — the
append slot. `pvtKey` is taken from there, whose `publicKeyFmt` is always
`ID_NONE`, making the `pvtKey->publicKeyFmt == certId` replace branch dead code;
the `else` branch then discards the search result with `certIdx = destIdx;`.

Reloading a certificate for an algorithm that already has one therefore appends
a duplicate slot instead of replacing it. KEX selects the signing slot with a
first-match-and-break scan, so the **stale certificate at the lower index keeps
being served** — a renewed certificate silently has no effect until restart. The
call still returns `WS_SUCCESS`. Additional consequences: the old DER is never
freed, the same `x509v3-*` name is emitted twice in
`server_host_key_algorithms`, and each reload burns a slot until
`WOLFSSH_MAX_PVT_KEYS` is exhausted.

Sibling `SetHostPrivateKey()` gets this right — its `while` loop puts the match
test in the loop condition, so it stops on the matching slot.

### Fix (`src/internal.c`)
Select the destination slot from the search result before using it:

```c
/* Replace the matching slot if the search found one, else append. */
destIdx = HINTISSET(certIdx) ? certIdx : ctx->privateKeyCount;
```

`HINTISSET()` is the file's existing sentinel idiom, and `certIdx` is only ever
assigned an in-range index, so the replace branch becomes reachable and frees
the old DER in place. The append path is unchanged.

Also frees `der` on the `destIdx >= WOLFSSH_MAX_PVT_KEYS` path — ownership
transfers from `wolfSSH_ProcessBuffer()`, which does not free on error. That
matches what `SetHostPrivateKey()` already does, and the fix makes the path
genuinely reachable.

Closes F-8810.

### Tests (`tests/api.c`)
Extended `test_wolfSSH_CTX_UseCert_buffer()`, which already loaded the same
certificate as PEM then DER but never checked the bookkeeping. Now asserts
`privateKeyCount` is unchanged, the slot holds a different pointer, and
`publicKeyAlgoCount == 1`. Added a table-full case asserting
`WS_CTX_KEY_COUNT_E` to cover the free-on-error path.

### Verification
- `make check`: 11 passed, 1 skipped, 0 failed (from a clean rebuild).
- Negative control: reverting the one-line fix fails the new assertion with
  `1 != 2`; injecting a second `WFREE` on the error path trips ASan at
  `SetHostCertificate internal.c:2386`, pinning the test to that branch.
- Clean under `gcc-13 -Werror` across 6 configs, including one with X.509 certs
  disabled.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -2547,30 +2547,35 @@ static int SetHostCertificate(WOLFSSH_CTX* ctx,
         }
     }
 
+    /* Replace the matching slot if the search found one, else append. */
+    destIdx = HINTISSET(certIdx) ? certIdx : ctx->privateKeyCount;
+
     if (destIdx >= WOLFSSH_MAX_PVT_KEYS) {
+        WFREE(der, ctx->heap, dynamicType);
         ret = WS_CTX_KEY_COUNT_E;
     }
     else {
         WOLFSSH_PVT_KEY* pvtKey = ctx->privateKey + destIdx;
 
-        if (pvtKey->publicKeyFmt == certId) {
-            if (pvtKey->cert != NULL) {
-                WFREE(pvtKey->cert, ctx->heap, dynamicType);
-            }
+        /* Copy the paired key into the slot before claiming it, so a
+         * failure here leaves the table unchanged. */
+        ret = UpdateHostCertificates(ctx, keyIdx, destIdx);
+        if (ret != WS_SUCCESS) {
+            WFREE(der, ctx->heap, dynamicType);
         }
         else {
-            certIdx = destIdx;
-            ctx->privateKeyCount++;
-            pvtKey->publicKeyFmt = certId;
-        }
+            if (pvtKey->publicKeyFmt == certId) {
+                if (pvtKey->cert != NULL) {
+                    WFREE(pvtKey->cert, ctx->heap, dynamicType);
+                }
+            }
+            else {
+                ctx->privateKeyCount++;
+                pvtKey->publicKeyFmt = certId;
+            }
 
-        pvtKey->cert = der;
-        pvtKey->certSz = derSz;
-
-        if (ret == WS_SUCCESS) {
-            ret = UpdateHostCertificates(ctx, keyIdx, certIdx);
-        }
-        if (ret == WS_SUCCESS) {
+            pvtKey->cert = der;
+            pvtKey->certSz = derSz;
             RefreshPublicKeyAlgo(ctx);
         }
     }
--- a/tests/api.c
+++ b/tests/api.c
@@ -714,6 +714,12 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
     WOLFSSH_CTX* ctx = NULL;
     byte* cert = NULL;
     word32 certSz = 0;
+#ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
+    byte* key = NULL;
+    word32 keySz = 0;
+    word32 count = 0;
+    byte lastFmt = ID_NONE;
+#endif
 
     ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL);
     AssertNotNull(ctx);
@@ -732,6 +738,8 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
 #ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
     AssertIntEQ(WS_SUCCESS,
             wolfSSH_CTX_UseCert_buffer(ctx, cert, certSz, WOLFSSH_FORMAT_PEM));
+    AssertIntEQ(1, ctx->privateKeyCount);
+    AssertNotNull(ctx->privateKey[0].cert);
 #endif
 
     AssertIntEQ(WS_BAD_FILETYPE_E,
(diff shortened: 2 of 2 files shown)


src/internal.c
```diff
@@ -2377,7 +2377,11 @@ static int SetHostCertificate(WOLFSSH_CTX* ctx,
         }
     }
 
+    /* Replace the matching slot if the search found one, else append. */
+    destIdx = HINTISSET(certIdx) ? certIdx : ctx->privateKeyCount;
+
```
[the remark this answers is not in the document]
yosuke-wolfssl: The mechanics here are right — the loop has no `break`, so `certIdx` holds the
last match while KEX takes the first. But the precondition can't hold once this
change is in.

Duplicate slots are unreachable. Three writers ever claim a slot, and all three
are find-or-append:

- `SetHostCertificate()` — with this change, a pre-existing match sets `certIdx`
  and takes the replace path, so it can no longer append a second slot with a
  `publicKeyFmt` that is already present.
- `SetHostPrivateKey()` — its `while` loop puts the match test in the loop
  condition, so it stops on the match.
- `wolfSSH_SetHostTpmKey()` — same shape.

The two id spaces are disjoint: `IdentifyCert()` returns only `x509v3-*` ids
(and rejects a key type with no x509v3 name rather than falling back to the
plain name), while `IdentifyAsn1Key()` returns only plain key ids.
`UpdateHostCertificates()` writes `key`/`keySz`/`isTpm` but never
`publicKeyFmt`. A CTX starts at `privateKeyCount == 0` with every slot
`ID_NONE`, and there is no API that constructs or deserializes one with
pre-populated slots.

So the duplicate state described was reachable only through the bug this PR
removes. Afterwards there is at most one matching slot, and first-match and
last-match are the same slot. The same reasoning covers `keyIdx`.

Recording only the first match would be unreachable defensive code. A `break`
specifically is worse than a no-op: it would abandon the `keyIdx` half of the
scan whenever the cert slot sits at a lower index than its key slot, which is
the "load a certificate then load a private key" ordering — so
`UpdateHostCertificates()` would get no key hint. That tradeoff is covered in
the PR description.

tests/api.c
```diff
@@ -714,6 +723,22 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
 #ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
     AssertIntEQ(WS_SUCCESS,
             wolfSSH_CTX_UseCert_buffer(ctx, cert, certSz, WOLFSSH_FORMAT_ASN1));
+    /* Reloading a certificate replaces the slot rather than appending a
+     * duplicate, so the count and the advertised algorithm list hold. */
+    AssertIntEQ(count, ctx->privateKeyCount);
+    AssertPtrNE(lastCert, ctx->privateKey[0].cert);
```

tests/api.c
```diff
@@ -714,6 +723,22 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
 #ifndef WOLFSSH_NO_ECDSA_SHA2_NISTP256
     AssertIntEQ(WS_SUCCESS,
             wolfSSH_CTX_UseCert_buffer(ctx, cert, certSz, WOLFSSH_FORMAT_ASN1));
```
yosuke-wolfssl, line 784: Fixed

tests/api.c

tests/api.c
```diff
@@ -707,17 +715,54 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
...
+    AssertIntEQ(2, ctx->publicKeyAlgoCount);
+    /* The replaced slot keeps a fresh copy of the matching private key. */
+    AssertIntEQ(ctx->privateKey[1].keySz, ctx->privateKey[0].keySz);
+    AssertIntEQ(0, XMEMCMP(ctx->privateKey[0].key, ctx->privateKey[1].key,
+                ctx->privateKey[0].keySz));
+
+    /* Every slot taken and none matching: the load is rejected and the
+     * DER released rather than leaked. */
+    lastFmt = ctx->privateKey[0].publicKeyFmt;
+    ctx->privateKey[0].publicKeyFmt = ID_NONE;
```
yosuke-wolfssl, line 810: Fixed

tests/api.c
```diff
@@ -707,17 +715,63 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
...
+    AssertIntEQ(0, XMEMCMP(ctx->privateKey[0].key, ctx->privateKey[1].key,
+                ctx->privateKey[0].keySz));
+
+    /* Every slot taken but one matching: the slot is replaced in place
+     * rather than rejected for want of an append slot. */
+    ctx->privateKeyCount = WOLFSSH_MAX_PVT_KEYS;
+    AssertIntEQ(WS_SUCCESS,
+            wolfSSH_CTX_UseCert_buffer(ctx, cert, certSz, WOLFSSH_FORMAT_ASN1));
+    AssertIntEQ(WOLFSSH_MAX_PVT_KEYS, ctx->privateKeyCount);
+    AssertIntEQ(certSz, ctx->privateKey[0].certSz);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Agreed — both assertions held trivially, so the case only tested the return
code.

Took the intent but not the mechanism: saving `ctx->privateKey[0].cert` and
`AssertPtrNE` against it would reintroduce the dangling-pointer read flagged
earlier at `tests/api.c:729`, since the replace branch frees that pointer
first. The case now reloads a third certificate (`server-cert.der`, 798 B)
over the slot holding `fred-cert.der` (805 B), so `certSz` and an `XMEMCMP`
compare against something genuinely different.

Confirmed load-bearing: an implementation returning `WS_SUCCESS` without
storing the new DER now fails the case with `798 != 0`.

tests/api.c
```diff
@@ -707,17 +715,63 @@ static void test_wolfSSH_CTX_UseCert_buffer(void)
...
+    ctx->privateKeyCount = count;
```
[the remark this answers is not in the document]
yosuke-wolfssl: Correct — only `privateKeyCount` is restored.

Not adding executable cleanup for it. The fabricated-count cases are last in
the function, `wolfSSH_CTX_free()` follows and walks `privateKeyCount`, and
`publicKeyAlgo` is a fixed array in the struct — so nothing observes the stale
entries and there is nothing to leak. Restoring it would mean an extra
`wolfSSH_CTX_UseCert_buffer()` call that tests nothing.

Annotated at the restore point instead, for the next editor:

```c
/* publicKeyAlgo stays stale from the fabricated count; ctx freed below. */
ctx->privateKeyCount = count;
```

src/internal.c
padelsbach: Can you also fix the case when `UpdateHostCertificates` returns error?  I think we need to decrement `privateKeyCount` and cleanup other parts of the table

padelsbach (changes requested):

src/internal.c
yosuke-wolfssl: Fixed by reordering rather than adding rollback: SetHostCertificate() now calls
UpdateHostCertificates() before committing anything to the slot. That call is the
only fallible step (its sole failure return is the key-copy WMALLOC), and it does
not touch the slot's cert or publicKeyFmt -- so on failure the DER is freed, the
error returns, and count/slot/algorithms are exactly as before.

The append path was the real hazard: the slot was already counted and marked with
the x509v3 algorithm while its key was still NULL. RefreshPublicKeyAlgo() is
skipped on the error path, but any later successful load reruns it and would
advertise an algorithm whose slot has no key. The replace path is covered too --
the old certificate is no longer freed before the fallible call.

Not adding a unit test. The path is unreachable except under allocation failure,
and driving it needs a process-wide allocator hook keyed on byte size; I
prototyped that and it works, but it depends on wolfSSL's internal allocation
sizes and os-check runs against the two newest wolfSSL releases resolved at run
time, so a wolfSSL release could break our CI on its own. There is also no error
branch left to test -- nothing is mutated before the fallible call. The ordering
requirement is stated in the comment above it.

yosuke-wolfssl:
Hi @padelsbach ,
I fixed the issue you mentioned. Could you please review it again ?

padelsbach (approved):

PR #1153, opened 2026-08-10
wolfsshd: base Match blocks on the global config and keep included ones

yosuke-wolfssl:
### Problem
`HandleMatch()` built each `Match` node with `wolfSSHD_ConfigCopy(*conf)` — a copy of the *preceding* Match node rather than the global config. Since `wolfSSHD_GetUserConf()` returns the first matching node and stops, a user selected by a later block silently inherited every setting changed by earlier blocks that did not apply to them:

```
ForceCommand /bin/global
Match User alice
    ForceCommand /bin/alice
Match Group staff          # sets no ForceCommand
    PasswordAuthentication no
```

A `staff` user other than alice resolved to `ForceCommand /bin/alice`. This affected every copied field (`forceCmd`, `chrootDir`, `authKeysFile`, `userCAKeysFile`, `permitRootLogin`, …) and is usually fail-open.

A second defect in the same linkage: `HandleMatch()` linked the new node onto the parse cursor (`(*conf)->next = newConf`). An `Include`d file's Match nodes are appended to the list but do not move the caller's cursor, so a later `Match` in the parent **overwrote** the only pointer to them — silently dropping those blocks and leaking them.

### Fix (`apps/wolfsshd/configuration.c`)
- **`head` pointer** added to `WOLFSSHD_CONFIG`, set in `wolfSSHD_ConfigNew()` and propagated by `wolfSSHD_ConfigCopy()`. `HandleMatch()` now copies `(*conf)->head`, so each block carries the global defaults plus only its own overrides.
- **Tail append**: `HandleMatch()` walks from `head` to the end of the list before linking, instead of assuming the parse cursor is the tail. Included Match blocks survive and are no longer leaked. Match *scope* is deliberately unchanged — a directive after an `Include` still belongs to the global config, which is what OpenSSH does.
- **`wolfSSHD_ConfigNew()`** now stores its `heap` argument; the previous code discarded it, so allocations and frees could use different heaps.
- **`apps/wolfsshd/auth.c`**: the `RequestAuthentication` comment describing the old inheritance is now false — a user whose own `Match` never set `TrustedUserCAKeys` keeps the global value and is no longer rejected for certificate auth. The fail-closed branch for a genuinely different CA file is unchanged.

Closes f_8815.

### Tests
`test_GetUserConfMatchNoInherit` and `test_ConfigIncludeMatchChain` in `apps/wolfsshd/test/test_configuration.c`. The latter also asserts that a directive placed after an `Include` resolves for an unmatched user, pinning the scope behavior. `test_IncludeRecursionBound`'s inline file writing was factored into a shared `WriteConfigFile()` helper.

### Verification
- `make check`: 11 passed, 0 failed; `test_configuration` 55 cases, 255 assertions, 0 failures.
- Negative controls: reverting the head copy, the tail walk, or the include scope each fails its corresponding test.
- Include/`Match` scope cross-checked against OpenSSH 10.2p1 using `sshd -T`.
- gcc-13 `-Werror` clean across 6 configs, plus a non-unix (`-U__unix__ -U__APPLE__`) `-Wall -Wextra` pass; ASan + UBSan clean.

### DIFF
--- a/apps/wolfsshd/configuration.c
+++ b/apps/wolfsshd/configuration.c
@@ -95,6 +95,7 @@ struct WOLFSSHD_CONFIG {
     char* pidFile;
     char* authorizedUPNDomains; /* allowlist of UPN realms for cert auth */
     WOLFSSHD_CONFIG* next; /* next config in list */
+    WOLFSSHD_CONFIG* head; /* global config the Match nodes branch from */
     long  loginTimer;
     word16 port;
     byte usePrivilegeSeparation:2;
@@ -107,8 +108,8 @@ struct WOLFSSHD_CONFIG {
 };
 
 /* Maximum depth of nested Include directives. Bounds the recursion
- * through wolfSSHD_ConfigLoad -> ParseConfigLine -> HandleConfigOption
- * -> HandleInclude -> wolfSSHD_ConfigLoad. */
+ * through ConfigLoad -> ParseConfigLine -> HandleConfigOption
+ * -> HandleInclude -> ConfigLoad. */
 #ifndef WOLFSSHD_MAX_INCLUDE_DEPTH
 #define WOLFSSHD_MAX_INCLUDE_DEPTH 16
 #endif
@@ -238,6 +239,8 @@ WOLFSSHD_CONFIG* wolfSSHD_ConfigNew(void* heap)
         WMEMSET(ret, 0, sizeof(WOLFSSHD_CONFIG));
 
         /* default values */
+        ret->heap = heap;
+        ret->head = ret;
         ret->port = 22;
         ret->passwordAuth = 1;
         ret->pubKeyAuth = 1;
@@ -348,6 +351,7 @@ static WOLFSSHD_CONFIG* wolfSSHD_ConfigCopy(WOLFSSHD_CONFIG* conf)
             newConf->permitEmptyPasswords   = conf->permitEmptyPasswords;
             newConf->authKeysFileSet        = conf->authKeysFileSet;
             newConf->strictModes            = conf->strictModes;
+            newConf->head                   = conf->head;
         }
         else {
             wolfSSHD_ConfigFree(newConf);
@@ -729,7 +733,7 @@ static int HandleInclude(WOLFSSHD_CONFIG *conf, const char *value, int depth)
     int ret = WS_SUCCESS;
 
     /* No value, nothing to do */
-    if (!value || value[0] == '\0') {
+    if (conf == NULL || value == NULL || value[0] == '\0') {
         ret = WS_BAD_ARGUMENT;
     }
 
@@ -1142,6 +1146,7 @@ static int CheckMatchSelectors(const char* value)
 static int HandleMatch(WOLFSSHD_CONFIG** conf, const char* value, int valueSz)
 {
     WOLFSSHD_CONFIG* newConf = NULL;
+    WOLFSSHD_CONFIG* tail;
     int ret = WS_SUCCESS;
 
     if (conf == NULL || *conf == NULL || value == NULL) {
@@ -1161,9 +1166,8 @@ static int HandleMatch(WOLFSSHD_CONFIG** conf, const char* value, int valueSz)
         }
     }
 
-    /* create new configure for altered options specific to the match */
     if (ret == WS_SUCCESS) {
-        newConf = wolfSSHD_ConfigCopy(*conf);
+        newConf = wolfSSHD_ConfigCopy((*conf)->head);
         if (newConf == NULL) {
             ret = WS_MEMORY_E;
         }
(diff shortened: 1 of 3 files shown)


apps/wolfsshd/configuration.c
```diff
@@ -238,6 +239,7 @@ WOLFSSHD_CONFIG* wolfSSHD_ConfigNew(void* heap)
         WMEMSET(ret, 0, sizeof(WOLFSSHD_CONFIG));
 
         /* default values */
+        ret->head = ret;
         ret->port = 22;
         ret->passwordAuth = 1;
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

apps/wolfsshd/configuration.c
padelsbach, line 110: I think this comment needs updating

padelsbach: AI tells me this block is no longer required

apps/wolfsshd/test/test_configuration.c
```diff
@@ -3955,7 +4159,12 @@ static int test_ConfigParseAuthorizedUPNDomains(void)
                 ret = WS_FATAL_ERROR;
             }
         }
-        Log(ret == WS_SUCCESS ? " PASSED.\n" : " FAILED.\n");
+        if (ret == WS_SUCCESS) {
```
padelsbach: Is this a related change? Is it needed?

apps/wolfsshd/configuration.c
```diff
@@ -718,22 +722,25 @@ static int HandlePort(WOLFSSHD_CONFIG* conf, const char* value)
...
     int ret = WS_SUCCESS;
 
     /* No value, nothing to do */
-    if (!value || value[0] == '\0') {
+    if (conf == NULL || *conf == NULL || value == NULL || value[0] == '\0') {
         ret = WS_BAD_ARGUMENT;
     }
 
     if (ret == WS_SUCCESS) {
+        heap = (*conf)->heap;
```
yosuke-wolfssl: Fixed

apps/wolfsshd/configuration.c
```diff
@@ -348,6 +351,7 @@ static WOLFSSHD_CONFIG* wolfSSHD_ConfigCopy(WOLFSSHD_CONFIG* conf)
             newConf->permitEmptyPasswords   = conf->permitEmptyPasswords;
             newConf->authKeysFileSet        = conf->authKeysFileSet;
             newConf->strictModes            = conf->strictModes;
+            newConf->head                   = conf->head;
         }
         else {
```
yosuke-wolfssl, line 356: Fixed

apps/wolfsshd/configuration.c
yosuke-wolfssl: Thanks, updated.

yosuke-wolfssl: The AI is right — it was dead code. On failure newConf is already freed and set to NULL by the cleanup block just above, so the else branch could never free anything. Removed.

apps/wolfsshd/test/test_configuration.c
yosuke-wolfssl: Not related — that was an unintended drive-by style change to a test this PR doesn't otherwise touch. Reverted to the original one-line Log(...).

apps/wolfsshd/configuration.c
```diff
@@ -1161,9 +1166,8 @@ static int HandleMatch(WOLFSSHD_CONFIG** conf, const char* value, int valueSz)
         }
     }
 
-    /* create new configure for altered options specific to the match */
     if (ret == WS_SUCCESS) {
-        newConf = wolfSSHD_ConfigCopy(*conf);
+        newConf = wolfSSHD_ConfigCopy((*conf)->head);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Confirmed, but pre-existing: master's HandleMatch also snapshots at parse time (it copies the parse cursor), so later global directives never reached Match nodes there either. This PR changes which base is copied, not the timing. The real fix — buffering Match bodies and replaying them onto the fully-parsed global config, as OpenSSH does — is planned as a follow-up PR; this one stays scoped to the inheritance and include-chain fixes.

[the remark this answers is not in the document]
yosuke-wolfssl: Same root cause as the sibling finding on this line — pre-existing parse-time snapshot semantics. Covered by the planned follow-up PR (see other thread); its regression test targets exactly this layout: Include containing a Match block, PasswordAuthentication no after it.

yosuke-wolfssl:
Hello @padelsbach ,
Thank you for reviewing. I fixed the issues you mentioned.
Could you review this again ?

padelsbach (approved):

padelsbach (approved):

PR #1155, opened 2026-08-10
Fix build with --disable-server

padelsbach:
Build was broken with unused symbols when configured with `--disable-server`.


padelsbach:
jenkins retest this please

ejohnstown (approved):

PR #1170, opened 2026-08-14
examples: use the selected algorithm for the default public key

yosuke-wolfssl:
### Problem

`examples/sftpclient/sftpclient.c` passed a literal `0` for `ClientUsePubKey()`'s `userEcc` argument while passing the real `userEcc` to `ClientSetPrivateKey()`. With no `-i` given, that argument selects between the two compiled-in default keys. Under `WOLFSSH_NO_RSA` the RSA branch is compiled out entirely, so `ret` stays at its initial `0` and the call reports success having loaded nothing — the client then offers an ECC private key with no matching public key and public-key auth fails.

Two adjacent instances of the same class: `examples/scpclient/scpclient.c` hardcoded `0` on both calls with no `WOLFSSH_NO_RSA` fallback, and `examples/client/client.c` had no fallback either.

| Client, `WOLFSSH_NO_RSA` | Before | After |
|---|---|---|
| `wolfsftp` | public key not loaded | ECC default |
| `wolfscp` | neither key loaded | ECC default |
| `client` | private key not loaded | ECC default |

### Fix (`examples/client/common.c`)

- **`ClientSetPrivateKey()` / `ClientUsePubKey()`** — all four built-in key selections gain an `#else` that names the missing algorithm on stderr and returns `WS_NOT_COMPILED`, so a compiled-out default fails at startup instead of silently. Only the `privKeyName`/`pubKeyName == NULL` path reaches these arms, so `-i` and `-j` are unaffected.
- **`sftpclient.c`** passes `userEcc` instead of `0`.
- **`scpclient.c`** gains a `userEcc` and passes it to both calls.
- **`client.c`** and **`scpclient.c`** default `userEcc = 1` under `WOLFSSH_NO_RSA`, as sftpclient already did. Required, not cosmetic: without it the new `WS_NOT_COMPILED` converts their silent no-key-loaded into `err_sys()` at startup.

Closes f-8829.

### Tests

`scripts/sftp.test` and `scripts/scp.test` authenticate with `-u jill -P upthehill`, so the built-in public-key path has no coverage today. Verified with an out-of-tree harness driving each client at `echoserver` with `-u hansel` and no `-i`: both clients fail before the fix under `WOLFSSH_NO_RSA`, pass after, and pass throughout with RSA enabled. Not added to the repo — CI runs no `WOLFSSH_NO_RSA` job for it to execute under. Happy to add both if a job is wanted.

### Verification

- gcc-13 `-O2 -Werror`, 7 configs (`--enable-all`, Zephyr defines, sftp-only, scp-only, sshd, smallstack, `-DWOLFSSH_NO_RSA`): clean.
- `make check` default config: 11 passed, 1 skipped, 0 failed.
- `make check` under `-DWOLFSSH_NO_RSA`: 10 passed, 1 skipped, 1 failed. The failure is `apps/wolfsshd/test/test_configuration`, pre-existing and unrelated — its `BuildAuthKeysLine()` hardcodes `"ssh-rsa"`, and that binary links only `apps/wolfsshd/*.c`.

### DIFF
--- a/examples/client/common.c
+++ b/examples/client/common.c
@@ -1008,12 +1008,23 @@ int ClientSetPrivateKey(const char* privKeyName, int userEcc,
     (void)tpmKeyAuth; /* Not used */
 
     if (privKeyName == NULL) {
+    #if defined(WOLFSSH_NO_RSA) && defined(WOLFSSH_NO_ECC)
+        /* No built-in key to load. Leave the client to authenticate
+         * some other way rather than failing here. */
+        userPrivateKeySz = 0;
+        userPrivateKeyType = NULL;
+        (void)userEcc;
+        (void)heap;
+    #else
         if (userEcc) {
         #ifndef WOLFSSH_NO_ECC
             userPrivateKeySz = sizeof(userPrivateKeyBuf);
             ret = wolfSSH_ReadKey_buffer(hanselPrivateEcc, hanselPrivateEccSz,
                     WOLFSSH_FORMAT_ASN1, &userPrivateKey, &userPrivateKeySz,
                     &userPrivateKeyType, &userPrivateKeyTypeSz, heap);
+        #else
+            fprintf(stderr, "ECC not compiled in, no default private key\n");
+            ret = WS_NOT_COMPILED;
         #endif
         }
         else {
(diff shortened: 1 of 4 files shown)


examples/client/common.c
```diff
@@ -1014,6 +1014,9 @@ int ClientSetPrivateKey(const char* privKeyName, int userEcc,
             ret = wolfSSH_ReadKey_buffer(hanselPrivateEcc, hanselPrivateEccSz,
                     WOLFSSH_FORMAT_ASN1, &userPrivateKey, &userPrivateKeySz,
                     &userPrivateKeyType, &userPrivateKeyTypeSz, heap);
+        #else
+            fprintf(stderr, "ECC not compiled in, no default private key\n");
+            ret = WS_NOT_COMPILED;
```
yosuke-wolfssl, line 1027: Fixed

yosuke-wolfssl, line 1027: Fixed

ejohnstown (approved):

PR #1171, opened 2026-08-17
internal: send a disconnect when key exchange fails

yosuke-wolfssl:
### Problem

Key exchange failures aborted the handshake without transmitting
`SSH_MSG_DISCONNECT`, leaving the peer with a dropped transport and no reason
code. RFC 8731 sec. 3 makes this normative for curve25519/448: clients and
servers *"MUST also abort if the length of the received public keys are not the
expected lengths. An abort for these purposes is defined as a disconnect
(SSH_MSG_DISCONNECT) of the session and SHOULD use the
SSH_DISCONNECT_KEY_EXCHANGE_FAILED reason."*

### Fix (`src/internal.c`)

Three guards at KEX message boundaries:

| Handler | Role | Error | Reason sent |
|---|---|---|---|
| `DoKexDhInit()` | server | `WS_CRYPTO_FAILED`, `WS_PUBKEY_REJECTED_E` | `3` KEY_EXCHANGE_FAILED |
| `DoKexDhReply()` | client | `WS_CRYPTO_FAILED` | `3` KEY_EXCHANGE_FAILED |
| `DoKexDhReply()` | client | `WS_PUBKEY_REJECTED_E` | `9` HOST_KEY_NOT_VERIFIABLE |
| `DoKexDhGexGroup()` | client | `WS_CRYPTO_FAILED`, `WS_DH_SIZE_E` | `3` KEY_EXCHANGE_FAILED |

`WS_PUBKEY_REJECTED_E` reaches the two handlers from different sources, so it
maps to different reasons. In `DoKexDhInit()` it is the peer's `e` failing an
input check, which is a key exchange failure. In `DoKexDhReply()` it is only
the `publicKeyCheckCb` rejection — an application trust decision about the host
key. The `(void)` cast prevents `SendDisconnect()`'s own status from masking
the KEX error.

The finding's second half, an explicit 32-byte curve25519 length check, is
redundant: `wc_curve25519_check_public()` already rejects any length other than
`CURVE25519_KEYSIZE` before doing anything else.

Closes `f-8838`.

### Tests (`tests/regress.c`)

Duplex-harness mutator modes, plus outbound-disconnect recording on both
endpoints. Seven cases assert the reason code reaches the wire:

| Case | Role | Error |
|---|---|---|
| Truncated `f` | client | `WS_CRYPTO_FAILED` |
| Truncated `e` | server | `WS_CRYPTO_FAILED` |
| Zero-length `e` | server | `WS_PUBKEY_REJECTED_E` |
| GEX prime under the floor | client | `WS_DH_SIZE_E` |
| GEX generator below 2 | client | `WS_CRYPTO_FAILED` |
| No pubkey-check callback | client | `WS_PUBKEY_REJECTED_E` |
| Callback rejects host key | client | `WS_PUBKEY_REJECTED_E` |

A successful handshake asserts no disconnect, so the tests cannot pass on a
detector that always fires. DH and ML-KEM are covered by construction: a
truncated DH mpint is still a valid group element.

Two harness changes support them. `LocateSinglePacketPayload()` factors out the
packet framing walk that all three single-packet rewriters now share, and the
harness KEX algorithm falls back to `curve25519-sha256`, then
`ecdh-sha2-nistp256`, when no plain `diffie-hellman-group*` is built — without
that fallback the whole block compiles out and the suite reports green while
running nothing.

### Verification

- Clean under `gcc-13 -Werror` across 6 configs; `make check` 10 pass, 1 skip.
- ASan + UBSan clean, 0 reports.
- Negative controls, each run on its own: deleting a guard fails the test that
  pins it; narrowing the GEX guard to `WS_DH_SIZE_E` alone fails the
  bad-generator case while the shrunk-prime case still passes; forcing
  `LocateSinglePacketPayload()` to report no-match fails the signature-mutation
  tests; and removing the KEX algorithm fallback compiles all five
  disconnect tests out of a build with the plain DH groups disabled.

### Not in this PR

- Host key **signature** verification failures (`WS_RSA_E`, `WS_ECC_E`,
  `WS_ED25519_E`, `WS_MLDSA_E`) still abort without a disconnect. No RFC
  mandates a disconnect there, and unlike the callback rejection above they are
  cryptographic checks rather than an application trust decision.
- The mirror-image server-side GEX path (`DoKexDhGexRequest()` ->
  `SelectKexDhGexGroup()` -> `WS_DH_SIZE_E`) is likewise silent. RFC 4419
  sec. 3 says only that "the key exchange fails", with no disconnect mandate,
  so this is a consistency item tracked as a follow-up.
- `DoKexDhInit()`'s guard does not match `WS_DH_SIZE_E`, which the same handler
  can see: for `ID_DH_GEX_SHA256` with no cached group — a peer that skipped
  GEX_REQUEST — `SendKexDhReply()` reaches `GetDHPrimeGroup()`, which
  re-selects through `SelectKexDhGexGroup()`. Widening that guard belongs with
  the server-side GEX item above.
- A build with every plain DH group, curve25519 and ECDH nistp256 disabled but
  GEX left on still compiles the harness out, since GEX cannot serve as its
  default KEX algorithm. Covering that needs the signature tests gated
  separately from the GEX ones.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -5944,6 +5944,12 @@ static int DoKexDhInit(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
         ret = SendKexDhReply(ssh);
     }
 
+    /* RFC 8731 sec. 3: a rejected key exchange input aborts with a
+     * disconnect */
+    if (ret == WS_CRYPTO_FAILED || ret == WS_PUBKEY_REJECTED_E) {
+        (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_KEY_EXCHANGE_FAILED);
+    }
+
     return ret;
 }
 
@@ -7627,6 +7633,15 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 
     if (sigKeyBlock_ptr)
         WFREE(sigKeyBlock_ptr, ssh->ctx->heap, DYNTYPE_PRIVKEY);
+    /* RFC 4253 11.1: WS_PUBKEY_REJECTED_E here is only the host key check,
+     * which is server authentication, so it gets its own reason. */
+    if (ret == WS_CRYPTO_FAILED) {
+        (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_KEY_EXCHANGE_FAILED);
+    }
+    else if (ret == WS_PUBKEY_REJECTED_E) {
+        (void)SendDisconnect(ssh,
+                WOLFSSH_DISCONNECT_HOST_KEY_NOT_VERIFIABLE);
+    }
     WLOG(WS_LOG_DEBUG, "Leaving DoKexDhReply(), ret = %d", ret);
     return ret;
 }
@@ -8025,6 +8040,12 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
 
         *idx = begin;
         ret = SendKexDhInit(ssh);
+    }
+
+    /* A group under the enforced floor (RFC 8270) or a bad generator ends
+     * the key exchange, so tell the peer why. */
+    if (ret == WS_CRYPTO_FAILED || ret == WS_DH_SIZE_E) {
+        (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_KEY_EXCHANGE_FAILED);
     }
 
     return ret;
@@ -16281,9 +16302,8 @@ int SendKexDhGexRequest(WOLFSSH* ssh)
         }
     }
 
-    /* RFC 4419 sec. 3 requires min <= preferred <= max on the wire. Clamp
-     * preferred into that range in both directions; riding it up with a raised
-     * min or down under a lowered max keeps the advertised triple ordered. */
+    /* Keep the advertised triple ordered: clamp preferred into [min, max]
+     * in both directions. */
     if (ret == WS_SUCCESS) {
         if (ssh->handshake->dhGexPreferredSz < ssh->handshake->dhGexMinSz)
             ssh->handshake->dhGexPreferredSz = ssh->handshake->dhGexMinSz;
--- a/tests/regress.c
+++ b/tests/regress.c
@@ -493,6 +493,10 @@ static void TestKbInfoResponseMismatchKeepsFraming(void)
         #define KEXDH_REPLY_REGRESS_KEX_ALGO "diffie-hellman-group14-sha1"
     #elif !defined(WOLFSSH_NO_DH_GROUP1_SHA1)
         #define KEXDH_REPLY_REGRESS_KEX_ALGO "diffie-hellman-group1-sha1"
+    #elif !defined(WOLFSSH_NO_CURVE25519_SHA256)
+        #define KEXDH_REPLY_REGRESS_KEX_ALGO "curve25519-sha256"
+    #elif !defined(WOLFSSH_NO_ECDH_SHA2_NISTP256)
+        #define KEXDH_REPLY_REGRESS_KEX_ALGO "ecdh-sha2-nistp256"
     #endif
 #endif
 
@@ -564,10 +568,29 @@ static word32 LoadFileBuffer(const char* path, byte* buf, word32 bufSz)
     #define REGRESS_DEFAULT_KEY_PATH REGRESS_SERVER_KEY_ED25519_PATH
 #endif
 
-/* KEXDH_REPLY mutation modes for the duplex mutator. */
+/* KEX algorithm for the truncated peer-key tests */
+#if !defined(WOLFSSH_NO_CURVE25519_SHA256)
+    #define REGRESS_TRUNC_KEX_ALGO "curve25519-sha256"
+#elif !defined(WOLFSSH_NO_ECDH_SHA2_NISTP256)
+    #define REGRESS_TRUNC_KEX_ALGO "ecdh-sha2-nistp256"
+#endif
+
+/* KEX algorithm for the GEX group test */
+#ifndef WOLFSSH_NO_DH_GEX_SHA256
+    #define REGRESS_GEX_KEX_ALGO "diffie-hellman-group-exchange-sha256"
+    /* Well under WOLFSSH_DH_GEX_MIN_BITS, so the client's range check
+     * rejects the group before any primality work. */
+    #define REGRESS_GEX_SHRUNK_PRIME_SZ 128U
+#endif
+
 #define REGRESS_MUTATE_SIG_NAME 0
 #define REGRESS_MUTATE_SIG_DATA 1
 #define REGRESS_MUTATE_SIG_NAME_OVERRUN 2
+#define REGRESS_MUTATE_F_TRUNC 3
+#define REGRESS_MUTATE_E_TRUNC 4
+#define REGRESS_MUTATE_E_EMPTY 5
+#define REGRESS_MUTATE_GEX_GROUP_SHRINK 6
+#define REGRESS_MUTATE_GEX_GEN_BAD 7
 
 typedef struct {
     byte data[REGRESS_DUPLEX_QUEUE_SZ];
@@ -589,7 +612,9 @@ typedef struct DuplexEndpoint {
     DuplexQueue inbound;
     struct DuplexEndpoint* peer;
     KexReplyMutator* mutator;
+    word32 disconnectReason;
     byte isServer;
+    byte sawDisconnect;
 } DuplexEndpoint;
 
 typedef struct {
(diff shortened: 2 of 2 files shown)


tests/regress.c
```diff
@@ -852,6 +874,97 @@ static int RewriteKexDhReplyPacket(const byte* packet, word32 packetSz,
...
+        const byte* packet, word32 packetSz)
+{
+    word32 offset = 0;
+
+    while (packetSz - offset >= UINT32_SZ + PAD_LENGTH_SZ + MSG_ID_SZ) {
+        word32 curPacketSz = ReadUint32(packet + offset) + UINT32_SZ;
+
+        if (curPacketSz > packetSz - offset || curPacketSz < UINT32_SZ) {
+            return;
+        }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

tests/regress.c
```diff
@@ -852,6 +874,102 @@ static int RewriteKexDhReplyPacket(const byte* packet, word32 packetSz,
...
+         * padding, so a shorter declared packet cannot hold the id read
+         * below. */
+        if (curPacketSz > packetSz - offset ||
+                curPacketSz < UINT32_SZ + PAD_LENGTH_SZ + MSG_ID_SZ +
+                        MIN_PAD_LENGTH) {
+            return;
+        }
+
+        if (packet[offset + UINT32_SZ + PAD_LENGTH_SZ] == MSGID_DISCONNECT &&
+                curPacketSz >= UINT32_SZ + PAD_LENGTH_SZ + MSG_ID_SZ +
```
yosuke-wolfssl: Fixed. The reason code is now bounded against the payload instead of the buffer:

```c
packetLen = curPacketSz - UINT32_SZ;
padLen = packet[offset + UINT32_SZ];
if (... == MSGID_DISCONNECT &&
        packetLen >= padLen + PAD_LENGTH_SZ + MSG_ID_SZ + UINT32_SZ)
```

Written as an addition on the right so the comparison cannot underflow when `padLen` is large.

tests/regress.c
```diff
@@ -922,6 +1049,16 @@ static int DuplexSend(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
         }
     }
 
+    /* Only scan plaintext records. AEAD modes send the packet length as
+     * cleartext AAD, so an encrypted payload byte can look like a message
+     * id and trip the sniffer. Every disconnect under test predates NEWKEYS. */
+    if (ssh != NULL && ssh->encryptId == ID_NONE &&
```
yosuke-wolfssl, line 1185: Correct, and kept as-is for now. Gating on `ssh->encryptId` is what fixed a real CI flake: AEAD sends the packet length as cleartext AAD, so an encrypted payload byte could look like a message id (reproduced at 4/25 runs, 0/100 after the gate).

Every disconnect under test is sent before NEWKEYS, so detection is unaffected; only the breadth of the success-path negative control is reduced. The per-endpoint NEWKEYS latch would restore that write and is a fair follow-up.

src/internal.c
```diff
@@ -7391,6 +7397,11 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
 
     if (sigKeyBlock_ptr)
         WFREE(sigKeyBlock_ptr, ssh->ctx->heap, DYNTYPE_PRIVKEY);
+    /* RFC 4253 7.1: tell the peer the key exchange failed. Both codes are
+     * fatal, never a would-block. */
+    if (ret == WS_CRYPTO_FAILED || ret == WS_PUBKEY_REJECTED_E) {
```
yosuke-wolfssl: Fixed. Added the same guard at the tail of `DoKexDhGexGroup()`. Confirmed reachable: `diffie-hellman-group-exchange-sha256` is in `cannedKexAlgoNames[]` by default, and the five `ValidateKexDhGexGroup()` rejections are peer-driven (non-prime, even prime, generator out of range, bit range).

Note the server side differs: `DoKexDhGexRequest()` -> `SelectKexDhGexGroup()` returns `WS_DH_SIZE_E`, not `WS_CRYPTO_FAILED`, so this guard does not cover an unsatisfiable client range. Left out deliberately rather than widening the condition here.

src/internal.c
```diff
@@ -7391,6 +7397,11 @@ static int DoKexDhReply(WOLFSSH* ssh, byte* buf, word32 len, word32* idx)
...
+        (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_KEY_EXCHANGE_FAILED);
```
yosuke-wolfssl, line 7639: Confirmed as a real issue, but pre-existing and out of scope here. This PR touches no cipher-install code.

`BundlePacket()` gates encryption on `ssh->encryptId` alone, so any send after a failed `CipherSetKey()` reaches `Encrypt()` with a torn-down context: a rekey whose install fails leaves the old cipher freed and `encryptId` still set, and `wolfSSH_SendDisconnect()` from application code does the same. Both paths exist without this change.

The guard here adds another call site into that pre-existing defect, not the defect itself. Filing separately; the fix belongs in `SendNewKeys()` (clear `encryptId` when `CipherSetKey()` fails) and wants its own test for a failing cipher install.

src/internal.c
```diff
@@ -7791,6 +7802,12 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
         ret = SendKexDhInit(ssh);
     }
 
+    /* RFC 4253 7.1: tell the peer the key exchange failed. Both codes are
+     * fatal, never a would-block. */
+    if (ret == WS_CRYPTO_FAILED || ret == WS_PUBKEY_REJECTED_E) {
```
yosuke-wolfssl: Fixed

tests/regress.c
```diff
@@ -5957,6 +6195,11 @@ int main(int argc, char** argv)
     TestKexDhReplyRejectsEd25519CorruptSig();
     #endif
     TestKexDhReplyRejectsSigNameOverrun();
+    #ifdef REGRESS_TRUNC_KEX_ALGO
+    TestKexDhReplyTruncatedFSendsDisconnect();
+    TestKexDhInitTruncatedESendsDisconnect();
+    TestKexDhInitEmptyESendsDisconnect();
```
yosuke-wolfssl, line 7028: Won't fix

src/internal.c
yosuke-wolfssl, line 7639: Fixed

src/internal.c
```diff
@@ -7791,6 +7802,12 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
...
+        (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_KEY_EXCHANGE_FAILED);
```
yosuke-wolfssl, line 8048: Won't fix

ejohnstown (changes requested):
I sent a change request directly.

src/internal.c
```diff
@@ -7791,6 +7806,12 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
         ret = SendKexDhInit(ssh);
     }
 
+    /* RFC 4253 7.1: WS_DH_SIZE_E is the GEX modulus range rejection,
+     * which only this handler raises. */
+    if (ret == WS_CRYPTO_FAILED || ret == WS_DH_SIZE_E) {
```
yosuke-wolfssl, line 8047: Fixed

tests/regress.c
```diff
@@ -1290,6 +1558,131 @@ static void TestKexDhReplyRejectsSigNameOverrun(void)
     FreeKexReplyHarness(&harness);
 }
 
+#ifdef REGRESS_TRUNC_KEX_ALGO
+/* A short peer public key fails the key exchange, and RFC 4253 7.1 wants the
```
ejohnstown: Should be RFC 8731 sec. 3.

src/internal.c
```diff
@@ -8019,6 +8034,12 @@ static int DoKexDhGexGroup(WOLFSSH* ssh,
         ret = SendKexDhInit(ssh);
     }
 
+    /* RFC 4419 sec. 3: a group outside the requested range fails the key
```
ejohnstown: The DoKexDhGexGroup() comment attributes to RFC 4419 sec. 3 a rule that section does not contain – “the key exchange fails” attaches only to the e/f range check; group sizing is SHOULD-level. RFC 8270 is already cited nearby for the real floor.

ejohnstown (changes requested):
Also, the PR body’s deferral of F3 (from private issue list) covers only half of F3. The second half - DoKexDhInit()‘s guard omits WS_DH_SIZE_E, reachable via GetDHPrimeGroup() to SelectKexDhGexGroup() – is in your private response document but not in the body.

tests/regress.c
yosuke-wolfssl: Fixed in `0edb38e1`, now at `tests/regress.c:1583`:

```c
/* RFC 8731 sec. 3: a peer public key of the wrong length aborts with a
 * disconnect. The truncated key is rejected during key agreement, before
 * signature verification. */
```

Same section `DoKexDhInit()`'s comment cites, so the test and the code it
exercises point at the same authority now.

src/internal.c
yosuke-wolfssl: Fixed in `0edb38e1`, now at `src/internal.c:8045`:

```c
/* A group under the enforced floor (RFC 8270) or a bad generator ends
 * the key exchange, so tell the peer why. */
```

Dropped the causal claim rather than keeping sec. 3 alongside RFC 8270 — the
floor this guard enforces is `WOLFSSH_DH_GEX_MIN_BITS` (2048), which is
RFC 8270's recommendation. Confirmed against the text: "the key exchange fails"
appears once in sec. 3, attached to e or f outside [1, p-1], as you said.

The same misattribution had spread to two more comments, both fixed in that
commit: `tests/regress.c:1655`, and `src/internal.c:16305` which claimed sec. 3
"requires min <= preferred <= max" — it states no such requirement. That last
one is outside this change; say the word and I'll drop it.

ejohnstown (approved):

PR #1173, opened 2026-08-18
Fix Windows SFTP open flags and wolfsshd -D argument parsing(f8827)

miyazakh:
## Summary

- `wolfSSH_SFTP_RecvOpen()`'s Windows (`USE_WINDOWS_API`) path built `dwCreationDisposition` by OR-ing together `OPEN_EXISTING` / `CREATE_ALWAYS` bits, but `CreateFile()`'s creation-disposition parameter is a single enumerated value, not a bitmask. TRUNC and EXCL were also never wired up (left under `#if 0`), and APPEND access was missing. Add `SFTP_WinCreationDisp()` to resolve the SFTP CREAT/EXCL/TRUNC flag combination to the one correct `CreateFile()` disposition, and OR in `FILE_APPEND_DATA` for `WOLFSSH_FXF_APPEND`.
- `wolfsshd`'s `-D` (foreground/no-daemon) argument check on Windows compared `cmdArgs[i]` with `WSTRCMP`, but `cmdArgs` entries come from `CommandLineToArgvW` and are wide strings. Comparing them as narrow `char*` data meant `-D` was never recognized. Use `wcscmp()` against `L"-D"` instead.
- Extend the `WOLFSSH_TEST_INTERNAL` regression-test plumbing in `wolfsftp.c`/`wolfsftp.h` to build under `USE_WINDOWS_API` too (it was previously guarded out on Windows), except for `wolfSSH_SFTP_TestInvalidateHeadFd()`, which stays POSIX-only since it manipulates a raw fd and Windows tracks a `HANDLE` instead.
- Add `TestSftpWindowsOpenFlagMatrix()` (`tests/regress.c`), which walks the `RecvOpen` CREAT/EXCL/TRUNC flag matrix on Windows and checks both the open result and the resulting file state for each case:
  - `WRITE` only, no `CREAT`, missing file -> fails, file not created
  - `WRITE|CREAT`, missing file -> creates it (`OPEN_ALWAYS`)
  - `WRITE|CREAT`, existing file -> opens without truncating
  - `WRITE|CREAT|TRUNC`, existing file -> truncates immediately
  - `WRITE|CREAT|EXCL`, existing file -> fails
  - `WRITE|CREAT|EXCL`, missing file -> succeeds
  - `READ|WRITE|CREAT`, missing file -> creates it

## Testing

- Built and ran the full test suite on Windows via MSYS2 MinGW64 (`_WIN32` -> `USE_WINDOWS_API`):

  ```
  PASS: tests/api.test.exe
  PASS: tests/testsuite.test.exe
  PASS: tests/kex.test.exe
  PASS: tests/regress.test.exe
  PASS: tests/unit.test.exe
  ```

  `tests/regress.test.exe` includes the new `TestSftpWindowsOpenFlagMatrix()`, exercising the corrected `CreateFile()` disposition logic end to end.
- `scripts/external.test` and `scripts/fwd.test` are skipped on Windows as expected (external network / Unix-only port forwarding).

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -2895,31 +2895,27 @@ static int StartSSHD(int argc, char** argv)
     if (isDaemon) {
         /* Set the logging to go to OutputDebugString */
         wolfSSH_SetLoggingCb(ServiceDebugCb);
+    }
 
-        if (ret == WS_SUCCESS) {
-            /* argv is being rebuilt from cmdArgs here, so argc must match
-             * cmdArgC, the count CommandLineToArgvW gave for that array. */
-            argc = (DWORD)cmdArgC;
+    if (ret == WS_SUCCESS) {
+        /* Rebuild argv from cmdArgs, not the caller's wargv: wargv is
+         * narrow data when called from main(), so it may not be a real
+         * wide string. cmdArgs is always correct either way. */
+        argc = (DWORD)cmdArgC;
 
-            /* we want the arguments to be normal char strings not wchar_t */
-            argv = (char**)WMALLOC(argc * sizeof(char*), NULL, DYNTYPE_SSHD);
-            if (argv == NULL) {
-                ret = WS_MEMORY_E;
-            }
-            else {
-                unsigned int z;
-                for (z = 0; z < argc; z++) {
-                    argv[z] = _convertHelper(cmdArgs[z], NULL);
-                }
+        /* we want the arguments to be normal char strings not wchar_t */
+        argv = (char**)WMALLOC(argc * sizeof(char*), NULL, DYNTYPE_SSHD);
+        if (argv == NULL) {
+            ret = WS_MEMORY_E;
+        }
+        else {
+            unsigned int z;
+            for (z = 0; z < argc; z++) {
+                argv[z] = _convertHelper(cmdArgs[z], NULL);
             }
         }
     }
-    else {
-        /* argc is left as the caller-supplied value here: it already
-         * matches wargv, which comes from the CRT's own command-line
-         * parser and is independent of CommandLineToArgvW's cmdArgC. */
-        argv = (char**)wargv;
-    }
+    (void)wargv;
 #endif
 
     signal(SIGINT, interruptCatch);
@@ -2936,7 +2932,8 @@ static int StartSSHD(int argc, char** argv)
         }
     }
 
-    while ((ch = mygetopt(argc, argv, "?f:p:h:dDE:o:t")) != -1) {
+    while (ret == WS_SUCCESS &&
+            (ch = mygetopt(argc, argv, "?f:p:h:dDE:o:t")) != -1) {
         switch (ch) {
         case 'f':
             configFile = myoptarg;
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -2846,6 +2846,23 @@ static char* _convertHelper(WCHAR* in, void* heap) {
     return ret;
 }
 
+/* free the argv/cmdArgs buffers built from the wide command line. Safe to
+ * call on partially initialized state: NULL argv or cmdArgs is ignored. */
+static void _freeWinArgs(char** argv, DWORD argc, LPWSTR* cmdArgs)
+{
+    DWORD z;
+
+    if (argv != NULL) {
+        for (z = 0; z < argc; z++) {
+            WFREE(argv[z], NULL, DYNTYPE_SSHD);
+        }
+        WFREE(argv, NULL, DYNTYPE_SSHD);
+    }
+    if (cmdArgs != NULL) {
+        LocalFree(cmdArgs);
+    }
+}
+
 static void StartSSHD(DWORD argc, LPTSTR* wargv)
 #else
 static int StartSSHD(int argc, char** argv)
@@ -2996,6 +3013,7 @@ static int StartSSHD(int argc, char** argv)
             #ifndef _WIN32
             return WS_FATAL_ERROR;
             #else
+            _freeWinArgs(argv, argc, cmdArgs);
             return;
             #endif
         #endif
@@ -3009,6 +3027,7 @@ static int StartSSHD(int argc, char** argv)
         #ifndef _WIN32
             return WS_SUCCESS;
         #else
+            _freeWinArgs(argv, argc, cmdArgs);
             return;
         #endif
 
@@ -3017,6 +3036,7 @@ static int StartSSHD(int argc, char** argv)
         #ifndef _WIN32
             return WS_SUCCESS;
         #else
+            _freeWinArgs(argv, argc, cmdArgs);
             return;
         #endif
         }
@@ -3149,11 +3169,9 @@ static int StartSSHD(int argc, char** argv)
                 if (SetServiceStatus(serviceStatusHandle, &serviceStatus) == FALSE) {
                     wolfSSH_Log(WS_LOG_ERROR, "[SSHD] Issue updating service status");
                 }
+                _freeWinArgs(argv, argc, cmdArgs);
                 return;
             }
-        }
-        if (cmdArgs != NULL) {
-            LocalFree(cmdArgs);
         }
     }
 #endif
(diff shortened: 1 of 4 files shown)


src/wolfsftp.c
```diff
@@ -2651,23 +2677,14 @@ int wolfSSH_SFTP_RecvOpen(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
-        if (reason & WOLFSSH_FXF_EXCL)
-            creationDisp |= CREATE_NEW;
-        if (reason & WOLFSSH_FXF_APPEND)
-            desiredAccess |= FILE_APPEND_DATA;
-    #endif
-    }
+    if (reason & WOLFSSH_FXF_APPEND)
+        desiredAccess |= FILE_APPEND_DATA;
+
+    creationDisp = SFTP_WinCreationDisp(reason);
```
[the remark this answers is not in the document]
miyazakh: Good catch, confirmed. On Windows, `FILE_APPEND_DATA` only forces writes to EOF when the handle lacks `FILE_WRITE_DATA` — but `GENERIC_WRITE` grants that implicitly, and most clients open append files with both `SSH_FXF_WRITE | SSH_FXF_APPEND`, so `WriteFile()` was honoring the client-supplied offset instead of appending. POSIX was fine since `O_APPEND` makes `pwrite()` ignore the offset.

Fixed in `src/wolfsftp.c`: added an `isAppend` flag to `WS_FILE_LIST`, set from `WOLFSSH_FXF_APPEND` in `RecvOpen`. In the Windows `RecvWrite` path, when that flag is set, the offset is now replaced with the current EOF (`GetFileSizeEx`) before the write, instead of trusting the client.

src/wolfsftp.c
```diff
@@ -2651,23 +2677,14 @@ int wolfSSH_SFTP_RecvOpen(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
...
-        if (reason & WOLFSSH_FXF_TRUNC)
-            creationDisp |= TRUNCATE_EXISTING;
-        if (reason & WOLFSSH_FXF_EXCL)
-            creationDisp |= CREATE_NEW;
-        if (reason & WOLFSSH_FXF_APPEND)
-            desiredAccess |= FILE_APPEND_DATA;
-    #endif
-    }
+    if (reason & WOLFSSH_FXF_APPEND)
+        desiredAccess |= FILE_APPEND_DATA;
```
miyazakh, line 2688: **Not a bug — false positive.**

`RecvWrite` already forces the offset to current EOF whenever `isAppend` is set (src/wolfsftp.c:4440-4455), regardless of whether `WRITE` was requested. That's exactly the required `SSH_FXF_APPEND` contract, not a violation of it.

The pre-PR open failure (`creationDisp == 0`) is now fixed as a side effect. No change needed.

tests/regress.c
```diff
@@ -3517,6 +3519,195 @@ static void TestSftpCloseFailureRemovesHandle(void)
...
+    SftpPutU32(WOLFSSH_HANDLE_ID_SZ, pkt + idx); idx += UINT32_SZ;
+    WMEMCPY(pkt + idx, handle, WOLFSSH_HANDLE_ID_SZ);
+    idx += WOLFSSH_HANDLE_ID_SZ;
+    AssertIntEQ(wolfSSH_SFTP_RecvClose(ssh, rid++, pkt, idx), WS_SUCCESS);
+
+    /* WRITE|CREAT|TRUNC: must truncate the existing file immediately. */
+    idx = 0;
+    SftpPutU32(pathSz, pkt + idx); idx += UINT32_SZ;
+    WMEMCPY(pkt + idx, path, pathSz); idx += pathSz;
+    SftpPutU32(WOLFSSH_FXF_WRITE | WOLFSSH_FXF_CREAT | WOLFSSH_FXF_TRUNC,
```
miyazakh, line 5529: Fixed

src/wolfsftp.c
```diff
@@ -2326,6 +2326,32 @@ static void SFTP_HandleIdNext(WOLFSSH* ssh, word32 id[2])
...
+        if (reason & WOLFSSH_FXF_EXCL)
+            disp = CREATE_NEW;
+        else if (reason & WOLFSSH_FXF_TRUNC)
+            disp = CREATE_ALWAYS;
+        else
+            disp = OPEN_ALWAYS;
+    }
+    else {
+        if (reason & WOLFSSH_FXF_TRUNC)
+            disp = TRUNCATE_EXISTING;
```
miyazakh, line 2350: Fixed

src/wolfsftp.c
miyazakh, line 2688: **Duplicate / false positive.**

Already handled at src/wolfsftp.c:4443-4457: when `isAppend` is set, `RecvWrite` overrides the client-supplied offset with the current EOF (via `GetFileSizeEx`) before calling `WriteFile`. `isAppend` is set for any open with `WOLFSSH_FXF_APPEND`, including `WRITE|APPEND`.

The suggested fix (track APPEND, seek to `FILE_END` in `RecvWrite`) is functionally identical to what's already there — same effect, different API (`GetFileSizeEx`+explicit offset vs `SetFilePointer(FILE_END)`).

No change needed.

apps/wolfsshd/wolfsshd.c
```diff
@@ -2855,7 +2855,9 @@ static int StartSSHD(int argc, char** argv)
 
     if (ret == WS_SUCCESS) {
         for (i = 0; i < argc; i++) {
-            if (WSTRCMP((char*)(cmdArgs[i]), "-D") == 0) {
+            /* cmdArgs entries are wide strings (CommandLineToArgvW); compare
+             * as such instead of reinterpreting as narrow char data. */
+            if (wcscmp(cmdArgs[i], L"-D") == 0) {
```
miyazakh, line 2906: Fixed

tests/regress.c
```diff
@@ -3517,6 +3519,195 @@ static void TestSftpCloseFailureRemovesHandle(void)
 
 #endif /* !NO_WOLFSSH_SERVER && !USE_WINDOWS_API && !NO_FILESYSTEM */
 
+#if !defined(NO_WOLFSSH_SERVER) && defined(USE_WINDOWS_API) && \
+        !defined(NO_FILESYSTEM)
+/* Walks the RecvOpen CREAT/EXCL/TRUNC flag matrix on Windows, checking both
+ * the open result and the resulting file state for each case. */
+static void TestSftpWindowsOpenFlagMatrix(void)
```
miyazakh, line 5419: Both cases it asks for already exist in `tests/regress.c`, added earlier in this session.

ejohnstown (changes requested):
I sent comments directly.

philljj (dismissed):
Has merge conficts in src/wolfsftp.c

src/wolfsftp.c
```diff
@@ -4413,6 +4438,25 @@ int wolfSSH_SFTP_RecvWrite(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
         }
         offset.Offset = (DWORD)strSz;
 
+        /* WOLFSSH_FXF_APPEND was requested at open: FILE_APPEND_DATA alone
+         * does not force writes to EOF once the handle also carries
+         * FILE_WRITE_DATA (granted implicitly by GENERIC_WRITE), so the
+         * client-supplied offset must be overridden with the current EOF. */
+        if (isAppend) {
+            if (GetFileSizeEx(fd, &fileSize) == 0) {
```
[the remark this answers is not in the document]
miyazakh: Fixed

apps/wolfsshd/wolfsshd.c
```diff
@@ -2894,24 +2895,27 @@ static int StartSSHD(int argc, char** argv)
...
-                    argv[z] = _convertHelper(cmdArgs[z], NULL);
-                }
+    if (ret == WS_SUCCESS) {
+        /* Rebuild argv from cmdArgs, not the caller's wargv: wargv is
+         * narrow data when called from main(), so it may not be a real
+         * wide string. cmdArgs is always correct either way. */
+        argc = (DWORD)cmdArgC;
+
+        /* we want the arguments to be normal char strings not wchar_t */
+        argv = (char**)WMALLOC(argc * sizeof(char*), NULL, DYNTYPE_SSHD);
```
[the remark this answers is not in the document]
miyazakh: Fixed

ejohnstown (approved):

PR #1174, opened 2026-08-18
tests: fix wolfSSHd harness bugs that hid skipped and failed tests

ejohnstown:
Three bugs in the wolfSSHd test harness let the suite report a pass while silently skipping tests or blaming the daemon for a broken client invocation. All changes are under `apps/wolfsshd/test/`, plus one `.gitignore` line; no library or app code is touched.

StrictModes count aborting the suite

- `grep -c` prints 0 and exits 1 on no match, so the `|| echo 0` fallback made the count "0\n0"; the arithmetic error unwound bash out of the test block, skipping the last eleven tests while the summary still passed.
- Added a `RUN_COMPLETE` sentinel checked before the summary, and matched the teardown `pkill` on the process name so it cannot kill the run.

`$PWD` used as a saved directory

- Eight scripts saved their start directory in `PWD`, which the shell rewrites on every `cd`, so `sshd_forcedcmd_test.sh`'s second scenario and the `sshd_x509_upn_fail.sh` log count never ran as written; save it in `TESTDIR` instead.
- Stop the daemon from a trap in `sshd_forcedcmd_test.sh`, and let `start_wolfsshd`'s `ps | grep` pipelines fail under `set -e`.

Privdrop test's client key paths

- The clients call `ChangeToWolfSshRoot()` before parsing arguments, so the relative key paths failed and every client died at "Error setting private key" -- read as a privilege-drop failure. Anchor the paths at the script's own directory.
- Print the client's own output at every failure exit.

### DIFF
--- a/apps/wolfsshd/test/sshd_forcedcmd_test.sh
+++ b/apps/wolfsshd/test/sshd_forcedcmd_test.sh
@@ -8,11 +8,26 @@ if [ -z "$1" ] || [ -z "$2" ]; then
     exit 1
 fi
 
-PWD=`pwd`
+# Not named PWD: the shell rewrites that variable on every cd, so a saved copy
+# would not survive the cd to the repository root below.
+TESTDIR=`pwd`
 USER=`whoami`
 TEST_PORT="$2"
 TEST_HOST="$1"
 source ./start_sshd.sh
+
+# Stop the daemon on every exit path. From the "set -e" below onward an aborted
+# client run would otherwise leave a root daemon holding the shared test port,
+# and every later test in the suite would talk to this config.
+cleanup() {
+    if [ -n "$PID" ]; then
+        stop_wolfsshd
+        PID=""
+    fi
+    return 0
+}
+trap cleanup EXIT
+
 cat <<EOF > sshd_config_test_forcedcmd
 Port $TEST_PORT
 Protocol 2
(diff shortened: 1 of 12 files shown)


apps/wolfsshd/test/start_sshd.sh
```diff
@@ -6,7 +6,9 @@ SSHD_KEYDIR=""
 
 # starts up a sshd session, takes in the sshd_config file as an argument
 start_wolfsshd() {
-    CURRENT_PIDS=`ps -e | grep wolfsshd | grep -oE "[0-9]+"`
+    # No daemon running is the normal case here, and grep exits 1 on no match,
+    # which would end a caller running under "set -e".
+    CURRENT_PIDS=`ps -e | grep wolfsshd | grep -oE "[0-9]+"` || true
```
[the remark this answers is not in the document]
ejohnstown: Confirmed and fixed (folded into this branch's TESTDIR commit): both snapshots now use pgrep -x wolfsshd | sort -n, since ps -e mixed the TIME field's clock digits in with the PID — with leftover daemons that picked a pre-existing process to stop, or a clock digit (sudo kill 1) once one accrued a second of CPU.

apps/wolfsshd/test/sshd_forcedcmd_test.sh
```diff
@@ -49,8 +64,9 @@ fi
 set -e
 echo exit | $TEST_SFTP -u $USER -i $PRIVATE_KEY -j $PUBLIC_KEY -h $TEST_HOST -p $TEST_PORT
 
-cd $PWD
+cd "$TESTDIR"
 stop_wolfsshd
```
padelsbach, line 63: AI says the `sudo kill $PID` in `stop_wolfsshd` will bypass `cleanup()` and return 1 instead of zero. Do we need to guard the `kill`?

ejohnstown: Yes. cleanup() did run -- it just killed the dead pid again. The problem was set -e aborting there before PID was cleared, so the ForceCommand-SFTP scenario was silently skipped. stop_wolfsshd now guards the kill and returns 0.

padelsbach (approved):

PR #1178, opened 2026-08-18
wolfsshd: match sshd's host key file checks

ejohnstown:
wolfsshd ran the host key, host certificate and TrustedUserCAKeys through wolfSSHD_OpenSecureFile(), rejecting symlinks, foreign owners and group writable ancestors that sshd accepts, with no sshd_config knob to relax it.

- Open the host certificate and TrustedUserCAKeys plainly, unchecked.
- Add wolfSSHD_OpenHostKeyFile() applying sshd's host key rule.
- Also refuse non-regular files, and open O_NONBLOCK, so a FIFO left in place of the key cannot stall startup.
- Leave wolfSSHD_OpenSecureFile() gating authorized_keys and shadow.
- Assert the new host key behaviour in the suite, and drop the root-owned trust anchor copies from start_sshd.sh.

Issue: F-5901

### DIFF
--- a/apps/wolfsshd/auth.c
+++ b/apps/wolfsshd/auth.c
@@ -1270,6 +1270,80 @@ int wolfSSHD_OpenSecureFile(const char* path, WUID_T ownerUid,
 #endif
 }
 
+/* Open the host private key. See auth.h for the contract. */
+int wolfSSHD_OpenHostKeyFile(const char* path, WFILE** out)
+{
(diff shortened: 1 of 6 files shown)


ejohnstown:
This has been superseded by PR #1186. This was fixing the wrong problem the incorrect way. StrictModes is an option for user authentication keys, not the host's keys.

PR #1180, opened 2026-08-19
Add a build option probe for the test scripts

ejohnstown:
- Test scripts gate on it with `grep -qx` instead of grepping usage text, `config.log`, daemon logs, or the client's algorithm list.
- Converted: `sftp.test`, `get-put.test`, `scp.test`, `fwd.test`, `external.test`, `run_all_sshd_tests.sh`, `ssh_kex_algos.sh`, `sshd_ossh_cert_test.sh`, `sshd_x509_upn_fail.sh`, `sshd_empty_password_test.sh`.
- `sshd_bad_sftp_test.sh`, `sshd_large_sftp_test.sh` and `sshd_scp_fail.sh` had no skip and failed outright without SFTP or SCP; they now skip.
- ML-DSA gate now checks the composite algorithm `MLDSA87_ES384`, since the ECDSA half can be missing on its own.
- Drops the usage lines that existed only for scripts to grep: the FPKI block in `wolfsshd` and the symlink note in the example client.

### DIFF
--- a/apps/wolfssh-options.c
+++ b/apps/wolfssh-options.c
@@ -1,0 +1,175 @@
+/* wolfssh-options.c
+ *
+ * Copyright (C) 2014-2026 wolfSSL Inc.
+ *
+ * This file is part of wolfSSH.
+ *
+ * wolfSSH is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfSSH is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with wolfSSH.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+/*
+ * Build option probe for the test scripts. Prints each enabled build option,
+ * one per line. Not installed; a build tree artifact only.
+ *
+ *     OPTIONS=$(./apps/wolfssh-options) || exit 1
+ *     if ! echo "$OPTIONS" | grep -qx "FPKI"; then
+ *         echo "built without FPKI, skipping"
+ *         exit 77
+ *     fi
+ *
+ * scripts/ runs from the build root and calls "./apps/wolfssh-options"; the
+ * wolfSSHd tests source ./wolfssh_options.sh for it. Match whole lines, so one
+ * option name cannot match another that has it as a prefix. A probe that will
+ * not run is a build problem, not an option being off.
+ *
+ * Each option prints under the same guard the library uses, so the output is
+ * the preprocessor's answer for this build. Covers every configure
+ * --enable/--disable that sets a macro, plus a few facts from the wolfSSL
+ * build; not --enable-examples, which defines nothing to test.
+ *
+ * To add an option, add a guard and a printf() below.
+ */
+
+
+#ifdef HAVE_CONFIG_H
+    #include <config.h>
+#endif
+#ifdef WOLFSSL_USER_SETTINGS
+    #include <wolfssl/wolfcrypt/settings.h>
+#else
+    #include <wolfssl/options.h>
+#endif
+#include <wolfssh/ssh.h>
+#include <wolfssh/internal.h>
+#include <stdio.h>
+
+/* internal.h derives the WOLFSSH_NO_* options from the wolfSSL build. */
+#ifndef _WOLFSSH_INTERNAL_H_
+    #error "wolfssh-options.c requires wolfssh/internal.h"
+#endif
+
+
+int main(void)
+{
+    /* Library features. NO_INLINE is shared with wolfSSL's --disable-inline,
+     * so INLINE is the effective state, not wolfSSH's configure answer. */
+#ifndef NO_INLINE
+    printf("INLINE\n");
+#endif
+#ifndef NO_WOLFSSH_SERVER
+    printf("SERVER\n");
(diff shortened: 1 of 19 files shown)


apps/wolfssh-options.c
```diff
@@ -0,0 +1,190 @@
...
+#ifdef WOLFSSH_CERTS
+    printf("CERTS\n");
+#endif
+#ifdef WOLFSSH_OSSH_CERTS
+    printf("OSSH_CERTS\n");
+#endif
+    /* wolfSSL can read the FASC-N, UUID and UPN out of a certificate. This is
+     * what gates wolfSSHd's AuthorizedUPNDomains check. */
+#ifdef WOLFSSL_FPKI
+    printf("FPKI\n");
```
padelsbach, line 131: In the printf, can we indicate where the define comes from (wolfssl or wolfssh)? Alternatively just print the full name of the define.

Specific concern around things like `WOLFSSL_FPKI` and `WOLFSSH_NO_FPKI`

apps/wolfssh-options.c
```diff
@@ -0,0 +1,190 @@
...
+ * along with wolfSSH.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+/*
+ * Build option probe for the test scripts. Prints the name of each build
+ * option that is enabled, one per line, nothing for the ones that are not.
+ * A test script gates itself on an option like this:
+ *
+ *     OPTIONS=$(./apps/wolfssh-options) || exit 1
```
padelsbach: How do we ensure this executable is up to date with the build? Do the include.am changes trigger a build of wolfssl-options whenever the top level make is run? If not, would it make sense to do that?

apps/wolfsshd/test/sshd_x509_upn_fail.sh
```diff
@@ -7,10 +7,12 @@
 
 PWD=`pwd`
 
-# The UPN domain check is compiled only when wolfSSL is built with FPKI. Probe
-# the daemon binary's help output, which prints an FPKI marker under the same
-# build guard, and skip when the check is not present.
-if ! ../wolfsshd "-?" 2>&1 | grep -q "FPKI"; then
+# The UPN domain check is compiled only when wolfSSL is built with FPKI.
+WOLFSSH_OPTIONS=$(../../../apps/wolfssh-options) || {
```
padelsbach: Can we make this relative to the repo root instead of hardcoded paths everywhere?

apps/wolfssh-options.c
```diff
@@ -0,0 +1,190 @@
...
+    #error "wolfssh-options.c requires wolfssh/internal.h"
+#endif
+
+
+int main(void)
+{
+    /* Library features. NO_INLINE is shared with wolfSSL, whose
+     * --disable-inline sets it too, so INLINE is the effective state rather
+     * than wolfSSH's own configure answer. */
+#ifndef NO_INLINE
```
padelsbach, line 68: Some general concerns about this static C based approach:

1. Manually created seems error prone and brittle. 
2. No differentiation between not-defined versus not implemented here versus a bogus/deprecated/misspelled define altogether.
3. Copy pasta

AI tells me that lint could be used to generate this dynamically via `configure.ac`, at least for wolfSSH. Maybe we could do that on wolfSSL too?

apps/wolfssh-options.c
```diff
@@ -0,0 +1,190 @@
...
+#else
+    #include <wolfssl/options.h>
+#endif
+#include <wolfssh/ssh.h>
+#include <wolfssh/internal.h>
+#include <stdio.h>
+
+/* internal.h derives the WOLFSSH_NO_* options from the wolfSSL build. */
+#ifndef _WOLFSSH_INTERNAL_H_
+    #error "wolfssh-options.c requires wolfssh/internal.h"
```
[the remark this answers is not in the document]
ejohnstown: This was done because Skoll said someone could remove the internal.h include and then the options wouldn't print out any more. This is intended.

apps/wolfssh-options.c
ejohnstown: I'm removing the FPKI_PROFILE check. That has its own test already. None of the wolfSSHd tests look for that. I added a condition to FPKI including needed WOLFSSH_CERTS.

The idea is these are printing out the names of options that are available for specific tests driven by scripts to run. Any sort of compound check to see if an option is available will have to be duplicated in the testing script. Or we can do the work in the utility.

apps/wolfssh-options.c
ejohnstown: It is error prone, but we should also be verifying new tests run by the automation when we add them.

Like in the earlier comment, we aren't using the tool to check for macros, we are checking for options. SFTP, SCP, FPKI, etc.

configure won't be picking up the flags that are derived in internal.h.

This isn't perfect, but it is better than adding print strings to release applications to tell if a test should run, or scanning log files trying to find options.

apps/wolfsshd/test/wolfssh_options.sh
```diff
@@ -0,0 +1,26 @@
...
+#     if ! wolfssh_has SFTP; then
+#         echo "built without SFTP, skipping"
+#         exit 77
+#     fi
+#
+# The tests run from this directory, so the one path to the build tree lives
+# here. WOLFSSH_ROOT is absolute, so it survives their cd ../../.. . A probe
+# that will not run is a build problem, not an option being off.
+
+WOLFSSH_ROOT=$(cd "$(dirname "$0")/../../.." && pwd) || exit 1
```
padelsbach, line 16: If we're going to wrap it in a script, could we use `make` to rebuild the executable and ensure the flags are not stale?

ejohnstown: We already have that. configure and automake set up wolfssh-options to be dependent on internal.h and the other headers it includes. It should rebuild if any of those deps are touched.

padelsbach (approved):

PR #1181, opened 2026-08-20
wolfsftp: buffer the SFTP DATA length across partial reads

yosuke-wolfssl:
### Problem
`wolfSSH_SFTP_SendReadPacket()` decoded the 4-byte string length prefixing an SFTP `DATA` reply with a single unchecked read into a stack buffer:

```c
byte szFlat[UINT32_SZ];                       /* uninitialized */
ret = wolfSSH_stream_read(ssh, szFlat, UINT32_SZ);
if (ret < 0) { ... }                          /* short reads fall through */
ato32(szFlat, &sz);
```

`wolfSSH_stream_read()` returns `min(bufSz, available)`, so a short *positive* return is routine whenever a peer's `DATA` reply straddles SSH packets. Only `ret < 0` was checked, so 1-3 delivered bytes reached `ato32()`, which read the rest of `szFlat` as uninitialized stack (CWE-457). Because `szFlat` was a stack local, the prefix bytes already consumed were unrecoverable, so the retry could not resynchronize. Closes f-8828.

### Fix (`src/wolfsftp.c`)
`STATE_SEND_READ_FTP_DATA` now decodes the length out of `state->buffer`, which lives in `ssh->sendReadState` and so survives across calls:

- **`wolfSSH_SFTP_buffer_set_size(..., UINT32_SZ)`** trims the allocation `STATE_SEND_READ_GET_HEADER` already made rather than taking a second one, and rejects a header declaring fewer than 4 payload bytes.
- **`wolfSSH_SFTP_buffer_read()`** accumulates short positive reads and retains `idx` on `WS_WANT_READ`, so a retry resumes mid-prefix.
- **`wolfSSH_SFTP_buffer_ato32()`** decodes after a rewind, bounds-checked against the buffer.
- **`sz > outSz`** is applied before allocating for the payload, and `wolfSSH_SFTP_buffer_create()`'s return is checked.
- **`ssh->error`** is set on every failure exit -- `WS_BUFFER_E` on the size and decode helpers, `WS_RECV_OVERFLOW_E` on the `outSz` bound, `WS_MEMORY_E` on the allocation -- so the caller's `NoticeError()` cannot misclassify one as retryable.
- **`szFlat`** is removed.

This is the same transformation commit `6f0cbe3f` made for the sibling VERSION-header defect (f-7505), so both halves of the client read path now use one pattern.

### Test harness (`tests/unit.c`)
Both tests are network-free, built on the existing `SftpClientNewSession()` + `wolfSSH_TestChannelPutData()` harness. No new test hooks; `SftpBuildData()` delegates to the existing `SftpBuildVersion()`.

- **`test_SftpSendReadPacketSplit()`** stages a `DATA` reply cut at each of the four points across the length prefix (0 through 3 of the 4 bytes delivered first) and asserts the first call returns `WS_FATAL_ERROR`/`WS_WANT_READ` with the send-read state retained, and that the retry returns the full payload intact.
- **`test_SftpSendReadPacketOverflow()`** delivers the header and the whole prefix in one read so the decode completes, declaring 8 payload bytes against a 4-byte `out`, and asserts `WS_FATAL_ERROR` with `WS_RECV_OVERFLOW_E` and the send-read state freed rather than retained.

### Verification
- Negative controls: against the base `src/wolfsftp.c` the split test fails at the three split points that land mid-prefix (+1, +2, +3) and passes at +0, where the whole prefix arrives in one read; with the `ssh->error = WS_RECV_OVERFLOW_E` line removed the overflow test fails on its error-code assertion. Both pass in full with the fix.
- `tests/unit.test`: 112 passed, 0 failed.
- `make check`: 12 passed, 0 failed, 1 skipped (`external.test`) of 13.
- Clean under gcc-13 `-Werror` across 6 configs (enable-all, zephyr-defines, sftp-only, scp-only, default, smallstack).

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -8422,7 +8422,6 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
         const word32* ofst, byte* out, word32 outSz)
 {
     WS_SFTP_SEND_READ_STATE* state = NULL;
-    byte szFlat[UINT32_SZ];
     int ret = WS_SUCCESS;
     word32 sz;
 
@@ -8557,8 +8556,16 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
 
             case STATE_SEND_READ_FTP_DATA:
                 WLOG(WS_LOG_SFTP, "SFTP SEND_READ STATE: FTP_DATA");
-                /* get size of string and place it into out buffer */
-                ret = wolfSSH_stream_read(ssh, szFlat, UINT32_SZ);
+                /* Trim the buffer holding the header's payload down to the
+                 * string length, which can arrive split over several reads. */
+                ret = wolfSSH_SFTP_buffer_set_size(&state->buffer, UINT32_SZ);
+                if (ret != WS_SUCCESS) {
+                    ssh->error = WS_BUFFER_E;
+                    state->state = STATE_SEND_READ_CLEANUP;
+                    continue;
+                }
+
+                ret = wolfSSH_SFTP_buffer_read(ssh, &state->buffer, UINT32_SZ);
                 if (ret < 0) {
                     if (NoticeError(ssh)) {
                         return WS_FATAL_ERROR;
@@ -8566,11 +8573,26 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
                     state->state = STATE_SEND_READ_CLEANUP;
                     continue;
                 }
-                ato32(szFlat, &sz);
-                wolfSSH_SFTP_buffer_create(ssh, &state->buffer, sz);
-                if (wolfSSH_SFTP_buffer_size(&state->buffer) > outSz) {
+
+                /* get size of the data string */
+                wolfSSH_SFTP_buffer_rewind(&state->buffer);
+                ret = wolfSSH_SFTP_buffer_ato32(&state->buffer, &sz);
+                if (ret != WS_SUCCESS) {
+                    ssh->error = WS_BUFFER_E;
+                    state->state = STATE_SEND_READ_CLEANUP;
+                    continue;
+                }
+                if (sz > outSz) {
                     WLOG(WS_LOG_SFTP, "Server sent more data then expected");
+                    ssh->error = WS_RECV_OVERFLOW_E;
                     ret = WS_FATAL_ERROR;
+                    state->state = STATE_SEND_READ_CLEANUP;
+                    continue;
+                }
+
+                ret = wolfSSH_SFTP_buffer_create(ssh, &state->buffer, sz);
+                if (ret != WS_SUCCESS) {
+                    ssh->error = WS_MEMORY_E;
                     state->state = STATE_SEND_READ_CLEANUP;
                     continue;
                 }
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -16165,6 +16165,160 @@ static int test_SftpClientPutWriteStatusFail(void)
     return 0;
 }
 #endif /* WOLFSSH_TEST_SFTP_PUT */
+
+/* Builds an SFTP DATA reply in "out": the 9 byte header carrying the request
+ * id where a VERSION message carries the version, then the data string as the
+ * trailing bytes. Returns the size written, or 0 if it does not fit. */
+static word32 SftpBuildData(byte* out, word32 outSz, word32 reqId,
+        word32 dataSz)
+{
+    word32 msgSz;
+
+    msgSz = SftpBuildVersion(out, outSz,
+            MSG_ID_SZ + UINT32_SZ + UINT32_SZ + dataSz, WOLFSSH_FTP_DATA,
+            reqId, UINT32_SZ + dataSz);
+    if (msgSz > 0) {
+        /* the trailing bytes open with the data string length */
+        PutU32BE(out + WOLFSSH_SFTP_HEADER, dataSz);
+    }
+
+    return msgSz;
+}
+
+
+/* Drives wolfSSH_SFTP_SendReadPacket() over a DATA reply delivered in two
+ * pieces, split after "split" bytes, with RecvAlwaysWantRead standing in for a
+ * non-blocking socket. Returns 0, or a negative sentinel on a setup failure. */
+static int SftpClientDriveReadSplit(word32 dataSz, word32 split,
+        int* firstRet, int* firstErr, int* stateKept, int* secondRet,
+        byte* out, word32 outSz)
+{
+    WOLFSSH_CTX* ctx = NULL;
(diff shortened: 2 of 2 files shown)


src/wolfsftp.c
```diff
@@ -8558,16 +8557,24 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
...
                     if (NoticeError(ssh)) {
                         return WS_FATAL_ERROR;
                     }
                     state->state = STATE_SEND_READ_CLEANUP;
                     continue;
                 }
-                ato32(szFlat, &sz);
+                ato32(wolfSSH_SFTP_buffer_data(&state->buffer), &sz);
                 wolfSSH_SFTP_buffer_create(ssh, &state->buffer, sz);
                 if (wolfSSH_SFTP_buffer_size(&state->buffer) > outSz) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed but leave sz==0 path since it's pre-existing and reachable with malformed server only.
And even then, the pre-existing behavior already failed safely:

tests/unit.c
```diff
@@ -15449,6 +15449,130 @@ static int test_SftpClientRecvInitVersion(void)
...
+    wolfSSH_free(ssh);
+    wolfSSH_CTX_free(ctx);
+    return result;
+}
+
+
+/* Regression for the SFTP DATA reply losing bytes when its four byte string
+ * length arrives split: the read must report WS_WANT_READ, keep the send read
+ * state, then complete with the payload intact on the retry. */
+static int test_SftpSendReadPacketSplit(void)
```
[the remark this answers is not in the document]
yosuke-wolfssl: Accurate on the coverage, but both paths are pre-existing and outside this PR's scope.

This PR only moves the four-byte data-string length decode off the stack and into the send-read state so it survives a partial read. The `sz > outSz` bound and the `sz == 0` handling below it are unchanged.

Neither is reachable from a conforming server:

- **`sz > outSz`** — the server must honor the length the client requested. `wolfSSH_SFTP_RecvRead()` reads at most `strSz` and errors out if its own read returns more, so it cannot emit an oversized data string.
- **`sz == 0`** — end of file is signalled with `SSH_FXP_STATUS` / `SSH_FX_EOF`, not an empty `DATA`. `wolfSSH_SFTP_RecvRead()` takes the status path when the file read returns 0 bytes.

A regression in the relocated decode is already caught by the cases that are here: a wrong offset or byte order yields either a length that trips the bound (so the return no longer equals `dataSz`) or a mismatched payload.

Both are worth hardening against a malformed peer, but as a separate PR — this one stays on the reported issue.

src/wolfsftp.c
```diff
@@ -8558,17 +8557,29 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
...
                 if (ret < 0) {
                     if (NoticeError(ssh)) {
                         return WS_FATAL_ERROR;
                     }
                     state->state = STATE_SEND_READ_CLEANUP;
                     continue;
                 }
-                ato32(szFlat, &sz);
-                wolfSSH_SFTP_buffer_create(ssh, &state->buffer, sz);
+                ato32(wolfSSH_SFTP_buffer_data(&state->buffer), &sz);
```
ejohnstown: Why didn't you use wolfSSH_SFTP_buffer_ato32()?

src/wolfsftp.c
```diff
@@ -8558,17 +8557,29 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
 
             case STATE_SEND_READ_FTP_DATA:
                 WLOG(WS_LOG_SFTP, "SFTP SEND_READ STATE: FTP_DATA");
+                /* The string length can arrive split over several reads */
+                ret = wolfSSH_SFTP_buffer_create(ssh, &state->buffer,
```
ejohnstown: Would it make sense to just reuse the existing buffer we've already allocated?

ejohnstown (changes requested):

src/wolfsftp.c
yosuke-wolfssl: No good reason — switched to it. It needs a rewind first: `wolfSSH_SFTP_buffer_read()` leaves `idx == sz == UINT32_SZ`, so `buffer_ato32()`'s `idx + UINT32_SZ > sz` guard would reject the call. So it's now `wolfSSH_SFTP_buffer_rewind()` then `wolfSSH_SFTP_buffer_ato32()`, which is the same shape `STATE_SEND_READ_FTP_STATUS` uses a few lines down. Bonus is the bounds check that the bare `ato32()` didn't have.

src/wolfsftp.c
yosuke-wolfssl: Yes, done. `STATE_SEND_READ_GET_HEADER` already allocated `state->buffer` at `UINT32_SZ + dataSz`, so this is now `wolfSSH_SFTP_buffer_set_size(&state->buffer, UINT32_SZ)`, trimming it in place instead of `buffer_create()` freeing and re-allocating. `data` and `idx` are untouched, so the partial-prefix resume still works across the `WS_WANT_READ` return, and `set_size()`'s `sz > buffer->sz` guard also rejects a header that declared fewer than `UINT32_SZ` payload bytes — a bound the old code didn't have.

The `buffer_create(..., sz)` further down has to stay for now. `STATE_SEND_READ_REMAINDER` reads the payload straight into `out` and uses `buffer.sz` purely as a countdown of bytes still owed, and `set_size()` can only shrink — so `buffer_create()` is the only way to set that count. Dropping the allocation means giving `WS_SFTP_SEND_READ_STATE` its own counter field and reworking `REMAINDER`. Glad to do that as a follow-up, but `REMAINDER` is the one state this fix has to leave provably untouched, so I'd rather not fold it in here.

ejohnstown (approved):

src/wolfsftp.c
```diff
@@ -8553,24 +8552,46 @@ int wolfSSH_SFTP_SendReadPacket(WOLFSSH* ssh, byte* handle, word32 handleSz,
...
+                wolfSSH_SFTP_buffer_rewind(&state->buffer);
+                ret = wolfSSH_SFTP_buffer_ato32(&state->buffer, &sz);
+                if (ret != WS_SUCCESS) {
+                    ssh->error = WS_BUFFER_E;
+                    state->state = STATE_SEND_READ_CLEANUP;
+                    continue;
+                }
+                if (sz > outSz) {
                     WLOG(WS_LOG_SFTP, "Server sent more data then expected");
                     ret = WS_FATAL_ERROR;
```
ejohnstown, line 8588: Set ssh->error here, too. I think it should be WS_RECV_OVERFLOW_E.

ejohnstown (changes requested):

src/wolfsftp.c
yosuke-wolfssl: Fixed as suggested

ejohnstown (approved):

PR #1182, opened 2026-08-20
wolfsftp: report a failed remote write from SFTP put

yosuke-wolfssl:
### Problem

`wolfSSH_SFTP_Put()` reports a successful upload after the server rejects a write. When `wolfSSH_SFTP_SendWritePacket()` returns a non-positive size for a non-retryable reason — chiefly a server `FXP_STATUS != FX_OK`, which surfaces as `WS_SFTP_STATUS_NOT_OK` — the code only consulted `NoticeError()` and dropped the error. The `do/while (sz > 0)` loop then ended normally, and `STATE_PUT_CLOSE_REMOTE` overwrote `ret` with the successful close result, so `STATE_PUT_CLEANUP` returned `WS_SUCCESS`.

Trigger: open a remote file successfully, then have the server fail the write because its filesystem is full or read-only. The application is told the transfer succeeded while the remote file is truncated or empty. Closes f-10543.

### Fix (`src/wolfsftp.c`)

`STATE_PUT_WRITE` now records the failure and routes around the close that would mask it:

```c
WLOG(WS_LOG_SFTP, "Error writing packet");
if (ssh->error == WS_SUCCESS) {
    ssh->error = (sz < 0) ? sz : WS_FATAL_ERROR;
}
ret = WS_FATAL_ERROR;
/* no remote close, it would overwrite ret */
state->handleSz = 0;
state->state = STATE_PUT_CLOSE_LOCAL;
break;
```

- **`ssh->error`** is filled in only when unset, so a specific code such as `WS_SFTP_BAD_HEADER` survives; the return stays `WS_FATAL_ERROR` per convention.
- **`state->handleSz = 0`** makes `STATE_PUT_CLOSE_REMOTE` skip the close, the same mechanism the `STATE_PUT_OPEN_REMOTE` failure path already uses.
- **`continue` after the loop** reaches cleanup with `ret` intact.

This mirrors `wolfSSH_SFTP_Get()`, which likewise skips its remote close on a hard read failure. The trade-off is deliberate: the server-side handle stays open until session teardown, and a local `ret` cannot survive a `WS_WANT_READ` from the close, since it is re-initialised on every entry.

### Tests

`test_SftpClientPutWriteStatusFail()` reuses the existing in-memory client harness (`SftpClientNewSession()`, `wolfSSH_TestChannelPutData()`) to stage an `FXP_HANDLE` open reply and an `FXP_STATUS` failure for the only write, then asserts `WS_FATAL_ERROR`, `WS_SFTP_STATUS_NOT_OK`, and that `putState` was freed.

### Verification

- `make check`: 11 passed, 1 skipped, 0 failed.
- Negative control: the test fails with the `wolfsftp.c` change reverted.
- Preflight: gcc-13 `-Werror`, lint plus 6 configs clean.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -10130,6 +10130,15 @@ int wolfSSH_SFTP_Put(WOLFSSH* ssh, char* from, char* to, byte resume,
                         if (NoticeError(ssh)) {
                             return WS_FATAL_ERROR;
                         }
+                        WLOG(WS_LOG_SFTP, "Error writing packet");
+                        if (ssh->error == WS_SUCCESS) {
+                            ssh->error = (sz < 0) ? sz : WS_FATAL_ERROR;
+                        }
+                        ret = WS_FATAL_ERROR;
+                        /* no remote close, it would overwrite ret */
+                        state->handleSz = 0;
+                        state->state = STATE_PUT_CLOSE_LOCAL;
+                        break;
                     }
                     else {
                         AddAssign64(state->pOfst, sz);
@@ -10147,6 +10156,9 @@ int wolfSSH_SFTP_Put(WOLFSSH* ssh, char* from, char* to, byte resume,
                 if (ssh->sftpInt) {
                     wolfSSH_SFTP_SaveOfst(ssh, from, to, state->pOfst);
                     ssh->sftpInt = 0;
+                }
+                if (ret != WS_SUCCESS) {
+                    continue;
                 }
                 FALL_THROUGH;
 
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -94,6 +94,18 @@
 #define WOLFSSH_TEST_PREAD_PWRITE
 #endif
 
+/* The SFTP put regression drives a client session over a scratch file, so it
+ * needs mkstemp() and a mounted filesystem. */
+#if defined(WOLFSSH_SFTP) && !defined(NO_WOLFSSH_CLIENT) && \
+    !defined(NO_FILESYSTEM) && !defined(USE_WINDOWS_API) && \
+    !defined(WOLFSSL_NUCLEUS) && !defined(WOLFSSH_ZEPHYR) && \
+    !defined(FREESCALE_MQX) && !defined(MICROCHIP_MPLAB_HARMONY) && \
+    !defined(WOLFSSH_USER_FILESYSTEM) && !defined(WOLFSSH_FATFS)
+#include <stdlib.h>
+#include <unistd.h>
+#define WOLFSSH_TEST_SFTP_PUT
+#endif
+
 /* A port whose seek type cannot reach 4 GiB rejects the offset instead of
  * truncating it. Needs a mounted filesystem for the scratch file. */
 #if defined(WOLFSSH_ZEPHYR) && !defined(NO_FILESYSTEM) && \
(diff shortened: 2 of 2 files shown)


src/wolfsftp.c
```diff
@@ -10130,6 +10130,15 @@ int wolfSSH_SFTP_Put(WOLFSSH* ssh, char* from, char* to, byte resume,
                         if (NoticeError(ssh)) {
                             return WS_FATAL_ERROR;
                         }
+                        WLOG(WS_LOG_SFTP, "Error writing packet");
+                        if (ssh->error == WS_SUCCESS) {
+                            ssh->error = (sz < 0) ? sz : WS_FATAL_ERROR;
+                        }
+                        ret = WS_FATAL_ERROR;
+                        /* no remote close, it would overwrite ret */
+                        state->handleSz = 0;
```
yosuke-wolfssl, line 10139: Intentional, and the trade-off is the one the inline comment documents.

Preserving the error across the close needs storage in `WS_SFTP_PUT_STATE`: `ret` is a local re-initialised on every entry, so a `WS_WANT_READ` from `wolfSSH_SFTP_Close()` returns to the caller and the retry loses the error — a second local only fixes blocking sockets. No state struct in `wolfsftp.c` carries an error, so that member was rejected in review and the shape of `wolfSSH_SFTP_Get()` taken instead, which likewise skips its remote close on a hard read failure. The handle is released at session teardown.

yosuke-wolfssl, line 10139: Intentional, and the trade-off is the one the inline comment documents.

Preserving the error across the close needs storage in `WS_SFTP_PUT_STATE`: `ret` is a local re-initialised on every entry, so a `WS_WANT_READ` from `wolfSSH_SFTP_Close()` returns to the caller and the retry loses the error — a second local only fixes blocking sockets. No state struct in `wolfsftp.c` carries an error, so that member was rejected in review and the shape of `wolfSSH_SFTP_Get()` taken instead, which likewise skips its remote close on a hard read failure. The handle is released at session teardown.

tests/unit.c
```diff
@@ -15449,6 +15461,94 @@ static int test_SftpClientRecvInitVersion(void)
...
+
+
+/* An upload whose write is answered with a failure status must report that
+ * failure. The open reply hands out a handle and the write reply rejects the
+ * only write, leaving nothing else in the transfer able to fail. */
+static int test_SftpClientPutWriteStatusFail(void)
+{
+    WOLFSSH_CTX* ctx = NULL;
+    WOLFSSH*     ssh = NULL;
+    char   tmpFile[] = "/tmp/wolfssh_sftp_putXXXXXX";
```
yosuke-wolfssl, line 15486: Following the dominant convention in this file: literal `/tmp` templates at tests/unit.c:13597, 13756, 13874, 14048, 14188 and tests/api.c:2538, 4668. The sibling that honours `TMPDIR` does so for the reason its own comment gives — it can leave a 4 GiB sparse file behind. This test writes 10 bytes.

tests/unit.c
```diff
@@ -15449,6 +15461,94 @@ static int test_SftpClientRecvInitVersion(void)
...
+            PutU32BE(reply + WOLFSSH_SFTP_HEADER + UINT32_SZ, 0);
+            PutU32BE(reply + WOLFSSH_SFTP_HEADER + (UINT32_SZ * 2), 0);
+        }
+        if (replySz == 0 || wolfSSH_TestChannelPutData(ssh->channelList,
+                    reply, replySz) != WS_SUCCESS) {
+            result = -983;
+        }
+    }
+    if (result == 0) {
+        if (wolfSSH_SFTP_Put(ssh, tmpFile, remote, 0, NULL) != WS_FATAL_ERROR)
```
yosuke-wolfssl: Half addressed. Added a second case where the write reply carries a request id the client never sent, so `SendWritePacket` sets `WS_SFTP_BAD_REQ_ID` before returning and the `if (ssh->error == WS_SUCCESS)` guard must preserve it. Removing that guard now fails the test.

`sz == 0` looks unreachable: `SendWritePacket` returns `sentSzSave`, positive whenever `inSz > 0`, and the loop breaks earlier when `rSz == 0`. Kept as a defensive arm rather than tested.

src/wolfsftp.c
```diff
@@ -10144,6 +10153,9 @@ int wolfSSH_SFTP_Put(WOLFSSH* ssh, char* from, char* to, byte resume,
                     }
                 } while (sz > 0 && ssh->sftpInt == 0);
 
+                if (ret != WS_SUCCESS) {
+                    continue;
```
yosuke-wolfssl: Valid, thanks — fixed. The interrupt block now runs before the error `continue`, so the offset is saved and `ssh->sftpInt` cleared on the failure path as it was before this change.

Note `wolfSSH_SFTP_Get()` has the same latch on its own error path (src/wolfsftp.c:9903-9909); that one is pre-existing and untouched here.

src/wolfsftp.c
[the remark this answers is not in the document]
yosuke-wolfssl: Keeping the current shape. Recording the failure in `WS_SFTP_PUT_STATE` and letting the close run does work — it was the first option we weighed — but no state struct in `wolfsftp.c` carries an error today, and this follows `wolfSSH_SFTP_Get()`, which likewise skips its remote close on a hard read failure. Happy to switch to the state-field version if a reviewer prefers it.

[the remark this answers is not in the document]
yosuke-wolfssl: Keeping the current shape. Recording the failure in `WS_SFTP_PUT_STATE` and letting the close run does work — it was the first option we weighed — but no state struct in `wolfsftp.c` carries an error today, and this follows `wolfSSH_SFTP_Get()`, which likewise skips its remote close on a hard read failure. Happy to switch to the state-field version if a reviewer prefers it.

One correction: `ssh->fileList` is not unbounded. `SFTP_FileHandleCapped()` is checked before every open (src/wolfsftp.c:2401, 2632, 4105) and refuses past `WOLFSSH_MAX_SFTP_HANDLES` (64) with `FXP_FAILURE`, and the list is freed at session teardown. The effect is that a client looping over failing puts stops being able to open files after 64, not unbounded descriptor growth.

tests/unit.c
```diff
@@ -15449,6 +15461,132 @@ static int test_SftpClientRecvInitVersion(void)
...
+
+/* Drives wolfSSH_SFTP_Put() over a staged handle reply and a write answered by
+ * an FXP_STATUS carrying the request id "openId + writeIdOfst". Returns 0, or
+ * a negative sentinel on a setup failure. */
+static int SftpClientDrivePutFail(word32 writeIdOfst, int* putRet, int* putErr,
+        int* stateFreed)
+{
+    WOLFSSH_CTX* ctx = NULL;
+    WOLFSSH*     ssh = NULL;
+    char   tmpFile[] = "/tmp/wolfssh_sftp_putXXXXXX";
```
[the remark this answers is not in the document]
yosuke-wolfssl: Same as the previous round: this follows the dominant convention in the file — literal `/tmp` templates at tests/unit.c:13597, 13756, 13874, 14048, 14188 and tests/api.c:2538, 4667. The one test that resolves `TMPDIR` does so for the reason its own comment gives: it can leave a 4 GiB sparse file behind. This test writes 10 bytes.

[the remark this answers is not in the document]
yosuke-wolfssl: Same as the previous round: this follows the dominant convention in the file — literal `/tmp` templates at tests/unit.c:13597, 13756, 13874, 14048, 14188 and tests/api.c:2538, 4667. The one test that resolves `TMPDIR` does so for the reason its own comment gives: it can leave a 4 GiB sparse file behind. This test writes 10 bytes.

ejohnstown (approved):

PR #1188, opened 2026-08-20
wolfSSH_stream_read consumes before crediting and always reports the byte count.

yosuke-wolfssl:
### Problem

`ssh_worker()` in the echo server treats every positive `write()`, `send()`, and
`wolfSSH_ChannelIdSend()` return as a complete transfer. `SendChannelData()` clamps
each send to `min(peerWindowSz, peerMaxPacketSz, maxPacketSz)` and returns that
clamped count, so a send larger than the peer's window returns short. Every call
site then overwrites or discards its source buffer — the forwarding path resets
`fwdBufferIdx` to `0` on any positive result — and the unsent suffix is lost
silently. Reachable through the shell, agent, and forwarding paths.

Separately, `wolfSSH_stream_read()` credits `inputBuffer->idx` bytes but advanced
`idx` only afterwards, so a read never credits its own bytes. A single read that
drains the entire receive window leaves the window at zero with nothing left to
trigger a credit, wedging the stream permanently. Latent at the default 128 KB
window; immediate with a smaller one.

### Fix (`examples/echoserver/echoserver.c`)

Two mechanisms, split by which side owns the backpressure:

- **Local descriptors** (pty, agent and forwarding sockets) stall only until the fd
  drains, so `app_write_all()` advances past each partial transfer and returns
  `WS_FATAL_ERROR` if it cannot complete.
- **SSH channels** stall until the peer's window reopens, which needs the loop to
  process inbound packets. Each direction stages into its `WS_AppCtx` buffer; the
  flush subtracts what was sent and `WMEMMOVE`s the remainder for the next pass.

| Send result | Handling |
|---|---|
| `> 0` | subtract, keep the remainder |
| `WS_CHANNEL_NOT_CONF`, `WS_CHAN_RXD`, `WS_WINDOW_FULL`, `WS_REKEYING` | hold, retry on inbound data |
| `WS_WANT_WRITE` | hold, and add `sshFd` to a write set so `select()` wakes on writability |

A descriptor stays out of the read set while its buffer is pending. Closes f-10544.

### Fix (`src/ssh.c`)

`wolfSSH_stream_read()` consumes before crediting and always reports the byte count,
recording a failed adjust in `ssh->error` — the contract `_ChannelReadExt()` already
follows, so returning an error after consuming cannot drop the caller's bytes.

### Tests

- `test_wolfSSH_stream_read_WindowCredit()` (`tests/api.c`) round-trips 2000 bytes
  through a 1024-byte client receive window against the echo server.
- `test_stream_read_deferredWindowAdjust()` (`tests/unit.c`) reads a full window with
  an IO send that reports `WS_CBIO_ERR_WANT_WRITE`.

### Verification

- Full suite green: `unit`, `api`, `testsuite`, `regress`, `kex`, `auth`, and the
  `scripts/` tests including `fwd.test`.
- Clean under `-Werror` across six GCC configurations; ASan + UBSan clean.
- Both tests fail with either fix reverted; the unit test also fails against an
  intermediate variant that consumed the bytes but returned the error.

### DIFF
--- a/examples/echoserver/echoserver.c
+++ b/examples/echoserver/echoserver.c
@@ -109,12 +109,16 @@
     #define SOCKET_ECONNRESET ECONNRESET
     #define SOCKET_ECONNABORTED ECONNABORTED
     #define SOCKET_EWOULDBLOCK EWOULDBLOCK
+    #define SOCKET_EAGAIN EAGAIN
+    #define SOCKET_EINTR EINTR
 #else
     #include <WS2tcpip.h>
     #define SOCKET_ERRNO WSAGetLastError()
     #define SOCKET_ECONNRESET WSAECONNRESET
     #define SOCKET_ECONNABORTED WSAECONNABORTED
     #define SOCKET_EWOULDBLOCK WSAEWOULDBLOCK
+    #define SOCKET_EAGAIN WSAEWOULDBLOCK
+    #define SOCKET_EINTR WSAEINTR
 #endif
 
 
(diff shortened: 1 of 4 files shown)


src/ssh.c
```diff
@@ -1288,11 +1288,12 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
...
             WMEMCPY(buf, inputBuffer->buffer + inputBuffer->idx, n);
+            inputBuffer->idx += n;
             ret = _UpdateChannelWindow(ssh->channelList);
-            if (ret == WS_SUCCESS) {
-                inputBuffer->idx += n;
-                ret = n;
+            if (ret != WS_SUCCESS) {
+                ssh->error = ret;
             }
+            ret = n;
```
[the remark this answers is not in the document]
yosuke-wolfssl: ok

tests/unit.c
```diff
@@ -6063,6 +6063,137 @@ static int test_SendChannelData_zeroPeerMaxPacket(void)
...
+    word32           i;
+
+    for (i = 0; i < (word32)sizeof(in); i++) {
+        in[i] = (byte)i;
+    }
+
+    ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL);
+    if (ctx == NULL)
+        return -6980;
+    wolfSSH_SetIOSend(ctx, WantWriteIoSend);
```
yosuke-wolfssl, line 6088: Fixed in 45b5130a — both tests and both call sites are now inside
`#ifndef NO_WOLFSSH_SERVER`. Verified by building `--disable-server`, which
failed as reported and now passes `unit.test`.

examples/echoserver/echoserver.c
```diff
@@ -1001,43 +1081,42 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
+                                cnt_w = app_write_all(childFd,
+                                        threadCtx->channelBuffer, cnt_r, 0);
+                                if (cnt_w <= 0)
+                                    break;
                             }
+                            else
                             #endif
-                            if (cnt_w <= 0)
-                                break;
+                            if (shellBufferIdx == 0) {
```
yosuke-wolfssl: Fixed in 01249984. Introduced by this PR — master read unconditionally on every
`WS_CHAN_RXD`; staging the tail added the `shellBufferIdx == 0` guard.

The read now appends into the buffer's free space, a full buffer sets
`shellRxPending` instead of dropping the event, and a retry site before the
flush drains it. `WINCREDIT_PAYLOAD_SZ` raised to 16000 covers it: with the old
guard restored the test fails `4096 != 16000`.

examples/echoserver/echoserver.c
```diff
@@ -1244,26 +1355,34 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
-                    else if (cnt_w == WS_CHANNEL_NOT_CONF ||
-                            cnt_w == WS_CHAN_RXD) {
-                    #ifdef SHELL_DEBUG
-                        printf("Waiting for channel open confirmation.\n");
-                    #endif
-                    }
-                    else {
-                        break;
+            }
+            if (fwdBufferIdx > 0) {
```
yosuke-wolfssl: Fixed in 01249984 as recommended: `fwdBufferIdx` cleared in the
`WS_CHANNEL_CLOSED` handler, flush gated on `APP_STATE_CONNECTED`.
`scripts/fwd.test` passes.

examples/echoserver/echoserver.c
yosuke-wolfssl: Same defect and fix as comment 3840176262 on this line. Addressed in 01249984.

examples/echoserver/echoserver.c
```diff
@@ -975,17 +1054,18 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
-            if (fwdFd >= 0
+            if (fwdFd >= 0 && fwdBufferIdx == 0
                     && threadCtx->fwdCtx.state == APP_STATE_CONNECTED) {
                 FD_SET(fwdFd, &readFds);
                 if (fwdFd > maxFd)
                     maxFd = fwdFd;
             }
             #endif /* WOLFSSH_FWD */
-            rc = select((int)maxFd + 1, &readFds, NULL, NULL, NULL);
+            rc = select((int)maxFd + 1, &readFds, &writeFds, NULL, NULL);
```
yosuke-wolfssl, line 1065: Fixed in 01249984, with one correction: the echo path can't reach this.
`wolfSSH_CTX_SetWindowPacketSize()` sets the value a side both advertises and
sends with, so the staged amount never exceeds the clamp — measured, every
drain was `1024` with `0` left. The reachable direction is local-fd → channel,
where the read is bounded by `EXAMPLE_BUFFER_SZ`, not by the peer.

Each flush is now `do { ... } while (cnt_w > 0 && bufferIdx > 0)`: a packet-size
clamp returns short but positive and the loop continues, a window clamp returns
`WS_WINDOW_FULL` and parks in `select()`. No new state, no polling.

yosuke-wolfssl:
Rework

PR #1190, opened 2026-08-21
Shutdown channel lookup and disconnect state

ejohnstown:
Two independent connection-teardown fixes. A channel lookup that matched the wrong ID field, and a disconnect that left no lasting mark on the session. Both covered by unit.c / regress.c tests that fail without the change.

### wolfSSH_shutdown() channel lookup (F-8817)

- Searched by peerChannel while telling ChannelFind() to match the local ID field. The session channel is the head of the list, so shutdown takes it directly; ChannelAppend() appends to the tail, so the head stays the channel the teardown is for. The NULL result had been skipping the EOF, exit-status and close sends and the drain that waits on the peer's close.
- Only bites when the two IDs differ, which is why the single-channel tests, where both sides pick 0, never caught it.

### Terminal disconnect state (F-8837)

SSH_MSG_DISCONNECT set only ssh->error, which wolfSSH_stream_read() clears on entry. New WOLFSSH.disconnected, set by DoDisconnect() before the payload decode and by SendDisconnect() before the send, so a partial or failed send still ends the session. RFC 4253 section 11.1.

The invariant is: after a disconnect nothing more goes out, but what already arrived can still be read. It is documented in wolfssh/ssh.h and on the field in wolfssh/internal.h.

Nothing more goes out. One SendAfterDisconnect() helper gates every send entry point declared below that comment in ssh.h:

- wolfSSH_stream_send(), wolfSSH_stream_exit(), wolfSSH_ChannelIdSend(), wolfSSH_ChannelIdSendExt(), wolfSSH_extended_data_send(), wolfSSH_global_request()
- wolfSSH_TriggerKeyExchange(), wolfSSH_SendIgnore(), wolfSSH_SendDisconnect() -- emitting KEXINIT or MSG_IGNORE after a disconnect is what section 11.1 forbids, and one disconnect ends the session, so a second is refused. The last two needed a NULL check of their own, having relied on the callee for it; the return for a NULL ssh is unchanged at WS_BAD_ARGUMENT.
- wolfSSH_ChangeTerminalSize(), so the ssh.h contract is true for every send below it.

wolfSSH_shutdown() is a send path too, and is declared above that comment. It drops the channel when the flag is set, skipping the EOF, exit-status and close sends along with the wait for a close the peer will never send.

wsHighwater(), the default highwater callback, skips the rekey on a disconnected session. Without that, gating wolfSSH_TriggerKeyExchange() would have turned a firing high water mark into a failure for a packet that had gone out fine, since wolfSSH_SendPacket() returns HighwaterCheck()'s result.

wolfSSH_worker() stays ungated on purpose: wolfSSH_shutdown() and several examples pump it after an error, and it is the retry path for a short disconnect send.

What already arrived can still be read. The read calls are deliberately not gated: channel data that arrived before the disconnect was legitimately sent and is still the caller's. Gating wolfSSH_ChannelIdRead() and wolfSSH_extended_data_read() would discard it on exactly the paths wolfsshd uses to drain shell output. wolfSSH_stream_read() and wolfSSH_stream_peek() follow the invariant too: they drain their buffer first and report WS_DISCONNECT only once it runs dry, or straight away when the channel is gone and there is no buffer left.

### Tests

Every gate has a regression test in tests/regress.c, checked against the unfixed code. The send sweep asserts each entry point reports WS_DISCONNECT and that io.outSz never moves afterwards; separate tests cover the buffered-data drain, the shutdown teardown, the highwater callback and the terminal resize. tests/unit.c covers the shutdown channel lookup with a peer ID that differs from the local one.

The disconnect tests run on a session past user auth. Without that, IsMessageAllowed() blocks the sends on its own and the "nothing on the wire" assertions hold even with the gates removed.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1198,6 +1198,7 @@ int wolfSSH_TriggerKeyExchange(WOLFSSH* ssh)
 int wolfSSH_stream_peek(WOLFSSH* ssh, byte* buf, word32 bufSz)
 {
     WOLFSSH_BUFFER* inputBuffer;
+    word32 avail;
 
     WLOG(WS_LOG_DEBUG, "Entering wolfSSH_stream_peek()");
 
@@ -1214,11 +1215,22 @@ int wolfSSH_stream_peek(WOLFSSH* ssh, byte* buf, word32 bufSz)
     }
 
     inputBuffer = &ssh->channelList->inputBuffer;
-    bufSz = min(bufSz, inputBuffer->length - inputBuffer->idx);
+    avail = inputBuffer->length - inputBuffer->idx;
+
+    /* Report the disconnect only once the buffered data is drained, the
+     * same way wolfSSH_stream_read() does. Callers use this to tell a
+     * drained channel from one with more to come, and a dead session is
+     * neither. */
+    if (avail == 0 && ssh->disconnected) {
+        ssh->error = WS_DISCONNECT;
+        return WS_FATAL_ERROR;
+    }
+
+    bufSz = min(bufSz, avail);
     if (buf != NULL) {
         WMEMCPY(buf, inputBuffer->buffer + inputBuffer->idx, bufSz);
     }
-    return bufSz;
+    return (int)bufSz;
 }
 
 
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1257,7 +1257,10 @@ int wolfSSH_stream_peek(WOLFSSH* ssh, byte* buf, word32 bufSz)
         return WS_BAD_ARGUMENT;
     }
 
-    if (ssh->isKeying) {
+    /* A rekey the peer abandoned with a disconnect never completes, since
+     * only NEWKEYS clears the flag. Report the dead session instead, or the
+     * caller turns the crank forever. */
+    if (ssh->isKeying && !ssh->disconnected) {
         ssh->error = WS_REKEYING;
         return WS_REKEYING;
     }
@@ -1324,7 +1327,9 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
         return WS_ERROR;
     }
 
-    if (ssh->isKeying) {
+    /* See wolfSSH_stream_peek(): a disconnect ends a rekey that can no
+     * longer finish, so it outranks it here too. */
+    if (ssh->isKeying && !ssh->disconnected) {
         ssh->error = WS_REKEYING;
         return WS_FATAL_ERROR;
     }
@@ -1374,8 +1379,11 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
         }
     }
 
-    /* update internal input buffer based on data read */
-    if (ret == WS_SUCCESS && !ssh->isKeying) {
+    /* update internal input buffer based on data read. DoReceive() above may
+     * have started a rekey, which holds the copy back -- unless a disconnect
+     * came with it, since then the rekey never finishes and the buffered data
+     * would never be handed back. */
+    if (ret == WS_SUCCESS && (!ssh->isKeying || ssh->disconnected)) {
         int n;
 
         n = min(bufSz, inputBuffer->length - inputBuffer->idx);
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1113,9 +1113,9 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
     if (ssh == NULL || ssh->channelList == NULL)
         ret = WS_BAD_ARGUMENT;
 
-    /* look up the channel if it still exists */
+    /* The session channel is the head of the list. */
     if (ret == WS_SUCCESS) {
-        channel = ChannelFind(ssh, ssh->channelList->peerChannel, WS_CHANNEL_ID_SELF);
+        channel = ssh->channelList;
     }
 
     /* if channel close was not already sent then send it */
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1118,14 +1118,38 @@ static int SendAfterDisconnect(WOLFSSH* ssh)
 }
 
 
+/* A disconnect of ours left queued by a short send still has to reach the
+ * peer, and flushing bytes that are already bundled is not the new traffic
+ * RFC 4253 section 11.1 forbids. Only our own disconnect qualifies: a
+ * disconnect from the peer leaves nothing queued but unrelated traffic,
+ * which the session is over for. Call only after a NULL check of ssh. */
+static int FlushQueuedDisconnect(WOLFSSH* ssh)
+{
+    if (!ssh->disconnectTxd || !wolfSSH_OutputPending(ssh))
+        return 0;
+
+    WLOG(WS_LOG_DEBUG, "Flushing a disconnect left queued by a short send");
+    return 1;
+}
+
+
 int wolfSSH_shutdown(WOLFSSH* ssh)
 {
     int ret = WS_SUCCESS;
+    int flushRet = WS_SUCCESS;
     WOLFSSH_CHANNEL* channel = NULL;
 
     WLOG(WS_LOG_DEBUG, "Entering wolfSSH_shutdown()");
 
-    if (ssh == NULL || ssh->channelList == NULL)
+    if (ssh == NULL)
+        ret = WS_BAD_ARGUMENT;
+
+    /* This is a teardown call, so a disconnect of ours left queued by a
+     * short send goes out here, with or without a channel to tear down. */
+    if (ret == WS_SUCCESS && FlushQueuedDisconnect(ssh))
+        flushRet = wolfSSH_SendPacket(ssh);
+
+    if (ret == WS_SUCCESS && ssh->channelList == NULL)
         ret = WS_BAD_ARGUMENT;
 
     /* The session channel is the head of the list. */
@@ -1133,16 +1157,27 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
         channel = ssh->channelList;
     }
 
+    /* Session already over. Drop the channel to skip the teardown sends
+     * and the wait for a close that will not come. RFC 4253 section 11.1. */
+    if (channel != NULL && ssh->disconnected) {
+        WLOG(WS_LOG_DEBUG, "Session already disconnected, nothing to send");
+        /* An unfinished flush owns ssh->error. Callers gate their retry on
+         * WS_WANT_WRITE, so overwriting it strands the queued disconnect. */
+        if (flushRet == WS_SUCCESS)
+            ssh->error = WS_DISCONNECT;
+        channel = NULL;
+    }
+
     /* if channel close was not already sent then send it */
     if (channel != NULL && !channel->closeTxd) {
        if (ret == WS_SUCCESS) {
-           ret = SendChannelEof(ssh, ssh->channelList->peerChannel);
+           ret = SendChannelEof(ssh, channel->peerChannel);
        }
 
        /* continue on success and in case where queueing up send packets */
        if (ret == WS_SUCCESS ||
                (ret != WS_BAD_ARGUMENT && ssh->error == WS_WANT_WRITE)) {
-           ret = SendChannelExit(ssh, ssh->channelList->peerChannel,
+           ret = SendChannelExit(ssh, channel->peerChannel,
            #if defined(WOLFSSH_TERM) || defined(WOLFSSH_SHELL)
                ssh->exitStatus);
            #else
@@ -1153,7 +1188,7 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
        /* continue on success and in case where queueing up send packets */
        if (ret == WS_SUCCESS ||
                (ret != WS_BAD_ARGUMENT && ssh->error == WS_WANT_WRITE))
-           ret = SendChannelClose(ssh, ssh->channelList->peerChannel);
+           ret = SendChannelClose(ssh, channel->peerChannel);
     }
 
 
@@ -1171,6 +1206,11 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
         WLOG(WS_LOG_DEBUG, "channel list was already removed");
         ret = WS_CHANNEL_CLOSED;
     }
+
+    /* An unfinished flush outranks the channel status: the caller has to
+     * come back for the rest of the disconnect. */
+    if (flushRet != WS_SUCCESS)
+        ret = flushRet;
 
     WLOG(WS_LOG_DEBUG, "Leaving wolfSSH_shutdown(), ret = %d", ret);
     return ret;
@@ -1571,9 +1611,13 @@ int wolfSSH_SendDisconnect(WOLFSSH *ssh, word32 reason)
         return WS_BAD_ARGUMENT;
 
     /* One disconnect ends the session; a second is more traffic on a
-     * connection that is already over. */
-    if (SendAfterDisconnect(ssh))
+     * connection that is already over. A short send leaves the first one
+     * queued, though, so that retry goes through. */
+    if (SendAfterDisconnect(ssh)) {
+        if (FlushQueuedDisconnect(ssh))
+            return wolfSSH_SendPacket(ssh);
         return WS_FATAL_ERROR;
+    }
 
     return SendDisconnect(ssh, reason);
 }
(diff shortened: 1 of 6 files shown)


src/ssh.c
```diff
@@ -1228,7 +1229,15 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
...
+    if (ssh == NULL || buf == NULL || bufSz == 0)
+        return WS_BAD_ARGUMENT;
+
+    if (ssh->disconnected) {
+        ssh->error = WS_DISCONNECT;
+        return WS_FATAL_ERROR;
+    }
+
+    if (ssh->channelList == NULL)
         return WS_BAD_ARGUMENT;
```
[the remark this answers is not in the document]
ejohnstown: Good catch, and it was worse than a consistency gap. `wolfSSH_stream_read()`/`wolfSSH_stream_send()` are the client-side API -- `apps/wolfssh`, `examples/client`, `tpmcertserver`. `wolfsshd` never calls either one; its shell and SFTP loops run entirely on `wolfSSH_ChannelIdSend()`, `wolfSSH_ChannelIdRead()` and `wolfSSH_extended_data_send()`, and `echoserver`'s multi-channel path does the same. So as originally scoped the flag never reached the daemon's data path at all.

Fixed in the new commit. There is now one `SendAfterDisconnect()` helper and all six send entry points go through it: `wolfSSH_stream_send()`, `wolfSSH_stream_exit()`, `wolfSSH_ChannelIdSend()`, `wolfSSH_ChannelIdSendExt()`, `wolfSSH_extended_data_send()` and `wolfSSH_global_request()`.

I did not extend it to the read calls, though, and that part of the suggestion I think is wrong. RFC 4253 section 11.1 forbids accepting what the peer sends *after* the disconnect; channel data that arrived *before* it was legitimately sent and is still the caller's. Gating `wolfSSH_ChannelIdRead()` and `wolfSSH_extended_data_read()` would discard it on exactly the paths wolfsshd uses to drain shell output. The invariant is now "after a disconnect nothing more goes out, but what already arrived can still be read", documented in `wolfssh/ssh.h` and on the field in `wolfssh/internal.h`. `wolfSSH_stream_read()` follows it too: it drains its buffer first and reports `WS_DISCONNECT` only once that runs dry.

`wolfSSH_worker()` stays ungated on purpose, since `wolfSSH_shutdown()` and several examples still pump it after an error.

Two regression tests cover it, both checked against the unfixed code: one that buffered data survives the disconnect and is handed back before `WS_DISCONNECT`, and one that all six send calls refuse without a byte leaving the session. Without the new gates that second test caught `wolfSSH_ChannelIdSend()` returning `WS_WINDOW_FULL` -- it really was still trying to transmit.

wolfssh/ssh.h
```diff
@@ -558,6 +558,11 @@ WOLFSSH_API int wolfSSH_CTX_SetWindowPacketSize(WOLFSSH_CTX* ctx,
 WOLFSSH_API int wolfSSH_accept(WOLFSSH* ssh);
 WOLFSSH_API int wolfSSH_connect(WOLFSSH* ssh);
 WOLFSSH_API int wolfSSH_shutdown(WOLFSSH* ssh);
+/* A disconnect, sent or received, ends the session. Nothing more goes out:
+ * every send call below reports WS_DISCONNECT from then on. Reads are not
```
ejohnstown: Correct, and thanks for catching the overreach in my own comment. Fixed in `e2e820ac` by gating rather than by narrowing the sentence: emitting KEXINIT or MSG_IGNORE after a disconnect is what RFC 4253 section 11.1 forbids, so the comment was right and the code was wrong.

`wolfSSH_TriggerKeyExchange()` and `wolfSSH_SendIgnore()` now take the `SendAfterDisconnect()` gate. I also swept in `wolfSSH_SendDisconnect()`, which you did not name but is declared below the same comment and was equally ungated -- leaving it would have earned a repeat finding on the next scan. A second disconnect is now refused; one ends the session, and the retry path for a short send is `wolfSSH_worker()`, which stays ungated on purpose. The only in-tree callers (`examples/portfwd/portfwd.c` for SendIgnore, `examples/echoserver/echoserver.c:1628` for SendDisconnect) fire before the flag is set, so neither is affected.

`wolfSSH_TriggerKeyExchange()` turned out to be the more useful of the two: `src/internal.c:557` is the highwater callback, so gating it also stops a rekey being started on a session the peer has already ended.

`wolfSSH_SendIgnore()` and `wolfSSH_SendDisconnect()` needed a NULL check of their own before the gate, since both previously relied on the callee for it. The return for a NULL `ssh` is unchanged at `WS_BAD_ARGUMENT`.

The three calls joined the existing send sweep in `tests/regress.c`, which asserts every send entry point reports `WS_DISCONNECT` and that `io.outSz` never moves afterwards. Checked against the unfixed code: without the gates it catches `wolfSSH_TriggerKeyExchange()` returning `WS_REKEYING` where the test wants `WS_FATAL_ERROR` -- it really was still emitting KEXINIT.

Leaving #7512 alone, as you suggested. `wolfSSH_SendIgnore()` still discards the caller's payload and sends 128 zero bytes; that is a separate defect with a separate fix and it is not touched here.

src/ssh.c
```diff
@@ -1170,6 +1185,9 @@ int wolfSSH_TriggerKeyExchange(WOLFSSH* ssh)
     if (ssh == NULL)
         ret = WS_BAD_ARGUMENT;
 
+    if (ret == WS_SUCCESS && SendAfterDisconnect(ssh))
```
ejohnstown, line 1231: Correct. HighwaterCheck()'s return propagates through wolfSSH_SendPacket() (internal.c:4321). Fixed by gating inside wsHighwater(), your second option. Covered by TestHighwaterQuietAfterDisconnect.

wolfssh/ssh.h
ejohnstown: Correct. Fixed by gating wolfSSH_ChangeTerminalSize(); ungated it returned WS_SUCCESS and put 48 bytes of window-change on the wire. All ten send calls declared below the comment now carry the gate. Covered by TestTerminalResizeBlockedAfterDisconnect.

src/ssh.c
```diff
@@ -1113,9 +1128,9 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
     if (ssh == NULL || ssh->channelList == NULL)
         ret = WS_BAD_ARGUMENT;
 
-    /* look up the channel if it still exists */
+    /* The session channel is the head of the list. */
     if (ret == WS_SUCCESS) {
-        channel = ChannelFind(ssh, ssh->channelList->peerChannel, WS_CHANNEL_ID_SELF);
+        channel = ssh->channelList;
```
ejohnstown, line 1157: Correct. Fixed: wolfSSH_shutdown() drops the channel when ssh->disconnected is set, skipping the teardown sends and the drain. Ungated it put 72 bytes out and set eofTxd/closeTxd. Covered by TestShutdownQuietAfterDisconnect.

src/ssh.c
```diff
@@ -1244,6 +1300,13 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
     inputBuffer = &ssh->channelList->inputBuffer;
     ssh->error = WS_SUCCESS;
 
+    /* Hand back whatever arrived before the disconnect, then report it once
+     * the buffer runs dry rather than going back to a dead transport. */
+    if (ssh->disconnected && inputBuffer->length - inputBuffer->idx == 0) {
```
ejohnstown, line 1345: Correct. Verified: 24 bytes of CHANNEL_WINDOW_ADJUST go out on the drain, from `wolfSSH_ChannelIdRead()`, `wolfSSH_extended_data_read()`, and `wolfSSH_stream_read()` from its second read on -- the first credits zero, since `_UpdateChannelWindow()` runs before the index advances.

Fixed as you suggested: `ChannelCreditWindow()` parks the credit when `ssh->disconnected`, as it already does mid-rekey.

One correction: `_ChannelReadExt()` already returned `bufSz` regardless of the adjust result, so only the `_UpdateChannelWindow()` callers dropped the byte count.

Covered by `TestDisconnectQuietWindowAdjust`, checked against the unfixed code.

src/ssh.c
```diff
@@ -1485,6 +1578,15 @@ int wolfSSH_SendIgnore(WOLFSSH* ssh, const byte* buf, word32 bufSz)
 int wolfSSH_SendDisconnect(WOLFSSH *ssh, word32 reason)
 {
     WLOG(WS_LOG_DEBUG, "Entering wolfSSH_SendDisconnect");
+
+    if (ssh == NULL)
+        return WS_BAD_ARGUMENT;
+
+    /* One disconnect ends the session; a second is more traffic on a
+     * connection that is already over. */
+    if (SendAfterDisconnect(ssh))
```
ejohnstown: Correct, and a third path is blocked too: `wolfSSH_worker()` does not flush either. `DoReceive()` reports a want-read as `WS_FATAL_ERROR`, never as `WS_WANT_READ`, so worker's flush branch is unreachable on an idle socket -- and `wolfSSH_SendPacket()` is `WOLFSSH_LOCAL`, so no public call was left.

Measured with a send callback that blocks once: `SendDisconnect` -> `WS_WANT_WRITE` with 24 bytes queued and nothing on the wire; retry -> `WS_FATAL_ERROR`; `shutdown` -> `WS_SUCCESS`, still nothing on the wire.

Fixed close to your suggestion: bundled bytes are not the new traffic section 11.1 forbids, so `wolfSSH_SendDisconnect()` and `wolfSSH_shutdown()` both retry the queued packet. With nothing queued the gate refuses as before.

The worker gap itself is pre-existing and left alone here.

Covered by `TestQueuedDisconnectFlushes` and `TestShutdownFlushesQueuedDisconnect`, checked against the unfixed code.

src/ssh.c
```diff
@@ -1142,6 +1165,13 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
     }
 
 
+    /* A short send can leave the disconnect itself queued. Nothing new goes
+     * out, but what is already bundled still has to reach the peer. */
+    if (channel == NULL && ret == WS_SUCCESS && ssh->disconnected &&
```
ejohnstown: Correct. Verified: with 32 bytes of channel data left queued by a short send, a peer disconnect made `wolfSSH_shutdown()` return `WS_SUCCESS` and put `CHANNEL_DATA` on the wire (msgid `0x5e`). My own invariant, broken by my own gate.

Fixed as you suggested: `SendDisconnect()` sets `ssh->disconnectTxd` once the packet is bundled, and one `FlushQueuedDisconnect()` helper gates both retries on that rather than on `ssh->disconnected`.

While in there, the flush now runs ahead of the channel-list test, so the peer's close retiring the last channel no longer strands the disconnect.

Covered by `TestPeerDisconnectKeepsTrafficQueued` and `TestShutdownFlushesWithNoChannel`, checked against the unfixed code.

src/ssh.c
```diff
@@ -1485,6 +1585,20 @@ int wolfSSH_SendIgnore(WOLFSSH* ssh, const byte* buf, word32 bufSz)
...
+    if (ssh == NULL)
+        return WS_BAD_ARGUMENT;
+
+    /* One disconnect ends the session; a second is more traffic on a
+     * connection that is already over. A short send leaves the first one
+     * queued, though, and flushing bytes that are already bundled is not new
+     * traffic, so the retry goes through. */
+    if (SendAfterDisconnect(ssh)) {
+        if (wolfSSH_OutputPending(ssh))
+            return wolfSSH_SendPacket(ssh);
```
ejohnstown, line 1638: Correct, the same defect at this site, and the `WS_SUCCESS` is the worse half: the caller is told a disconnect was sent when none was ever built.

Fixed with the flag you proposed. `wolfSSH_SendDisconnect()` now retries only output that `SendDisconnect()` itself bundled; after a peer-initiated disconnect it reports `WS_DISCONNECT` and leaves the queued traffic where it is.

Covered by `TestPeerDisconnectKeepsTrafficQueued`, checked against the unfixed code: ungated it returns `WS_SUCCESS` with 56 bytes out, first msgid `0x5e`.

src/ssh.c
```diff
@@ -1103,31 +1103,78 @@ int wolfSSH_connect(WOLFSSH* ssh)
...
     if (ret == WS_SUCCESS) {
-        channel = ChannelFind(ssh, ssh->channelList->peerChannel, WS_CHANNEL_ID_SELF);
+        channel = ssh->channelList;
+    }
+
+    /* Session already over. Drop the channel to skip the teardown sends
+     * and the wait for a close that will not come. RFC 4253 section 11.1. */
+    if (channel != NULL && ssh->disconnected) {
+        WLOG(WS_LOG_DEBUG, "Session already disconnected, nothing to send");
+        ssh->error = WS_DISCONNECT;
```
philljj: seems like this is a false positive, because your comment documents why you're doing this

ejohnstown: Valid, and fixed as suggested: `ssh->error` is set to `WS_DISCONNECT` only when the flush finished.

Confirmed before fixing. With the flush inside `wolfSSH_shutdown()` itself short-sending, it returned `WS_WANT_WRITE` (-1011) but reported `WS_DISCONNECT` (-1098), so the retry loops in `echoserver.c` and `sftpclient.c` were skipped. A regression test covers it; against the unfixed code it fails with `-1098 != -1011`.

One correction to the impact. The fix alone does not get the queued DISCONNECT to the peer for either caller cited. Both retry with `wolfSSH_worker()`, and on an idle socket `DoReceive()` reports want-read as `WS_FATAL_ERROR`, which misses the flush gate in `wolfSSH_worker()` (`WS_SUCCESS`/`WS_WANT_READ`/`WS_CHAN_RXD`), so the pending output is never attempted:

```
retry 0..9: ret=-1001 err=-1010 outSz=0 pending=1
```

Retrying with `wolfSSH_shutdown()` does flush it (24 bytes, msgid 1), and that is what the new test asserts. The worker gate is a pre-existing gap rather than something this PR introduced, so it stays out of scope here.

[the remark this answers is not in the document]
ejohnstown: Duplicate of #discussion_r3858807995, rescanned against pre-fix code. Already fixed on the branch: `ssh->error` is assigned only when `flushRet == WS_SUCCESS`, with a regression test.

src/ssh.c
```diff
@@ -1183,11 +1241,21 @@ int wolfSSH_TriggerKeyExchange(WOLFSSH* ssh)
...
+        /* No channel left to drain, so the disconnect is all there is. */
+        if (ssh->disconnected) {
+            ssh->error = WS_DISCONNECT;
+            return WS_FATAL_ERROR;
+        }
         return WS_BAD_ARGUMENT;
+    }
 
     if (ssh->isKeying) {
         ssh->error = WS_REKEYING;
```
philljj, line 1267: dumb question: if ssh->disconnected && ssh->isKeying is true, will this short circuit here and never advance state? And then get stuck spinning in SFTP_Subsystem()

```
            ret = wolfSSH_stream_peek(ssh, peek_buf, sizeof(peek_buf));         
            if (ret > 0) {                                                      
                ....                                                   
            }                                                                   
            else if (ret == WS_REKEYING) {                                      
                timeout = TEST_SFTP_TIMEOUT;                                    
                continue;                                                       
            }
```

philljj (changes requested):

src/ssh.c
ejohnstown: Confirmed, and fixed in `9e845f61`.

DISCONNECT is transport layer generic, so `IsMessageAllowed()` passes it
mid-KEX (`internal.c:715`, `:828`), and only NEWKEYS clears `isKeying`
(`:7767`, `:16212`) -- which never arrives. `stream_peek()` returns -1035,
`echoserver.c:1480` answers that with `continue`, and the loop polls a
one-second select for the life of the connection.

Three gates, not one. Against the unfixed code:

| site | result |
|---|---|
| `stream_peek()`, the line you flagged | -1035 |
| `stream_read()`, same early gate | -1001, error `WS_REKEYING` |
| `stream_read()`, the copy step | `WS_SUCCESS`, 0 bytes, no error |

All three now defer to the disconnect (`ssh.c:1266`, `:1335`, `:1389`).
The copy step still holds back a rekey `DoReceive()` started in the same
call. `ChannelCreditWindow()` already parks the credit, so the drain stays
off the wire. Covered by `TestDisconnectOutranksRekey`.

`wolfSSH_ChannelRead()` has the same gate but reports no disconnect at
all, so there is nothing to outrank; left alone.

A received CHANNEL_EOF masks the drain the same way. Pre-existing, going
to #1195 with the EOF lifecycle work; `ssh.h` says so now.

philljj (approved):

PR #1191, opened 2026-08-24
wolfsftp: keep the remote file when resuming a put

yosuke-wolfssl:
### Problem
Two defects, both leaving a corrupted file at the destination.

**Client** (`wolfSSH_SFTP_Put`, f-11659): resume mode restores the saved offset
and seeks the local source forward, but still opens the destination with
`WOLFSSH_FXF_TRUNC`. Truncation is the open's job (draft-ietf-secsh-filexfer-02
section 6.3), so the prefix already uploaded is discarded; the first write then
lands at the nonzero offset, which section 6.4 defines as zero-filling the gap.
A `reput` yields a hole followed by the tail, with no error returned.

**Windows server** (surfaced by the new test): `creationDisp |= ...` combined
`dwCreationDisposition` values that `CreateFileA` documents as mutually
exclusive. `WRITE|CREAT` became `CREATE_ALWAYS` and truncated a destination the
client asked to keep; `READ|WRITE|CREAT[|TRUNC]` became
`OPEN_EXISTING | CREATE_ALWAYS == OPEN_EXISTING`, so it failed when the file was
absent and never truncated when asked to. The POSIX server maps each flag
separately and is unaffected.

### Fix (`src/wolfsftp.c`)
`TRUNC` is added only when the write offset is zero. Dropping it alone would
still hole a missing or short destination, and would newly allow stale trailing
bytes on a longer one, so a new `STATE_PUT_STAT_REMOTE` validates first - the
same shape as OpenSSH's `sftp_upload()` resume branch, which stats the
destination and never sends `TRUNC` on a resume. It runs before the local open,
where the offset is consumed (`WFSEEK`, or the `OVERLAPPED` seed on Windows),
and does nothing unless the saved offset is nonzero, so a plain `put` costs no
extra round trip.

| `wolfSSH_SFTP_STAT()` on the destination | Action |
|---|---|
| size equals the saved offset | resume, no `TRUNC` |
| size differs, or no size attribute | restart from 0 with `TRUNC` |
| `WS_SFTP_STATUS_NOT_OK`, `WS_PERMISSIONS` | restart from 0 with `TRUNC` |
| `WS_WANT_READ` / `WS_WANT_WRITE` | retry in place |
| any other failure | re-save the offset, cleanup, error propagates |

`STATE_PUT_LOOKUP_OFFSET` also drops the offset when the local source is no
larger than it - OpenSSH's "destination file same size or larger" case - which
otherwise resumes into a zero-byte transfer that reports success. It uses the
`SFTP_GetAttributes()` available on every port, so no extra round trip.

The Windows disposition is now assigned, not OR'd:

```c
if (reason & WOLFSSH_FXF_TRUNC)
    creationDisp = CREATE_ALWAYS;
else
    creationDisp = OPEN_ALWAYS;
```

`WRITE|CREAT|APPEND`, which OpenSSH's `sftp` sends for `reput`, now maps to
`OPEN_ALWAYS`, so a Windows wolfSSH server no longer destroys the destination
for an OpenSSH resume.

`WS_SFTP_PUT_STATE` carries the returned attributes; it is defined only in
`wolfsftp.c`, so there is no header or ABI change.

### Tests (`tests/api.c`)
`test_wolfSSH_SFTP_PutResume()` is the first coverage of the resume path;
`scripts/get-put.test` already drives a plain put. Five cases, each checking the
destination byte-for-byte: resume onto a matching prefix, resume with the
destination deleted, resume with an offset that undershoots it, a plain put over
a longer destination, and a resume whose source is shorter than the offset. The
resume case stages a prefix unlike the source and expects it back untouched, so
a silent full re-upload fails the test. Excluded on Zephyr, which lacks the
hosted wrappers needed to stage the local source.

### Verification
- gcc-13 `-Werror` sweep over 6 configs (default, sftp-only, scp-only,
  enable-all, smallstack, Zephyr defines): clean, no warnings.
- `make check`, macOS `--enable-all`: 11 pass, 1 skip, 0 fail
  (`scripts/sftp.test` and `scripts/scp.test` need the serial re-run; their
  parallel failure is a harness `counter` bug, unrelated to this change).
- Negative control: with the offset-validation hunk reverted, the new
  source-too-short case fails.

### Not in this PR
- `SFTP_STAT()` does not surface the SSH_FX status, so `WS_SFTP_STATUS_NOT_OK`
  also covers `SSH_FX_FAILURE` and `SSH_FX_OP_UNSUPPORTED`. The restart branch
  is commented to say so; distinguishing them needs a new `SFTP_STAT()` return.
- Windows `SSH_FXF_EXCL` and `SSH_FXF_APPEND` are still unmapped (pre-existing,
  inside the same `#if 0`).
- The new state's want-read re-entry is not covered in-tree; the test client
  uses a blocking socket. Likewise the stat's permission-denied restart and its
  hard-error abort, both of which need a server that can fail a stat on demand.

### DIFF
--- a/tests/api.c
+++ b/tests/api.c
@@ -5680,6 +5680,270 @@ static void test_wolfSSH_SFTP_SaveOfst(void)
     wolfSSH_CTX_free(ctx);
 }
 
+
+#if !defined(NO_FILESYSTEM) && !defined(WOLFSSH_USER_FILESYSTEM) && \
+    !defined(WOLFSSH_ZEPHYR)
+
+#define SFTP_PUT_RESUME_SZ   1024
+#define SFTP_PUT_RESUME_OFST 256
+
+/* Fills buf with the source file's byte pattern. */
+static void sftpPutFillPattern(byte* buf, word32 sz)
+{
+    word32 i;
+
+    for (i = 0; i < sz; i++) {
+        buf[i] = (byte)(i * 7 + 1);
+    }
+}
+
+
+/* Writes sz bytes of buf to the local file name. Returns 0 on success. */
+static int sftpPutWriteFile(const char* name, const byte* buf, word32 sz)
+{
+    WFILE* fp = NULL;
+    int ret = 0;
+
+    if (WFOPEN(NULL, &fp, name, "wb") != 0 || fp == NULL)
+        return -1;
+
+    if (WFWRITE(NULL, buf, 1, sz, fp) != sz)
+        ret = -1;
+
+    WFCLOSE(NULL, fp);
+    return ret;
+}
+
+
+/* Returns 0 if name holds exactly sz bytes equal to expect. Asking for one
+ * byte more than expected also catches a file left too long. */
+static int sftpPutFileMatches(const char* name, const byte* expect, word32 sz)
+{
+    WFILE* fp = NULL;
+    byte got[SFTP_PUT_RESUME_SZ + 1];
+    int ret = 0;
+
+    if (sz >= sizeof(got))
+        return -1;
+
+    if (WFOPEN(NULL, &fp, name, "rb") != 0 || fp == NULL)
+        return -1;
+
+    if (WFREAD(NULL, got, 1, sz + 1, fp) != sz)
+        ret = -1;
+
+    if (ret == 0 && WMEMCMP(got, expect, sz) != 0)
+        ret = -1;
+
+    WFCLOSE(NULL, fp);
+    return ret;
+}
+
+
+/* Leaves sz bytes of buf in the remote file name, the way an interrupted
+ * upload would. Returns WS_SUCCESS on success. */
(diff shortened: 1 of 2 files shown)


tests/api.c
```diff
@@ -5680,6 +5680,252 @@ static void test_wolfSSH_SFTP_SaveOfst(void)
...
+    AssertNotNull(ssh);
+
+    sftpPutFillPattern(src, (word32)sizeof(src));
+    WMEMSET(stale, 0xFF, sizeof(stale));
+    AssertIntEQ(sftpPutWriteFile(srcName, src, (word32)sizeof(src)), 0);
+
+    /* the saved offset matches the destination, so the upload picks up where
+     * it left off and the staged prefix must survive */
+    (void)sftp_retry_remove(ssh, dstName); /* normally absent */
+    AssertIntEQ(sftpPutStageRemote(ssh, dstName, src, SFTP_PUT_RESUME_OFST),
```

ejohnstown (approved):

PR #1192, opened 2026-08-24
ssh.c: report the byte count from a channel read that defers its credit

yosuke-wolfssl:
### Problem

Two defects in the same three lines of `src/ssh.c`.

**1. A read that copied bytes reports failure.** `_ChannelRead()` — behind
`wolfSSH_ChannelRead()` and `wolfSSH_ChannelIdRead()` — consumes the bytes, then
returns the window-adjust send result:

```c
inputBuffer->idx += bufSz;                    /* bytes consumed */
updateResult = _UpdateChannelWindow(channel);
if (updateResult == WS_SUCCESS)
    updateResult = bufSz;
return updateResult;                          /* WS_WANT_WRITE -> "failed" */
```

`WS_WANT_WRITE` is routine on a non-blocking socket, not an error. A caller using the
usual `if (cnt_r <= 0) break;` shape tears the session down after the data has already
left the input buffer (`src/wolfscp.c`, `apps/wolfsshd/wolfsshd.c`). A hard transport
failure during the adjust was invisible on both paths.

**2. A channel that never gets its window back.** `wolfSSH_stream_read()` credited
before advancing `inputBuffer->idx`, so it credited the *previous* read's bytes. On a
session's first read `idx` is 0 and no `SSH_MSG_CHANNEL_WINDOW_ADJUST` is sent at all;
if that read drained the window, neither side can move.

### Fix (`src/ssh.c`)

Both paths report the bytes copied and keep the send result in `ssh->error`, matching
`_ChannelReadExt()`. `wolfSSH_stream_read()` advances `idx` before crediting, which is
also the fix for defect 2. `_ChannelRead()` retires a stale `WS_WANT_WRITE` only when
this read's own credit went out, and gains its sibling's `idx > length` guard and
`savedError` restore.

**`src/wolfsftp.c`:** `wolfSSH_SFTP_Close()`'s `STATE_CLOSE_SEND` arm was the file's
only ungated `NoticeError()` check, so a parked `WS_WANT_WRITE` made it report a
failed close. Now gated like its siblings.

### API compatibility

The three read entry points now return the byte count where a deferred or failed
adjust produced a negative return; the status moves to `wolfSSH_get_error()`.
`wolfSSH_ChannelReadExt()` already had this shape (`50ee1b61`). Documented on all
three prototypes. **Needs a `ChangeLog.md` line at release prep.**

### Tests (`tests/unit.c`)

One harness per path: a deferred adjust, then a hard failure, then a clean credit. The
`ChannelIdRead` harness adds an empty read, which must leave a seeded `WS_WANT_WRITE`
standing.

### Verification

`make check` passes, and `unit.test` is clean under ASan + UBSan and under `-Werror`
with gcc-13 in 6 configurations. Negative controls: reverting the `ssh->error`
recording fails the hard-failure assertions, restoring the old ordering sends 0
adjusts where the fix sends 1, and removing the retire gate lets an empty read clear a
`WS_WANT_WRITE`. (`scripts/sftp.test` and `scripts/scp.test` need a serial run on a
loaded machine — a pre-existing 2 second ready-file timeout, unrelated to this change.)

### Not in this PR

`wolfSSH_worker()` does not flush a parked adjust when its receive is idle, so a
read-only application whose peer has exhausted its window can stall — and there is no
public flush, since `wolfSSH_SendPacket()` and `wolfSSH_OutputPending()` are
`WOLFSSH_LOCAL`. Closing it means restructuring `wolfSSH_worker()`, which is
caller-visible; that is on its own branch. This PR neither creates nor widens the gap,
and its header no longer promises a flush that does not happen.

### DIFF
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -6488,6 +6488,220 @@ done:
     return result;
 }
 
+#ifndef NO_WOLFSSH_SERVER
+
+/* wolfSSH_stream_read() counterpart of test_ChannelExtDataCreditWantWrite():
+ * a deferred credit must not cost the caller the bytes already consumed, and
+ * an adjust that fails outright must still surface on wolfSSH_get_error(). */
+static int test_stream_read_deferredWindowAdjust(void)
+{
+    WOLFSSH_CTX*     ctx = NULL;
+    WOLFSSH*         ssh = NULL;
+    WOLFSSH_CHANNEL* ch  = NULL;
+    int              result = 0;
+    int              ret;
+    byte             in[64];
+    byte             out[64];
+    word32           i;
+
+    for (i = 0; i < (word32)sizeof(in); i++) {
+        in[i] = (byte)i;
+    }
+
+    ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_SERVER, NULL);
+    if (ctx == NULL)
+        return -6980;
+    wolfSSH_SetIOSend(ctx, WantWriteIoSend);
+
+    ssh = wolfSSH_new(ctx);
+    if (ssh == NULL) { result = -6981; goto done; }
+    /* Allow MSGID_CHANNEL_WINDOW_ADJUST on this bare session. */
+    ssh->acceptState = ACCEPT_SERVER_USERAUTH_SENT;
+
+    /* A window the size of the payload, so draining it in one read leaves
+     * windowSz at zero and _UpdateChannelWindow() has to credit. */
+    ch = ChannelNew(ssh, ID_CHANTYPE_SESSION,
+                    (word32)sizeof(in), DEFAULT_MAX_PACKET_SZ);
+    if (ch == NULL) { result = -6982; goto done; }
+    if (ChannelAppend(ssh, ch) != WS_SUCCESS) {
+        ChannelDelete(ch, ssh->ctx->heap);
+        result = -6983;
+        goto done;
+    }
+    ch->openConfirmed = 1;
+
+    if (wolfSSH_TestChannelPutData(ch, in, (word32)sizeof(in)) != WS_SUCCESS) {
+        result = -6984; goto done;
+    }
+    if (ch->windowSz != 0) { result = -6985; goto done; }
+
+    /* Every byte is reported, and they are the bytes that were put. */
+    ret = wolfSSH_stream_read(ssh, out, (word32)sizeof(out));
+    if (ret != (int)sizeof(in)) { result = -6990; goto done; }
+    if (WMEMCMP(out, in, sizeof(in)) != 0) { result = -6991; goto done; }
+
+    /* The deferral is observable, the window is credited locally, and the
+     * bytes are consumed rather than left for a re-read. */
+    if (ssh->error != WS_WANT_WRITE) { result = -6992; goto done; }
+    if (ch->windowSz != (word32)sizeof(in)) { result = -6993; goto done; }
+    if (ch->inputBuffer.length - ch->inputBuffer.idx != 0) {
+        result = -6994; goto done;
+    }
+
+    /* A peer that reset rather than blocked. wolfSSH_SendPacket() records only
+     * WS_WANT_WRITE, so the read path has to record a hard failure itself. */
+    wolfSSH_SetIOSend(ctx, FailIoSend);
+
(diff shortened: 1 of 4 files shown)


tests/unit.c
```diff
@@ -6063,6 +6063,141 @@ static int test_SendChannelData_zeroPeerMaxPacket(void)
...
+    if (ch->windowSz != 0) { result = -6985; goto done; }
+
+    /* Every byte is reported, and they are the bytes that were put. */
+    ret = wolfSSH_stream_read(ssh, out, (word32)sizeof(out));
+    if (ret != (int)sizeof(in)) { result = -6990; goto done; }
+    if (WMEMCMP(out, in, sizeof(in)) != 0) { result = -6991; goto done; }
+
+    /* The deferral is observable, the window is credited locally, and the
+     * bytes are consumed rather than left for a re-read. */
+    if (ssh->error != WS_WANT_WRITE) { result = -6992; goto done; }
```
yosuke-wolfssl, line 6545: Confirmed - fixed in the latest push. `wolfSSH_SendPacket()` does set `ssh->error = WS_WANT_WRITE` itself at `src/internal.c:4289`, so the `WantWriteIoSend` phase proved nothing about the added `ssh->error` write.

`test_stream_read_deferredWindowAdjust()` now has a second phase that swaps in `FailIoSend` (`WS_CBIO_ERR_GENERAL` -> `WS_SOCKET_ERROR_E`, which `SendPacket()` never records), puts another full window, and asserts the byte count is still returned, `ssh->error == WS_SOCKET_ERROR_E`, and the credit stays parked in `pendingWindowAdjust` - mirroring `test_ChannelReadExtHardFailureReported()`.

Negative control: stripping the two `ssh->error =` lines from `src/ssh.c` fails this test at `-6998` and the `ChannelIdRead` one at `-7024`, while both `WS_WANT_WRITE` phases still pass - exactly as you predicted.

tests/unit.c
```diff
@@ -6063,6 +6063,141 @@ static int test_SendChannelData_zeroPeerMaxPacket(void)
...
+    ssh->error = WS_SUCCESS;
+
+    /* Every byte is reported, and they are the bytes that were put. */
+    ret = wolfSSH_ChannelIdRead(ssh, ch->channel, out, (word32)sizeof(out));
+    if (ret != (int)sizeof(in)) { result = -7016; goto done; }
+    if (WMEMCMP(out, in, sizeof(in)) != 0) { result = -7017; goto done; }
+
+    /* The deferral is observable, the window is credited locally, and the
+     * bytes are consumed rather than left for a re-read. */
+    if (ssh->error != WS_WANT_WRITE) { result = -7018; goto done; }
```
yosuke-wolfssl, line 6649: Confirmed - fixed. `test_ChannelIdRead_deferredWindowAdjust()` gained the same `FailIoSend` phase, asserting a positive byte count with `ssh->error == WS_SOCKET_ERROR_E` and the credit left owed.

A follow-up review turned up a related gap on this path: `_ChannelRead()` recorded failures but never retired a stale `WS_WANT_WRITE`, and unlike `wolfSSH_stream_read()` these entry points do not reset `ssh->error` on entry - so a credit that later sent cleanly left `wolfSSH_get_error()` reporting a flush that was already done. `_ChannelRead()` now clears it the way `_ChannelReadExt()` does, covered by a third phase in this test.

ejohnstown (commented):
The flush advice added to three public prototypes in wolfssh/ssh.h is wrong in the case that matters. 
<details>
<summary>Details</summary>

Proved with a harness against libwolfssh_test.a: after a deferred adjust,  five wolfSSH_worker() calls on an idle socket produce zero IO sends and leave the 24-byte WINDOW_ADJUST sitting in the output buffer, because GetInputData() returns WS_FATAL_ERROR with WS_WANT_READ and that is outside worker’s flush gate. That is exactly the state stalled channel is in. The alternative remedy is unreachable: wolfSSH_SendPacket and wolfSSH_OutputPending are WOLFSSH_LOCAL, so a read-only application has no public flush at all. Not created by this PR, but this PR promotes the wrong advice into the public API documentation. (This is the same underlying gap as the existing worker-idle-flush note.)

</details>

ejohnstown (changes requested):

ejohnstown:
One other thing. Make the PR title match your one commit's title. Thanks!

ejohnstown (approved):

PR #1195, opened 2026-08-24
channel: half-close on the peer's EOF, add a public send-EOF API

ejohnstown:
`DoChannelEof()` answered a peer's half-close with an EOF of its own. That latched `eofTxd`, after which the "Cannot send data after EOF" gates in `SendChannelData()` and `SendChannelExtendedData()` refused every later send, so a peer that half-closed could never be replied to. `ssh host 'sort' < file` returns nothing on master today: wolfSSHd closes the child's stdin correctly and then cannot send the command's output back. RFC 4254 section 5.3 leaves that reply to the application, and no application could make it -- every `SendChannelEof()` caller bundled the EOF with the teardown, so there was no way to close one direction and keep reading.

### The library

- `DoChannelEof()` latches `eofRxd` and reports `WS_EOF` instead of echoing.  `wolfSSH_worker()` carries the status out with the channel it belongs to and does not mask it during a rekey, and `DoReceiveHandshake()` absorbs it in the three accept/connect states that can run with a channel already open: a legal EOF is not a handshake failure (F-1687)
- `wolfSSH_stream_read()` and `wolfSSH_stream_peek()` hand back what is buffered before they report the EOF. The `eofRxd` test moved behind the buffered-data test, so a half-close no longer strands what arrived with it.  `stream_read()` takes the head channel id before `DoReceive()` and keeps waiting when the EOF belongs to another channel, and loops only while the head is unchanged, since `DoChannelClose()` can free it
- `wolfSSH_ChannelSendEof()` and `wolfSSH_stream_send_eof()` are the missing half of the RFC's half-close: they close the sending direction and leave the receiving direction open. Both refuse a NULL session, a channel whose open is unconfirmed, a disconnected session and a packet between KEXINIT and NEWKEYS
- `SendChannelEof()` commits `eofTxd` once the EOF is in the output buffer, not only once the flush reports success, so the retry an application makes on `WS_WANT_WRITE` does not queue a second EOF behind the first. The one exception is the send failure that discards the buffer (F-8826)
- `SendChannelClose()` drops `closeTxd` on that same discard. A `closeTxd` claiming a close that never left made `wolfSSH_shutdown()`'s gate skip the teardown. It is a separate commit from the one above on purpose: that latch widens and this one narrows, and read as one change they read wrong
- `wolfSSH_ChannelExit()` leaves the channel on the list once it has sent the EOF and the close, so the application's pointer stays valid until the peer answers and the worker reports `WS_CHANNEL_CLOSED`. Removing it locally freed the pointer under the caller and left the peer's close matching nothing (F-8839)
- `DoChannelClose()` and `wolfSSH_ChannelExit()` answer only a channel whose open the peer has confirmed. `peerChannel` is 0 until then and both senders resolve by peer id, so a teardown aimed at an unconfirmed channel landed on whichever channel held peer id 0 -- normally the live session
- SFTP and SCP carry the status rather than aborting on it. A peer that closed its sending direction has not stopped a transfer that is still able to finish; `wolfSSH_SFTP_buffer_read()` also stops spending a receive on a peek that has already reported a hang-up

### The callers

The last commit is required by the library change, not tidying. All three worker ladders in wolfSSHd and the echoservers end in `else if (rc != WS_WANT_READ) break`, and wolfSSHd's reaches `kill(childPid, SIGKILL)`, so landing the library change alone ships green and kills the command on every half-close. wolfSSHd now closes the child's stdin off the channel's own EOF state; the echoservers hand back the backlog before answering with an EOF of their own; the clients treat a half-close as the graceful case; `portfwd` relays it to the local socket with `shutdown(SHUT_WR)` and is the working demonstration of the new API.

### Tests

`tests/unit.c` covers the half-close end to end, the channel id an EOF is reported against, the rekey case, the `channelEofCb` callback -- which had no test anywhere in the tree -- a handshake that survives an EOF, both latch arms through `FailIoSend`, and the unconfirmed-channel refusal on both send-EOF entry points. `apps/wolfsshd/test/sshd_stdin_eof_test.sh` is the end-to-end case against a live daemon; it cannot pass on master. The shared receive mock and packet builders are hoisted ahead of both endpoint regions first, in a relocation-only commit that reads with `--color-moved`.

### Merging

This branch is a linear stack. GitHub shows 13 commits and 19 files because the first three are #1211 and #1212; the last ten are this PR. #1212 must land first, since the wolfSSHd changes here are written on top of its drain. #1211 is preferred first as well -- the only code coupling is one adjacency in `src/ssh.c`, but landing this first would briefly double the post-disconnect chatter that #1211 removes. Merge order: #1211, #1212, this, then #1148.

The four commits behind F-8826, F-8839, the `closeTxd` latch and the unconfirmed-channel teardown fix master today and could each have been a same-day PR; worth knowing if anyone needs a backport.

Three things are deliberately left out rather than widening this further:

- a public accessor for the bytes buffered on a channel. #1212 reaches into `inputBuffer` because `wolfSSH_stream_peek()` goes blind once `eofRxd` is set, and it lands ahead of this branch, so the API cannot come from here
- a release note for `wolfSSH_worker()`'s changed return contract
- a half-close in `examples/client`, which is a feature rather than a fix

Cross-PR: #1188's unlanded half rewrites `echoserver`'s `ssh_worker()` with hunks that bracket this one's insertion point. They auto-merge today, but whoever lands second should confirm the EOF arm survived -- if it is lost the SIGKILL comes back and ships green.

The review threads below predate a re-cut of the branch and are stale anchors, not open questions. Each one carried its resolution and each fix is in the commit that now owns it; the two Fenrir comments on `wolfSSH_stream_send_eof()` are the unconfirmed-channel commit.

Issue: F-1687, F-8826, F-8839

### DIFF
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -908,6 +908,8 @@ static int SFTP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
                                 * if there is still pending sends */
                 }
                 if (error == WS_EOF) {
+                    /* An ordinary session end, not a failure. */
+                    ret = 0;
                     break;
                 }
             }
@@ -934,10 +936,19 @@ static int SFTP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
                     continue;
                 }
 
+                /* Drain what is buffered first. A rekey is not a drained
+                 * channel: peek reports it without looking. */
                 if (error == WS_EOF) {
-                    break;
+                    int peekRet = wolfSSH_stream_peek(ssh, NULL, 1);
+
+                    if (peekRet != WS_REKEYING && peekRet <= 0) {
+                        /* An ordinary session end, not a failure. */
+                        ret = 0;
+                        break;
+                    }
                 }
-                if (ret != WS_SUCCESS && ret != WS_CHAN_RXD) {
+                if (ret != WS_SUCCESS && ret != WS_CHAN_RXD
+                        && ret != WS_EOF) {
                     /* If not successful and no channel data, leave. */
                     break;
                 }
@@ -954,8 +965,11 @@ static int SFTP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
                     error == WS_CHAN_RXD || error == WS_REKEYING ||
                     error == WS_WINDOW_FULL)
                     ret = error;
-                if (error == WS_EOF)
+                if (error == WS_EOF) {
+                    /* An ordinary session end, not a failure. */
+                    ret = 0;
                     break;
+                }
                 continue;
             }
             else if (ret == WS_REKEYING) {
@@ -964,8 +978,11 @@ static int SFTP_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
             }
             else if (ret < 0) {
                 error = wolfSSH_get_error(ssh);
-                if (error == WS_EOF)
+                if (error == WS_EOF) {
+                    /* An ordinary session end, not a failure. */
+                    ret = 0;
                     break;
+                }
             }
 
             if (ret == WS_FATAL_ERROR && error == 0) {
@@ -1325,6 +1342,19 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
                     else if (rc == WS_CHANNEL_CLOSED) {
                         continue;
                     }
+                    else if (rc == WS_EOF) {
+                        /* The peer is done sending. No EOF of ours here: it
+                         * latches eofTxd and the child's remaining console
+                         * output would then be refused, which both send sites
+                         * below treat as fatal. wolfSSH_shutdown() sends it at
+                         * teardown, as the POSIX copy relies on. Closing the
+                         * write end of the child's stdin is still owed on this
+                         * platform, and so is the per-pass drain: ptyIn is the
+                         * terminal-resize context, and this copy reads into
+                         * shellBuffer, which the windowFull resend owes the
+                         * peer. Both want fixing where they can be tested. */
+                        continue;
+                    }
                     else if (rc != WS_WANT_READ) {
                         break;
                     }
@@ -1518,8 +1548,10 @@ static int SHELL_FlushOut(WOLFSSH* ssh, WS_SOCKET_T sshFd, word32 channelId,
             if (wolfSSH_worker(ssh, NULL) < 0) {
                 int err = wolfSSH_get_error(ssh);
 
+                /* A peer EOF during the final flush is expected. */
                 if (err != WS_WANT_READ && err != WS_WANT_WRITE &&
-                        err != WS_CHAN_RXD && err != WS_REKEYING) {
+                        err != WS_CHAN_RXD && err != WS_REKEYING &&
+                        err != WS_EOF) {
                     wolfSSH_Log(WS_LOG_ERROR,
                         "[SSHD] Issue draining connection on final flush");
                     return -1;
@@ -1824,11 +1856,9 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     word32 shellChannelId = 0;
     WOLFSSH_CHANNEL* shellChannel;
 
-    /* Name the session channel off the channel list rather than trusting
-     * DEFAULT_NEXT_CHANNEL to be 0, which a build can override. It is the
-     * only channel open at this point; the agent channel comes later. The
-     * loop below closes the child's stdin off this channel, so a wrong id
-     * there drops the peer's input instead of handing it over. */
+    /* Off the channel list, not DEFAULT_NEXT_CHANNEL, which a build can
+     * override. The session channel is the only one open here. lastRxId is
+     * no use: no request path sets it. A wrong id drops the peer's input. */
     shellChannel = wolfSSH_ChannelNext(ssh, NULL);
     if (shellChannel == NULL || wolfSSH_ChannelGetId(shellChannel,
             &shellChannelId, WS_CHANNEL_ID_SELF) != WS_SUCCESS) {
@@ -1959,6 +1989,9 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
                     }
                     peerConnected = 0;
                     continue;
+                }
+                else if (rc == WS_EOF) {
+                    /* Half-close, handled below. */
                 }
                 else if (rc == WS_WANT_WRITE) {
                     wantWrite = 1;
@@ -2716,7 +2749,7 @@ static void* HandleConnection(void* arg)
                 error = wolfSSH_get_error(ssh);
 
                 /* peer successfully closed down gracefully */
-                if (ret == WS_CHANNEL_CLOSED) {
+                if (ret == WS_CHANNEL_CLOSED || ret == WS_EOF) {
                     ret = 0;
                     break;
                 }
--- a/apps/wolfsshd/wolfsshd.c
+++ b/apps/wolfsshd/wolfsshd.c
@@ -1618,6 +1618,11 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
     int   peerConnected = 1;
     int   stdoutEmpty = 0;
     int   ptyReq = 0;
+    int   childInSz = 0;  /* Bytes read off the channel into channelBuffer
+                           * that the child has yet to take. The read is
+                           * destructive, so what a short write leaves is
+                           * carried to the next pass.  */
+    int   childInIdx = 0; /* How much of those the child has taken.  */
 
     childFd = -1;
     stdoutPipe[0] = -1;
@@ -1904,6 +1909,14 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         close(stdinPipe[0]);
     }
 
+    /* The loop below is the only reader of the child's output, so it must
+     * never be the thing waiting on the child. A descriptor that cannot be
+     * made non-blocking keeps the old behaviour: the write can stall, which
+     * is still better than dropping the peer's input on the floor. The read
+     * paths already treat EAGAIN as nothing to report. */
+    (void)SHELL_SetNonBlocking((!ptyReq || forcedCmd) ?
+            stdinPipe[1] : childFd);
+
     while (ChildRunning || windowFull || !stdoutEmpty || peerConnected) {
         byte tmp[2];
         fd_set readFds;
@@ -1913,6 +1926,8 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
         int cnt_w;
         WOLFSSH_CHANNEL* current;
         int pending = 0;
+        int childStalled = 0; /* The child would not take the rest of what it
+                               * is owed on this pass.  */
 
         FD_ZERO(&readFds);
         FD_SET(sshFd, &readFds);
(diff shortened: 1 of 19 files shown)


tests/unit.c
```diff
@@ -5960,8 +6034,290 @@ static int test_ChannelReadExtClearsStaleWantWrite(void)
...
+    WOLFSSH*         ssh = NULL;
+    WOLFSSH_CHANNEL* ch  = NULL;
+    int              result = 0;
+    int              ret;
+    word32           queued;
+
+    ctx = wolfSSH_CTX_new(WOLFSSH_ENDPOINT_CLIENT, NULL);
+    if (ctx == NULL)
+        return -1510;
+    wolfSSH_SetIOSend(ctx, WantWriteIoSend);
```
ejohnstown, line 8153: Confirmed. `gcc -fsyntax-only -DNO_WOLFSSH_SERVER` gives exactly one error, at this line; master's `unit.c` compiles clean, so it's a regression from this PR.

Fixed by moving `WantWriteIoSend` next to `DiscardIoSend`, outside the block. The test is a client-endpoint test, so moving the helper is the right half.

src/ssh.c
```diff
@@ -1252,7 +1260,13 @@ int wolfSSH_stream_read(WOLFSSH* ssh, byte* buf, word32 bufSz)
                     "Starting to receive data at current index of %u",
                     inputBuffer->idx);
             ret = DoReceive(ssh);
-            if (ssh->channelList == NULL || ssh->channelList->eofRxd)
+            /* An EOF ends the read only with nothing buffered. Off the
+             * current head: DoReceive() may have removed inputBuffer's
+             * channel. */
+            if (ssh->channelList == NULL
```
ejohnstown, line 1398: Confirmed, and reproduced: two channels, EOF staged for the non-head one, `wolfSSH_stream_read()` returned `-1031` with `ch0->eofRxd == 0`.

Fixed, but not with `WS_ERROR` as suggested. `DoReceive()` also latches `ssh->error = WS_EOF`, so `WS_ERROR` + `get_error() == WS_EOF` is byte-identical to a real head EOF. The read now keeps waiting instead; the head is still open and still has nothing buffered. Unlike the `WS_EXTDATA` case there's no buffered payload to strand. Regression test added; it fails without the fix.

src/internal.c
```diff
@@ -11314,7 +11312,11 @@ static int DoChannelClose(WOLFSSH* ssh,
 
     if (ret == WS_SUCCESS) {
         if (!channel->closeTxd) {
-            ret = SendChannelClose(ssh, channel->peerChannel);
+            /* RFC 4254 section 5.3 wants the EOF ahead of the close. */
+            ret = SendChannelEof(ssh, channel->peerChannel);
```
ejohnstown: Confirmed, and a regression. Measured by reverting just this hunk:

| | close bundled | `closeTxd` |
|---|---|---|
| master | yes | 1 |
| this PR | no | 0 |

Fixed: both packets sent unconditionally, flush status still returned. Also moved `ChannelRemove()` onto the want-write path -- since `wolfSSH_ChannelExit()` no longer removes, this is the only retirement point, and the pending flush lives on the output buffer, not the channel. Regression test added.

ejohnstown: Same defect as the thread above, fixed with it.

examples/echoserver/echoserver.c
```diff
@@ -1095,6 +1095,20 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
+                        /* The peer half-closed. Answer with our own EOF for
+                         * that channel: the library reports the event now
+                         * rather than echoing it, so a peer waiting on our
+                         * EOF would otherwise wait forever. Send once. */
+                        WOLFSSH_CHANNEL* eofChannel;
+
+                        eofChannel = wolfSSH_ChannelFind(ssh, lastChannel,
+                                WS_CHANNEL_ID_SELF);
+                        if (eofChannel != NULL) {
+                            wolfSSH_ChannelSendEof(eofChannel);
```
ejohnstown: Confirmed, and reverted. `SendChannelData()` returns `WS_EOF` once `eofTxd` is set (internal.c:20315) and `ssh_worker` breaks on the negative return, so the child's remaining output is dropped.

Worth recording why it was there: an earlier review round flagged that servers had stopped emitting CHANNEL_EOF and suggested replying on the peer's EOF. That suggestion was wrong -- the auto-reply is F-1687, the bug this PR removes from the library, so recreating it in the example is not the fix.

Both echoservers now leave the reply to the application, with a comment saying why. The remaining half of the earlier finding stands: the new APIs still have no non-test caller. Demonstrating them needs a "local side has finished producing output" signal the echoserver does not currently have.

ejohnstown: Same defect as the thread above, reverted with it.

src/internal.c
```diff
@@ -11311,17 +11309,36 @@ static int DoChannelClose(WOLFSSH* ssh,
...
         }
     }
 
-    if (ret == WS_SUCCESS) {
-        ret = ChannelRemove(ssh, channelId, WS_CHANNEL_ID_SELF);
-    }
+    /* Retire the channel once the close is bundled. Nothing past this point
+     * needs it, and a blocked flush is on the output buffer, not the
+     * channel, so waiting on the socket would strand the channel for good. */
+    if (ret == WS_SUCCESS || ret == WS_WANT_WRITE) {
```
ejohnstown, line 11408: Confirmed and fixed in 55612c27, but not by widening the wolfSSH_worker() flush condition. That block ends with `ret = sendRet`, so a still-blocked flush replaces WS_CHANNEL_CLOSED with WS_WANT_WRITE and costs the caller the graceful-close signal; tried it, it fails test_DoChannelCloseWantWrite. The worker now makes a separate best-effort flush that leaves both the status and the latched error alone.

Regression test added; it fails without the fix.

src/wolfsftp.c
```diff
@@ -757,7 +757,9 @@ static int wolfSSH_SFTP_buffer_read(WOLFSSH* ssh, WS_SFTP_BUFFER* buffer,
             }
         }
 
-        if (!wolfSSH_stream_peek(ssh, peekBuf, 1)) {
+        /* Nothing buffered, including the WS_ERROR a drained channel at EOF
+         * reports. Poll for the real status. */
+        if (wolfSSH_stream_peek(ssh, peekBuf, 1) <= 0) {
```
ejohnstown: Confirmed and fixed in 55612c27. A negative peek with WS_EOF latched now returns WS_FATAL_ERROR instead of polling. WS_REKEYING still polls: peek reports it before it looks at the buffer.

Not reachable through wolfsshd, which peeks before wolfSSH_SFTP_read() and breaks on WS_EOF. Regression test added; it fails without the fix.

philljj (dismissed):
Fenrir nits look valid

src/ssh.c
```diff
@@ -4052,11 +4122,30 @@ int wolfSSH_ChannelExit(WOLFSSH_CHANNEL* channel)
...
+    /* Only KEX traffic may go out mid-rekey, RFC 4253 section 7.1. */
+    if (ret == WS_SUCCESS && channel->ssh->isKeying) {
+        channel->ssh->error = WS_REKEYING;
+        ret = WS_REKEYING;
+    }
+
     if (ret == WS_SUCCESS)
-        ret = ChannelRemove(channel->ssh,
-                channel->peerChannel, WS_CHANNEL_ID_PEER);
+        ret = SendChannelEof(channel->ssh, channel->peerChannel);
```
ejohnstown, line 4460: Confirmed and reproduced: with the session channel at peer id 0 and a second channel unconfirmed, `wolfSSH_ChannelSendEof()` returned `WS_SUCCESS`, set the session's `eofTxd`, and left its sends failing `WS_EOF` for good. `wolfSSH_ChannelFwdNewLocal()`/`FwdNewRemote()` hand back exactly such a channel.

Fixed in 1dc7abd3 with the suggested `WS_CHANNEL_NOT_CONF`, after the rekey check so a mid-KEX call still reports `WS_REKEYING`. Regression test added; it fails without the fix.

`wolfSSH_ChannelExit()` has the same gap, but it is on master too, so not this PR's.

src/internal.c
```diff
@@ -11272,10 +11272,8 @@ static int DoChannelEof(WOLFSSH* ssh,
 
     if (ret == WS_SUCCESS) {
         channel->eofRxd = 1;
-        if (!channel->eofTxd) {
-            ret = SendChannelEof(ssh, channel->peerChannel);
-        }
         ssh->lastRxId = channelId;
+        ret = WS_EOF;
```
ejohnstown, line 11357: Confirmed, and a regression. Measured by reverting just the `ret = WS_EOF;` line:

| | `wolfSSH_accept()` | `get_error()` |
|---|---|---|
| master | -1001 | -1010 `WS_WANT_READ` |
| this PR | -1001 | -1031 `WS_EOF` |

Fixed in 1dc7abd3, but narrower than suggested. Only three loops can run with a channel already open -- ssh.c:732, 1049, 1106; the earlier states have no channel for `DoChannelEof()` to find, so an EOF there is `WS_INVALID_CHANID`, fatal on master too. Those three go through a `DoReceiveHandshake()` wrapper that absorbs `WS_EOF`; the rest still call `DoReceive()` directly.

Not rolled into `DoReceive()`: `wolfSSH_worker()` needs the `WS_EOF` to reach the application.

Regression test added; it fails without the fix.

src/ssh.c
```diff
@@ -1394,6 +1442,29 @@ int wolfSSH_ChannelIdSendExt(WOLFSSH* ssh, word32 channelId,
...
+        ret = WS_BAD_ARGUMENT;
+
+    /* Only KEX traffic may go out mid-rekey, RFC 4253 section 7.1. */
+    if (ret == WS_SUCCESS && ssh->isKeying) {
+        ssh->error = WS_REKEYING;
+        ret = WS_REKEYING;
+    }
+
+    if (ret == WS_SUCCESS)
+        ret = SendChannelEof(ssh, ssh->channelList->peerChannel);
```
ejohnstown, line 1602: Fixed: wolfSSH_stream_send_eof() now returns WS_CHANNEL_NOT_CONF on an unconfirmed head, with unit coverage. The reachable case is self-targeting, not cross-channel -- a lone unconfirmed head has peerChannel 0, so the lookup resolves to itself. Same change gates both EOF senders on SendAfterDisconnect().

ejohnstown, line 1602: Fixed: wolfSSH_stream_send_eof() now returns WS_CHANNEL_NOT_CONF on an unconfirmed head, with unit coverage. The reachable case is self-targeting, not cross-channel -- a lone unconfirmed head has peerChannel 0, so the lookup resolves to itself. Same change gates both EOF senders on SendAfterDisconnect().

examples/echoserver/echoserver.c
```diff
@@ -997,6 +1005,57 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
+                            eofOff = 0;
+                            /* A send is bounded by the peer's window and
+                             * packet size, so a short one is normal and the
+                             * rest of the chunk is still owed. */
+                            while (eofOff < eofRead) {
+                                eofSent = wolfSSH_ChannelIdSend(ssh,
+                                        threadCtx->shellCtx.channelId,
+                                        threadCtx->channelBuffer + eofOff,
+                                        eofRead - eofOff);
+                                if (eofSent <= 0)
```
ejohnstown: Confirmed and fixed. The tail could not stay in channelBuffer -- the WS_CHAN_RXD read just below reuses it in the same pass -- so the drain got its own buffer. The chunk and its offset now persist across worker passes, the next chunk is read only once the last is out, and the EOF waits on an explicit drained flag. Same change in the Espressif copy.

ejohnstown: Duplicate of the thread above; fixed with it.

examples/portfwd/portfwd.c
```diff
@@ -765,6 +767,44 @@ THREAD_RETURN WOLFSSH_THREAD portfwd_worker(void* args)
...
             }
 
+            /* Relay the half-close so a local reader waiting on end-of-input
+             * returns; nothing else relays it. Driven off the latched channel
+             * state, not the WS_EOF status: the flush inside wolfSSH_worker()
+             * can supersede that, and it is raised only once. Only the channel
+             * appFd is wired to, since half-closing the wrong socket truncates
+             * a live transfer. */
+            if (appFdSet && fwdChannel != NULL && !appFdHalfClosed
+                    && wolfSSH_ChannelGetEof(fwdChannel)) {
```
ejohnstown, line 789: Confirmed, and reproduced: with a peer that refuses the open, portfwd segfaults under libgmalloc before the fix and exits clean after. Fixed as suggested -- the channel id is stashed wherever the channel is created or adopted, re-resolved by id each pass, and the loop stops once the lookup fails.

examples/echoserver/echoserver.c
```diff
@@ -997,6 +1011,66 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
                 cnt_r = wolfSSH_worker(ssh, &lastChannel);
+
+                /* The peer is done sending: hand back the backlog and answer
+                 * its EOF, or a client that half-closed waits on a server
+                 * that never finishes -- the library no longer answers for
+                 * us. Off the channel's own state, not the WS_EOF status: the
+                 * flush inside wolfSSH_worker() can supersede that, and it is
+                 * raised once. Echo mode only; a shell child on a pty is
+                 * still producing, so its EOF waits for the child to exit. */
+                if (!eofAnswered && echoOnly) {
```
ejohnstown, line 1025: Confirmed and fixed. The status is now read once, right after wolfSSH_worker() returns. A clean drain was safe -- neither call writes ssh->error on success -- but a stalled one latches WS_WINDOW_FULL or WS_WANT_WRITE from the send, or WS_REKEYING from the read, and the ladder ends in "rc != WS_WANT_READ", so the session died with the backlog unsent. Same change in the Espressif copy.

examples/echoserver/echoserver.c
```diff
@@ -997,6 +1011,66 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
+                        if (eofDrained) {
+                            wolfSSH_ChannelSendEof(eofChannel);
+                            eofAnswered = 1;
+                            ChildRunning = 0;
+                        }
+                    }
+                }
+
                 if (cnt_r < 0) {
                     rc = wolfSSH_get_error(ssh);
```
ejohnstown: Duplicate of the thread above; fixed with it.

apps/wolfsshd/wolfsshd.c
```diff
@@ -1903,24 +1948,53 @@ static int SHELL_Subsystem(WOLFSSHD_CONNECTION* conn, WOLFSSH* ssh,
...
-                    maxFd = childFd;
+        }
+
+        {
+            struct timeval noWait;
+            struct timeval* timeout = NULL;
+
+            /* Work already in hand must not wait on the descriptors, but the
+             * poll still runs so this pass sees the child's output too. */
+            if (pending) {
```
ejohnstown: Confirmed, and measured: 8MB into a remote "sleep 12" pins wolfsshd at 100% of a core (602 ticks over a 6s window), 0% with the fix. Gated the zero timeout on childInIdx == childInSz as suggested; the child stdin is already in the write set, so the pass now waits there. New sshd_stdin_stall_test.sh asserts it and fails without the fix.

examples/echoserver/echoserver.c
```diff
@@ -997,8 +1011,70 @@ static int ssh_worker(thread_ctx_t* threadCtx)
...
+                                    eofRead - eofOff);
+                            if (eofSent <= 0)
+                                break;
+                            eofOff += eofSent;
+                        }
+
+                        /* Only an emptied channel earns the EOF; anything
+                         * else is retried on a later pass. */
+                        if (eofDrained) {
+                            wolfSSH_ChannelSendEof(eofChannel);
```
ejohnstown: Confirmed in part, fixed differently.

The latch is real: `wolfSSH_ChannelIdRead()` has no rekey guard, so a drained channel still reports zero mid-rekey, and `wolfSSH_ChannelSendEof()` returns `WS_REKEYING` ahead of `PreparePacket()` having queued nothing. Reachable -- the echoserver's own highwater callback rekeys on receive.

But the reply is not lost. `wolfSSH_shutdown()` runs next and sends `CHANNEL_EOF` while `!closeTxd`; that message is outside the transport range, so keying does not block it. The peer sees the same bytes, one packet later.

And latching only on `WS_SUCCESS` would stall: `WS_WANT_WRITE` already bundled the EOF and set `eofTxd`, so a retry queues nothing and the loop re-enters an untimed `select()` on read descriptors with the half-closed peer sending nothing.

`WS_REKEYING` is the only status that queues nothing and is still recoverable, so it is the one the latch now skips. Same change in the Espressif copy.

philljj (approved):

PR #1197, opened 2026-08-25
wolfsftp: apply the attributes SETSTAT and FSETSTAT acknowledge

yosuke-wolfssl:
### Problem
`SFTP_SetFileAttributes()` and `SFTP_SetFileAttributesHandle()` left SIZE and
UIDGID as `@TODO` no-ops yet returned `WS_SUCCESS`, so `RecvSetSTAT`/`RecvFSetSTAT`
replied `SSH_FX_OK` for truncate and chown requests that never happened.
`WSETTIME`/`WFSETTIME` were `(0)` stubs on every port, so TIME was discarded too.

Closes f-11658.

### Fix (`src/wolfsftp.c`)
Every requested attribute is applied or reported. A port with no wrapper yields
`WS_UNIMPLEMENTED_E`, answered as `SSH_FX_OP_UNSUPPORTED`; the 64-bit size is
bounds-checked by `wResolveOffset()` against `WOLFSSH_MAX_FILE_OFFSET`.

| Flag | POSIX | Windows |
|---|---|---|
| SIZE | `truncate` / `ftruncate` | unsupported |
| UIDGID | `chown` / `fchown` | unsupported |
| PERM | `chmod` / `fchmod` | `_wchmod` via new `WS_ChmodA()` |
| TIME | `futimens` / `utimensat` | unsupported |

**Windows.** SFTP paths arrive as `/C:/...`, which `_chmod()` rejects.
`WS_ChmodA()` strips the leading root via `TrimFileName()` as the other Windows
routines do, so `WCHMOD` works on Windows for the first time. It is left out
under `_WIN32_WCE`, which does not include `<io.h>`.

**Client companion fix.** `wolfSSH_SFTP_CHMOD()` echoed a whole STAT reply back,
so once the server honours SIZE a chmod asked it to `truncate()` the target —
which fails on a directory. It now sends only `WOLFSSH_FILEATRB_PERM`.

### Tests
`TestSftpSetStatAttributes()` drives both handlers for size, timestamp and
ownership, asserting the on-disk result via `WSTAT`, with separate root and
non-root paths. `tests/sftp.c` adds an end-to-end chmod of a directory under a
fixed umask; `tests/api.c`'s `STATE_SET_ATR_SEND` case sends PERM only.

### Verification
- Negative controls fail first: pre-fix the status is OK while the file is
  unchanged; without the client fix the directory chmod fails; stubbing
  `WFCHOWN` fails the give-away-chown case; removing `WTRUNCATE` fails the
  `api.c` case before the PERM narrowing.
- `regress.test` passes as an ordinary user and under `sudo`. `make check`:
  10 pass, 1 skip, 2 fail — both `scripts/{sftp,scp}.test`, which pass on
  serial re-run (known parallel flake).
- GCC preflight clean across 6 configs; ASan + UBSan clean. Windows MSVC builds
  with SFTP enabled, `api-test`/`unit-test` pass, and `chmod 400`/`chmod 600`
  over SFTP toggles the read-only attribute. OpenSSH interop: `chmod` on a
  directory and `put -p` both correct.

### Not in this PR
- No 32-bit build was run locally; the `wResolveOffset()` bound only bites where
  `SIZEOF_OFF_T != 8`.
- `SFTP_SetFileAttributes()` stops at the first failing attribute rather than
  continuing best-effort, and the `UNSUPPORTED` mapping has no POSIX test since
  it is reachable only on ports lacking the wrappers.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -6061,7 +6061,7 @@ int wolfSSH_SFTP_RecvLSTAT(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
     return ret;
 }
 
-#if !defined(USE_WINDOWS_API) && !defined(WOLFSSH_ZEPHYR) \
+#if !defined(_WIN32_WCE) && !defined(WOLFSSH_ZEPHYR) \
     && !defined(WOLFSSH_SFTP_SETMODE) && !defined(WOLFSSH_FATFS)
 /* Set the files mode
  * return WS_SUCCESS on success */
@@ -6097,29 +6097,47 @@ static int SFTP_SetFileAttributes(WOLFSSH* ssh,
         char* name, WS_SFTP_FILEATRB* atr)
 {
     int ret = WS_SUCCESS;
+#ifdef WTRUNCATE
+    word64 sz;
+#endif
 
     /* check if size attribute present */
-    if (atr->flags & WOLFSSH_FILEATRB_SIZE) {
-        /* @TODO set file size */
+    if (ret == WS_SUCCESS && (atr->flags & WOLFSSH_FILEATRB_SIZE)) {
+#ifdef WTRUNCATE
+        if (wResolveOffset(atr->sz, WOLFSSH_MAX_FILE_OFFSET, &sz) != 0
+                || WTRUNCATE(ssh->fs, name, sz) != 0) {
+            ret = WS_BAD_FILE_E;
+        }
+#else
+        ret = WS_UNIMPLEMENTED_E;
+#endif
     }
 
     /* check if uid and gid attribute present */
-    if (atr->flags & WOLFSSH_FILEATRB_UIDGID) {
-        /* @TODO set group and user id */
+    if (ret == WS_SUCCESS && (atr->flags & WOLFSSH_FILEATRB_UIDGID)) {
+#ifdef WCHOWN
+        if (WCHOWN(ssh->fs, name, atr->uid, atr->gid) != 0) {
+            ret = WS_BAD_FILE_E;
+        }
+#else
+        ret = WS_UNIMPLEMENTED_E;
+#endif
     }
 
-#if !defined(USE_WINDOWS_API) && !defined(WOLFSSH_ZEPHYR)
     /* check if permissions attribute present */
-    if (atr->flags & WOLFSSH_FILEATRB_PERM) {
+    if (ret == WS_SUCCESS && (atr->flags & WOLFSSH_FILEATRB_PERM)) {
         ret = SFTP_SetMode(ssh->fs, name, WOLFSSH_SFTP_SAFE_MODE(atr->per));
     }
-#endif
 
     /* check if time attribute present */
     if (ret == WS_SUCCESS && (atr->flags & WOLFSSH_FILEATRB_TIME)) {
+#ifdef WSETTIME
         if (WSETTIME(ssh->fs, name, atr->atime, atr->mtime) != 0) {
             ret = WS_BAD_FILE_E;
         }
+#else
+        ret = WS_UNIMPLEMENTED_E;
+#endif
     }
 
     /* check if extended attributes are present */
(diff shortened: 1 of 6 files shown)


src/wolfsftp.c
```diff
@@ -6078,29 +6078,47 @@ static int SFTP_SetFileAttributes(WOLFSSH* ssh,
         char* name, WS_SFTP_FILEATRB* atr)
 {
     int ret = WS_SUCCESS;
+#ifdef WTRUNCATE
+    word64 sz;
+#endif
 
     /* check if size attribute present */
     if (atr->flags & WOLFSSH_FILEATRB_SIZE) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Confirmed — on Windows `PERM` now works via `_chmod` while `TIME` has no wrapper,
so `PERM|TIME` chmods the file and then answers `SSH_FX_OP_UNSUPPORTED`.
Reachable with a stock client: OpenSSH's `put -p` sends exactly that pair.

Leaving it for this PR, though. Partial application on error is normal for SFTP —
OpenSSH's own `process_setstat()` runs every block and returns the last error, so
a failed `utimes()` there also leaves an applied `chmod()` behind. And
pre-scanning is the only option that discards work the server could do: `put -p`
on Windows would apply no permissions at all, where today they land.

The compile-time-capability half of the point is fair, since that much is
knowable up front. Tracking as a follow-up — the cleaner fix is
`WSETTIME`/`WFSETTIME` on Windows via `SetFileTime`, which removes the case.

src/wolfsftp.c
```diff
@@ -6124,29 +6142,52 @@ static int SFTP_SetFileAttributesHandle(WOLFSSH* ssh,
         WS_SFTP_FILEATRB* atr)
 {
     int ret = WS_SUCCESS;
+#ifdef WFTRUNCATE
+    word64 sz;
+#endif
 
     /* check if size attribute present */
     if (atr->flags & WOLFSSH_FILEATRB_SIZE) {
```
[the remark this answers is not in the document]
yosuke-wolfssl: Same reasoning as the path-based setter — follow-up rather than a change here.

One correction: under `USE_WINDOWS_API` this variant can't partially apply, since
`WFTRUNCATE`, `WFCHOWN` and `WFSETTIME` are all undefined and `PERM` is
`#ifndef USE_WINDOWS_API` — the first requested flag returns `WS_UNIMPLEMENTED_E`
with the file untouched. The real exposure is a POSIX build where `HAVE_FUTIMENS`
fails: `PERM|TIME` applies the mode, then reports unsupported.

src/wolfsftp.c
```diff
@@ -6042,7 +6042,7 @@ int wolfSSH_SFTP_RecvLSTAT(WOLFSSH* ssh, int reqId, byte* data, word32 maxSz)
     return ret;
 }
 
-#if !defined(USE_WINDOWS_API) && !defined(WOLFSSH_ZEPHYR) \
+#if !defined(WOLFSSH_ZEPHYR) \
```
yosuke-wolfssl: Fixed — `!defined(_WIN32_WCE)` added back to the `SFTP_SetMode()` guard.

Confirmed as described: `<io.h>` is `#ifndef _WIN32_WCE` in port.h's Windows
block, so dropping `!defined(USE_WINDOWS_API)` to enable `_chmod` on Windows took
WinCE's exclusion with it. Keeping the Windows support, excluding only WinCE.

Note the guards still differ on `FREESCALE_MQX`, where `SFTP_SetMode()` is
likewise compiled but unreferenced. That predates this PR, so I left it.

tests/regress.c
```diff
@@ -3522,6 +3522,204 @@ static void TestSftpCloseFailureRemovesHandle(void)
...
+    AssertTrue(fd >= 0);
+    AssertIntEQ(WPWRITE(ssh->fs, fd, (byte*)body, (word32)(sizeof(body) - 1),
+            ofst), (int)(sizeof(body) - 1));
+    AssertIntEQ(WCLOSE(ssh->fs, fd), 0);
+    AssertIntEQ(WSTAT(ssh->fs, path, &st), 0);
+    AssertIntEQ((int)st.st_size, (int)(sizeof(body) - 1));
+
+    /* a SETSTAT that shrinks the file has to truncate it */
+    WMEMSET(&atr, 0, sizeof(atr));
+    atr.flags = WOLFSSH_FILEATRB_SIZE;
```
yosuke-wolfssl, line 8393: Added — SETSTAT and FSETSTAT cases with `atr.sz[1] = 0x80000000`, asserting
`WOLFSSH_FTP_FAILURE` and an unchanged on-disk size.

Worth flagging that they are an equivalent mutant on a 64-bit host: with the
`wResolveOffset()` call removed entirely, both still pass, because
`(off_t)0x8000000000000000` is negative and `truncate()` rejects it with `EINVAL`
on its own. No 64-bit input can distinguish the two paths.

The guard has teeth where `SIZEOF_OFF_T != 8` — there a 4 GiB request casts to
`0` and `truncate()` *succeeds*, silently zeroing the file. So the cases will
catch a regression on a 32-bit builder, not on this one. That is the same gap the
PR body flags under "Not in this PR": no 32-bit build was run locally.

Also added an FSETSTAT UIDGID pair while here — `WFCHOWN` had no coverage at all.
That one does kill its mutant: stubbing `WFCHOWN` to return 0 fails the
give-away-chown case.

tests/regress.c
```diff
@@ -3522,6 +3522,238 @@ static void TestSftpCloseFailureRemovesHandle(void)
...
+    WMEMSET(&atr, 0, sizeof(atr));
+    atr.flags = WOLFSSH_FILEATRB_UIDGID;
+    atr.uid = (word32)st.st_uid;
+    atr.gid = (word32)st.st_gid;
+    AssertIntEQ(SftpSendSetSTAT(ssh, rid, path, &atr), WS_SUCCESS);
+    AssertSftpStatusReply(ssh, rid++, WOLFSSH_FTP_OK);
+
+    /* giving the file away is refused for an unprivileged server, and that
+     * refusal has to reach the client instead of an OK */
+    if (geteuid() != 0) {
```
yosuke-wolfssl, line 8441: Fixed — both UIDGID positive cases now have a root branch that chowns to a
different uid, confirms the change with `WSTAT`, and restores the original owner.

You're right that the no-op case alone proves nothing: a `WCHOWN` stubbed to
return 0 passes it. The pair does have teeth — stubbing `WFCHOWN` to `(0)` fails
the give-away case — and the suite now passes both as an ordinary user and under
`sudo`, so each branch has actually executed rather than one being skipped.

tests/sftp.c
```diff
@@ -207,6 +207,26 @@ static int checkCreatStripsSpecialBits(void)
...
+static int checkChmodDirectory(void)
+{
+    WSTAT_T st;
+
+    WMEMSET(&st, 0, sizeof(WSTAT_T));
+    if (WSTAT(NULL, "test-chmod-dir", &st) != 0) {
+        fprintf(stderr, "stat test-chmod-dir failed\n");
+        return 1;
+    }
+    if ((st.st_mode & 0777) != 0700) {
```
yosuke-wolfssl, line 222: Fixed — wolfSSH_SftpTest() installs SFTP_TEST_UMASK (0022) before spawning
threads and restores the caller's umask before returning, so the directory is
0755 after mkdir and 0700 only if the chmod landed.

Confirmed as reported. Negative control was reverting the
`state->atr.flags = WOLFSSH_FILEATRB_PERM` line in `wolfSSH_SFTP_CHMOD()`:

  old check, umask 0022  -> FAIL
  old check, umask 0077  -> PASS   <- the vacuous case
  new check, both umasks -> FAIL (mode=000755, want 0700)

Note on the first of the two suggested fixes: recording the mode after mkdir
and asserting the post-chmod mode differs does not work on its own. Under umask
0077 the mkdir mode is already 0700, which is the chmod target, so that
assertion fails on a correct build. Normalising the umask is needed either way.

philljj:
Retest this please.

(Agent went offline during the build)

ejohnstown (approved):

PR #1202, opened 2026-08-27
fix: always send USERAUTH_FAILURE on reject

ejohnstown:
RFC 4252 section 5.1 has the server answer a request it does not accept with `USERAUTH_FAILURE`. A callback returning `WOLFSSH_USERAUTH_REJECTED` now always gets that reply.

`NO_FAILURE_ON_REJECTED` suppressed it. The macro was never set by configure, named in a header, or documented, and it guarded all four method handlers alike.

- drop the macro and its four guards
- add `test_UserAuthRejectedSendsFailure()`, asserting on the wire, since the handler returns `WS_USER_AUTH_E` either way

Issue: F-11672

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -8524,11 +8524,9 @@ static int DoUserAuthRequestNone(WOLFSSH* ssh, WS_UserAuthData* authData,
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                 WLOG(WS_LOG_DEBUG, "DUARN: none rejected");
-                #ifndef NO_FAILURE_ON_REJECTED
                 /* Count before failing: a send that blocks returns
                  * WS_WANT_WRITE, which isn't fatal on its own. */
                 (void)SendUserAuthFailureCount(ssh, 0, countIt);
-                #endif
                 ret = WS_USER_AUTH_E;
             }
             else if (ret == WOLFSSH_USERAUTH_WOULD_BLOCK) {
@@ -8657,9 +8655,7 @@ static int DoUserAuthInfoResponse(WOLFSSH* ssh,
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                 WLOG(WS_LOG_DEBUG, "DUARKB: keyboard rejected");
-                #ifndef NO_FAILURE_ON_REJECTED
-                    authFailure = 1;
-                #endif
+                authFailure = 1;
                 authRejected = 1;
                 ret = WS_USER_AUTH_E;
             }
@@ -8772,9 +8768,7 @@ static int DoUserAuthRequestPassword(WOLFSSH* ssh, WS_UserAuthData* authData,
                 }
                 else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                     WLOG(WS_LOG_DEBUG, "DUARPW: password rejected");
-                    #ifndef NO_FAILURE_ON_REJECTED
-                        authFailure = 1;
-                    #endif
+                    authFailure = 1;
                     authRejected = 1;
                     ret = WS_USER_AUTH_E;
                 }
@@ -10204,9 +10198,7 @@ static int DoUserAuthRequestPublicKey(WOLFSSH* ssh, WS_UserAuthData* authData,
                 ret = WS_AUTH_PENDING;
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
-                #ifndef NO_FAILURE_ON_REJECTED
-                    authFailure = 1;
-                #endif
+                authFailure = 1;
                 authRejected = 1;
                 ret = WS_USER_AUTH_E;
             }
@@ -17458,6 +17450,15 @@ int SendUserAuthKeyboardRequest(WOLFSSH* ssh, WS_UserAuthData* authData)
             WLOG(WS_LOG_DEBUG, "SUAKR: keyboard setup callback would block");
             ssh->kbSetupPending = 0;
             return WS_AUTH_PENDING;
+        }
+        else if (ret == WOLFSSH_USERAUTH_REJECTED) {
+            /* A hard rejection ends the session here as it does in the other
+             * methods. Count before failing: a send that blocks returns
+             * WS_WANT_WRITE, which isn't fatal on its own. */
+            WLOG(WS_LOG_DEBUG, "SUAKR: keyboard setup rejected");
+            ssh->kbSetupPending = 0;
+            (void)SendUserAuthFailureCount(ssh, 0, 1);
+            return WS_USER_AUTH_E;
         }
         else {
             WLOG(WS_LOG_DEBUG, "Issue with keyboard auth setup, try another "
--- a/tests/unit.c
+++ b/tests/unit.c
@@ -7673,13 +7673,17 @@ static int UnitAuthKbSetupOk(byte authType, WS_UserAuthData* authData,
     }
     return WOLFSSH_USERAUTH_INVALID_PASSWORD;
 }
+#endif /* WOLFSSH_KEYBOARD_INTERACTIVE */
 
+#if defined(WOLFSSH_KEYBOARD_INTERACTIVE) || \
+    defined(WOLFSSH_ALLOW_USERAUTH_NONE) || !defined(WOLFSSH_NO_PUBKEY_AUTH)
 /* Build a USERAUTH_REQUEST payload for `service` and auth method `method`
  * with no method-specific fields, as DoUserAuthRequest() sees it with the
  * message id already consumed. The keyboard and none dispatch paths don't
- * parse trailing fields, so three strings (user, service, method) suffice. A
- * NULL method omits that field, which a bad service name never reaches.
- * Returns the payload size, 0 on overflow. */
+ * parse trailing fields, so three strings (user, service, method) suffice,
+ * and the publickey path appends its own fields to this. A NULL method omits
+ * that field, which a bad service name never reaches. Returns the payload
+ * size, 0 on overflow. */
 static word32 BuildAuthMethodRequest(byte* buf, word32 bufSz,
         const char* service, const char* method)
 {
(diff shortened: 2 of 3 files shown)


src/internal.c
```diff
@@ -8506,11 +8506,9 @@ static int DoUserAuthRequestNone(WOLFSSH* ssh, WS_UserAuthData* authData,
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                 WLOG(WS_LOG_DEBUG, "DUARN: none rejected");
-                #ifndef NO_FAILURE_ON_REJECTED
                 /* Count before failing: a send that blocks returns
                  * WS_WANT_WRITE, which isn't fatal on its own. */
                 (void)SendUserAuthFailureCount(ssh, 0, countIt);
-                #endif
                 ret = WS_USER_AUTH_E;
```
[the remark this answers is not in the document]
ejohnstown: Deliberate. WOLFSSH_USERAUTH_REJECTED ends the session, so the send result is not what decides the outcome. Propagating WS_WANT_WRITE would not retry anything: the request is already consumed and there is no resume state, so the caller would treat it as non-fatal, move to the next packet, and drop the rejection. The buffered FAILURE stays in ssh->outputBuffer for a later flush.

src/internal.c
```diff
@@ -8639,9 +8637,7 @@ static int DoUserAuthInfoResponse(WOLFSSH* ssh,
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
                 WLOG(WS_LOG_DEBUG, "DUARKB: keyboard rejected");
-                #ifndef NO_FAILURE_ON_REJECTED
-                    authFailure = 1;
-                #endif
+                authFailure = 1;
                 authRejected = 1;
                 ret = WS_USER_AUTH_E;
```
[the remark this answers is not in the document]
ejohnstown: Deliberate. WOLFSSH_USERAUTH_REJECTED ends the session, so the send result is not what decides the outcome. Propagating WS_WANT_WRITE would not retry anything: the request is already consumed and there is no resume state, so the caller would treat it as non-fatal, move to the next packet, and drop the rejection. The buffered FAILURE stays in ssh->outputBuffer for a later flush.

src/internal.c
```diff
@@ -10186,9 +10180,7 @@ static int DoUserAuthRequestPublicKey(WOLFSSH* ssh, WS_UserAuthData* authData,
                 ret = WS_AUTH_PENDING;
             }
             else if (ret == WOLFSSH_USERAUTH_REJECTED) {
-                #ifndef NO_FAILURE_ON_REJECTED
-                    authFailure = 1;
-                #endif
+                authFailure = 1;
                 authRejected = 1;
                 ret = WS_USER_AUTH_E;
             }
```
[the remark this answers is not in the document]
ejohnstown: Deliberate. WOLFSSH_USERAUTH_REJECTED ends the session, so the send result is not what decides the outcome. Propagating WS_WANT_WRITE would not retry anything: the request is already consumed and there is no resume state, so the caller would treat it as non-fatal, move to the next packet, and drop the rejection. The buffered FAILURE stays in ssh->outputBuffer for a later flush.

tests/unit.c
```diff
@@ -7959,6 +7959,118 @@ static int CaptureIoSendAuthSvc(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
...
+            "keyboard-interactive");
+    if (requestSz == 0)
+        return -800;
+    ret = RunRejectedRequest(request, requestSz, "keyboard-interactive");
+    if (ret != 0)
+        return -800 - ret;
+#endif
+
+#ifdef WOLFSSH_ALLOW_USERAUTH_NONE
+    requestSz = BuildAuthMethodRequest(request, (word32)sizeof(request),
```
ejohnstown, line 8218: Fixed. The builder now sits under `#if defined(WOLFSSH_KEYBOARD_INTERACTIVE) || defined(WOLFSSH_ALLOW_USERAUTH_NONE) || !defined(WOLFSSH_NO_PUBKEY_AUTH)`; the third arm is for the new publickey builder. Checked with `-fsyntax-only` over the combinations, including none-auth with keyboard-interactive off.

ejohnstown, line 8218: Same as the guard finding above; fixed.

tests/unit.c
```diff
@@ -7959,6 +7959,118 @@ static int CaptureIoSendAuthSvc(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
...
+
+/* RFC 4252 section 5.1: a rejected request is answered with
+ * USERAUTH_FAILURE. The handler returns WS_USER_AUTH_E either way, so the
+ * assertion is on what went out on the wire. Every method the build dispatches
+ * has its own reject branch, so each is driven here.
+ *
+ * DoUserAuthInfoResponse() is the one reject branch left uncovered; reaching
+ * it needs an INFO_RESPONSE shim. The keyboard case here is the setup
+ * callback, which answers the request before any prompt goes out. */
+static int test_UserAuthRejectedSendsFailure(void)
```
ejohnstown, line 8190: Fixed. Added `BuildAuthPkRequest()` and a publickey case driving a no-signature PK_OK probe, which reaches the reject branch without a real key. The coverage comment holds again, with a note that the case needs one of the offered key algorithms compiled in.

src/internal.c
```diff
@@ -17415,6 +17407,15 @@ int SendUserAuthKeyboardRequest(WOLFSSH* ssh, WS_UserAuthData* authData)
             ssh->kbSetupPending = 0;
             return WS_AUTH_PENDING;
         }
```
philljj, line 17453: my skoll run was angry about this:

### HIGH-1: Hard keyboard rejection can drop the failure packet under backpressure
- **File:** `src/internal.c:17410-17417`
- **Function:** `SendUserAuthKeyboardRequest`

**Description:** The newly added branch discards the result of `SendUserAuthFailureCount()` and immediately returns `WS_USER_AUTH_E`. If the I/O callback returns `WS_CBIO_ERR_WANT_WRITE`, `wolfSSH_SendPacket()` leaves `USERAUTH_FAILURE` queued, but `DoReceive()` subsequently replaces `ssh->error` with `WS_USER_AUTH_E` and reports a fatal accept failure. Nonblocking accept loops retry only `WS_WANT_WRITE`, while `wolfSSH_shutdown()` only automatically flushes a queued disconnect, so callers such as wolfsshd can free the session without transmitting this failure. Before this PR, setup rejection used the ordinary failure branch and propagated `WS_WANT_WRITE`, allowing the next `wolfSSH_accept()` call to drain the packet. The new path therefore does not uphold the documented guarantee that the server answers before ending the session.

**Code:**
```
else if (ret == WOLFSSH_USERAUTH_REJECTED) {
    WLOG(WS_LOG_DEBUG, "SUAKR: keyboard setup rejected");
    ssh->kbSetupPending = 0;
    (void)SendUserAuthFailureCount(ssh, 0, 1);
    return WS_USER_AUTH_E;
}
```

**Recommendation:** Defer the terminal authentication error until the queued failure packet has been flushed.

philljj (changes requested):
merge conflict in unit.c, and skoll High.

src/internal.c
ejohnstown: Confirmed the mechanism.

Not new here, though: NO_FAILURE_ON_REJECTED is defined nowhere in the
tree, so master already discards the same send in the password,
publickey, keyboard-response and none handlers. This adds a fifth site
and deletes the dead macro.

Deferring the error until the flush isn't the fix either. Propagating
WS_WANT_WRITE retries nothing -- the request is consumed, no resume
state -- so the caller moves on and drops the rejection. That's the
pre-PR behaviour this change removes.

The fix is in teardown: key wolfSSH_shutdown()'s flush on pending output
rather than disconnectTxd, ahead of the NULL channelList bail. Covers
all five sites, but it has to follow the channel EOF/close PR in flight,
so deferring it rather than patching one site here.

Also narrower than rated: only bites below the auth-failure cap, where
no disconnect is queued to flush it.

ejohnstown:
Rebased onto master, so the unit.c conflict is gone. The two versions of
the request builder are merged: it takes the service name, and keeps the
guard the publickey case needs.

The skoll High is answered on its thread -- real, but not new to this PR,
and the fix belongs in the teardown flush, which has to follow the channel
EOF/close PR in flight.

tests/unit.c
```diff
@@ -7969,6 +8020,129 @@ static int CaptureIoSendAuthSvc(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
...
+    return result;
+}
+
+
+/* RFC 4252 section 5.1: a rejected request is answered with
+ * USERAUTH_FAILURE. The handler returns WS_USER_AUTH_E either way, so the
+ * assertion is on what went out on the wire. Every method the build dispatches
+ * has its own reject branch, so each is driven here.
+ *
+ * DoUserAuthInfoResponse() is the one reject branch left uncovered; reaching
```
ejohnstown: Right, the shim exists. Added RunRejectedInfoResponse(): a setup-OK/response-REJECTED callback, an INFO_RESPONSE through wolfSSH_TestDoUserAuthInfoResponse(), and the wire assertion. Comment corrected.

One correction to the framing: nothing regressed at that branch. NO_FAILURE_ON_REJECTED is defined nowhere in the tree, so all four guard removals are no-ops in every build. The only new behaviour is the setup-reject branch, which the test already drove.

tests/unit.c
```diff
@@ -7969,6 +8020,129 @@ static int CaptureIoSendAuthSvc(WOLFSSH* ssh, void* buf, word32 sz, void* ctx)
...
+    ret = RunRejectedRequest(request, requestSz, "password");
+    if (ret != 0)
+        return -790 - ret;
+
+#ifdef WOLFSSH_KEYBOARD_INTERACTIVE
+    requestSz = BuildAuthMethodRequest(request, (word32)sizeof(request),
+            "ssh-connection", "keyboard-interactive");
+    if (requestSz == 0)
+        return -800;
+    ret = RunRejectedRequest(request, requestSz, "keyboard-interactive");
```
ejohnstown, line 8208: Confirmed: UnitAuthAlwaysReject stops at KEYBOARD_SETUP, so this case never reached DoUserAuthInfoResponse(). Covered now by RunRejectedInfoResponse(). Verified by dropping authFailure = 1 from that branch, which fails the new case.

philljj (approved):

PR #1213, opened 2026-08-28
ssh: flush the worker's queued output and report what it is waiting on

yosuke-wolfssl:
### Problem

A read-only application on a non-blocking socket could stall permanently.

A channel read credits the window, `ChannelCreditWindow()` bundles a
`CHANNEL_WINDOW_ADJUST` into `ssh->outputBuffer`, and the socket write blocks.
The credit is deliberately not re-parked — the packet is already encrypted and
sequenced, so re-sending it would inflate the peer's window — which leaves
`wolfSSH_SendPacket()` as the only thing that can discharge it.

The peer has spent its window and goes silent waiting for that adjust. The
application has nothing of its own to send, so it calls `wolfSSH_worker()`,
which is what `wolfssh/ssh.h` tells it to do. The worker gated its flush on
`DoReceive()`'s return value, and on an idle socket `DoReceive()` returns
`WS_FATAL_ERROR` — not one of the gated values. No write was attempted, on
that call or any later one.

### Fix (`src/ssh.c`)

- **`wolfSSH_worker()` flushes** whenever `ssh->outputBuffer` holds bytes,
  instead of only for `WS_SUCCESS`/`WS_WANT_READ`/`WS_CHAN_RXD`.
- **It reports a specific status.** `DoReceive()`'s `WS_FATAL_ERROR` is
  replaced with `ssh->error` when that holds a real code, so a caller can tell
  an idle socket from a decrypt or MAC failure.
- **The return describes the receive.** An owed flush is state, not an event,
  so it stays in `ssh->error`; only a send failure other than `WS_WANT_WRITE`
  reaches the return. No snapshot-and-restore of `ssh->error` is involved.
- Drops the nested second `DoReceive()`, its `WS_WINDOW_FULL` case, and the
  `WOLFSSH_TEST_BLOCK` fork of the function body.
- `GetInputData()` returns its condition and `DoReceive()` records it;
  `_ChannelRead()`, `_ChannelReadExt()` and
  `SendPendingChannelWindowAdjust()` lose their `savedError` dances.

### Callers (`apps/`, `examples/`)

- `FlushQueuedSend()` loops while `wolfSSH_get_error()` reports an owed flush
  within its deadline, and returns `WS_WANT_WRITE` if the queue is still out.
- `server_worker()` and `HandleConnection()` treat `WS_CHANNEL_CLOSED` from
  `wolfSSH_shutdown()` as a clean end alongside `WS_SOCKET_ERROR_E`; wolfsshd's
  graceful-shutdown drain keeps looping on `WS_EXTDATA`.
- `client`, `scpclient`, `sftpclient` and the `wolfssh` app gate on return
  values; 12 now-dead `WS_FATAL_ERROR` translations are removed.

### The contract this establishes

| Function shape | How status is reported |
|---|---|
| Returns a byte count (`stream_read`, `SendChannelData`, ...) | the return is data, so status goes in `ssh->error` |
| Returns a status (`wolfSSH_worker`, `wolfSSH_shutdown`, ...) | the return **is** the status; no second channel |

`wolfSSH_worker()` and `wolfSSH_shutdown()` move to the second row.
`SendChannelData()` and `SendChannelExtendedData()` are made to satisfy the
first properly, recording from a captured local rather than reading a possibly
stale `ssh->error`, and marking the owed flush when the flush of
already-queued output blocks — so the next call reports the block instead of
acknowledging a payload it never bundled.

**Two consequences to be aware of when integrating:**

- `wolfSSH_worker()` no longer returns `WS_WANT_WRITE` for an owed flush, and
  `wolfSSH_get_error()` after such a call reports `WS_WANT_WRITE` rather than
  the receive's status. Read the return for what happened; read
  `wolfSSH_get_error()` to ask whether a flush is still owed.
- When the receive and the flush both fail hard on the same call, the flush's
  error is what reaches the caller; the receive error is not preserved.

### Intentionally not in this PR

**`wolfSSH_SendPacket()` keeps writing `ssh->error`.** Removing it was tried
and reverted. It has 57 call sites; most hand the status onward without
recording it, and consumers several layers up read the field. Removing the
producer before its consumers deadlocked the SFTP suite under forced
write-blocking. Consumer-first is safe at every step; producer-first is not.

**Core public APIs still return generic `WS_FATAL_ERROR`.** `wolfSSH_accept()`,
`wolfSSH_connect()` and around ten others put the detail in `ssh->error`.
Migrating them is caller-visible for every integrator and wants its own audit.

**Example application loops are unchanged.** An app that selects only for readability
with no timeout still never calls `wolfSSH_worker()` once the peer goes silent,
so the flush gets no opportunity to run. `examples/echoserver` and
`examples/client` are in that shape. Correcting them requires rework on app side,
in different branch.

A follow-up PR covers these: ~198 `WS_FATAL_ERROR` collapse points in
`src/wolfsftp.c`, then its readers, then deleting the `SendPacket` write last.
Also, `examples/echoserver` and `examples/client` would be covered by
another PR.

### Verification

- `tests/unit.test` 119/0; `regress`, `api`, `auth`, `kex`, `testsuite` pass.
- `make check` clean apart from `sftp.test`/`scp.test`, which flake under
  parallel `make check` and pass serially — same as on `master`.
- Contention matrix (`-DWOLFSSH_TEST_BLOCK`, `WOLFSSH_BLOCK_PROB` 70/50/30):
  `sftp.test` passes at all three, as does a three-client sequential transfer
  (1KB/2MB/10MB) against a single non-blocking echoserver.
- gcc-13 `-Werror` clean across 6 configurations.

**Known coverage gap:** `scp.test` skips entirely in a `WOLFSSH_TEST_BLOCK`
build ("wolfscp client does not support non-blocking mode"), so SCP is affected
by these changes but unverified under contention.

### DIFF
--- a/apps/wolfssh/wolfssh.c
+++ b/apps/wolfssh/wolfssh.c
@@ -325,21 +325,26 @@ static int FlushQueuedSend(WOLFSSH* ssh, wolfSSL_Mutex* lock)
             wc_LockMutex(lock);
         }
         ret = wolfSSH_worker(ssh, NULL);
-        if (ret == WS_FATAL_ERROR) {
-            /* the session holds the detail behind a fatal error */
-            ret = wolfSSH_get_error(ssh);
-        }
         if (lock != NULL) {
             wc_UnLockMutex(lock);
         }
-    } while (ret == WS_WANT_WRITE && WTIME(NULL) < deadline);
 
-    /* The queue is out. Whatever the worker made of the peer's end of the
-     * conversation is for the reader to sort out. A rekey started on the way
-     * through is the reader's as well, the send itself went out. */
-    if (ret == WS_WANT_READ || ret == WS_CHAN_RXD || ret == WS_EXTDATA
-            || ret == WS_REKEYING) {
-        ret = WS_SUCCESS;
+        /* Anything outside the receive's own statuses is a failure. */
+        if (ret != WS_SUCCESS && ret != WS_WANT_READ && ret != WS_CHAN_RXD
+                && ret != WS_EXTDATA && ret != WS_REKEYING) {
+            break;
+        }
+    } while (wolfSSH_get_error(ssh) == WS_WANT_WRITE
+            && WTIME(NULL) < deadline);
+
+    /* Report only whether the queue went out. The deadline can run out
+     * with the packet still queued. */
+    if (ret == WS_SUCCESS || ret == WS_WANT_READ || ret == WS_CHAN_RXD
+            || ret == WS_EXTDATA || ret == WS_REKEYING) {
+        if (wolfSSH_get_error(ssh) == WS_WANT_WRITE)
+            ret = WS_WANT_WRITE;
+        else
+            ret = WS_SUCCESS;
     }
 
     return ret;
@@ -1353,9 +1358,6 @@ static THREAD_RETURN WOLFSSH_THREAD wolfSSH_Client(void* args)
 
         if (ret == WS_SUCCESS) {
             ret = wolfSSH_worker(ssh, NULL);
-            if (ret == WS_FATAL_ERROR) {
-                ret = wolfSSH_get_error(ssh);
-            }
             if (ret == WS_WANT_WRITE) {
                 /* The close messages are already out, whatever the drain
                  * still wants to send is a reply to the peer. */
--- a/src/internal.c
+++ b/src/internal.c
@@ -4344,6 +4344,8 @@ int wolfSSH_OutputPending(WOLFSSH* ssh)
 }
 
 
+/* Reports what the read did: WS_SUCCESS, WS_WANT_READ when the socket
+ * would block, or WS_MEMORY_E / WS_SOCKET_ERROR_E. */
 static int GetInputData(WOLFSSH* ssh, word32 size)
 {
     int in;
@@ -4351,11 +4353,6 @@ static int GetInputData(WOLFSSH* ssh, word32 size)
     /* Take into account the data already in the buffer. Update size
      * for what is missing in the request. */
     word32 haveDataSz;
-
-    /* reset want read state before attempting to read */
-    if (ssh->error == WS_WANT_READ) {
-        ssh->error = 0;
-    }
 
     haveDataSz = ssh->inputBuffer.length - ssh->inputBuffer.idx;
     if (haveDataSz >= size) {
@@ -4368,8 +4365,7 @@ static int GetInputData(WOLFSSH* ssh, word32 size)
     }
 
     if (GrowBuffer(&ssh->inputBuffer, size) < 0) {
-        ssh->error = WS_MEMORY_E;
-        return WS_FATAL_ERROR;
+        return WS_MEMORY_E;
     }
 
     /* read data from network */
@@ -4378,13 +4374,11 @@ static int GetInputData(WOLFSSH* ssh, word32 size)
                      ssh->inputBuffer.buffer + ssh->inputBuffer.length,
                      size);
         if (in == -1) {
-            ssh->error = WS_SOCKET_ERROR_E;
-            return WS_FATAL_ERROR;
+            return WS_SOCKET_ERROR_E;
         }
 
         if (in == WS_WANT_READ) {
-            ssh->error = WS_WANT_READ;
-            return WS_FATAL_ERROR;
+            return WS_WANT_READ;
         }
 
         if (in >= 0) {
@@ -4393,8 +4387,7 @@ static int GetInputData(WOLFSSH* ssh, word32 size)
         }
         else {
             /* all other unexpected negative values is a failure case */
-            ssh->error = WS_SOCKET_ERROR_E;
-            return WS_FATAL_ERROR;
+            return WS_SOCKET_ERROR_E;
         }
 
     } while (size);
@@ -7715,7 +7708,6 @@ int ChannelCreditWindow(WOLFSSH* ssh, WOLFSSH_CHANNEL* channel, word32 amount)
 static int SendPendingChannelWindowAdjust(WOLFSSH* ssh)
 {
     WOLFSSH_CHANNEL* cur;
-    int savedError;
     int ret = WS_SUCCESS;
 
     if (ssh == NULL)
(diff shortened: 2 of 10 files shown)


src/internal.c
```diff
@@ -20475,15 +20466,21 @@ int SendChannelData(WOLFSSH* ssh, word32 channelId,
...
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS) {
         ret = wolfSSH_SendPacket(ssh);
+        sendRet = ret;
+
+        /* The byte count is the return value, so a short write reaches the
+         * caller through wolfSSH_get_error(). */
+        if (sendRet != WS_SUCCESS)
+            ssh->error = sendRet;
+        if (sendRet == WS_WANT_WRITE)
```
yosuke-wolfssl, line 20483: Fixed. `plainSz` is now set where the flush at 20395 (`if (ssh->outputBuffer.length != 0)`) blocks, so the next call flushes first and returns `WS_WANT_WRITE`. That restores the pre-PR behaviour, which armed the brake from the ungated `if (ssh->error == WS_WANT_WRITE)` tail.

One correction: the flush at issue is the one at 20395, not the entry flush at 20383 — that one returns early and never reaches the `ret = dataSz` collapse.

Returning the status instead of `dataSz` is the better fix and also removes a payload loss that predates this PR. Deferred to the follow-up that migrates the SFTP readers off `ssh->error`, since it changes when they first see `WS_WANT_WRITE`.

yosuke-wolfssl, line 20483: Duplicate of the "Incorrect error handling" finding on this line; fixed by the same change. `plainSz` is set when the flush at 20395 blocks, so the next call reports the block instead of acknowledging the payload again.

src/internal.c
```diff
@@ -20598,15 +20596,19 @@ int SendChannelExtendedData(WOLFSSH* ssh, word32 channelId,
...
     /* at this point the data has been loaded into WOLFSSH structure and is
      * considered consumed */
-    if (ret == WS_SUCCESS)
+    if (ret == WS_SUCCESS) {
         ret = wolfSSH_SendPacket(ssh);
+        sendRet = ret;
+
+        if (sendRet != WS_SUCCESS)
+            ssh->error = sendRet;
+        if (sendRet == WS_WANT_WRITE)
```
yosuke-wolfssl, line 20615: Fixed, same as `SendChannelData`: `plainSz` is set when the flush of already-queued output blocks, so the next call reports the block rather than acknowledging stderr data that was never bundled.

tests/unit.c
```diff
@@ -6370,16 +6643,26 @@ static int test_ChannelReadExtClearsStaleWantWrite(void)
...
-    /* Nothing is queued, so no flush may be reported as owed. */
+
+    /* wolfSSH_worker() is what retires it: it flushes whatever is queued and
+     * rewrites ssh->error from the receive on every call. Nothing is queued
+     * here, so no flush is owed once it returns. */
+    s_recvPkt = NULL;
+    s_recvPktSz = 0;
+    s_recvPktOff = 0;
+    (void)wolfSSH_worker(ssh, NULL);
     if (wolfSSH_get_error(ssh) == WS_WANT_WRITE) { result = -1460; goto done; }
```
yosuke-wolfssl: Agreed — `PacketIoRecv()` always reports want-read here and `outputBuffer.length` is asserted 0 above, so the check could not fail. It now asserts `wolfSSH_worker()` returns `WS_WANT_READ` and `ssh->error == WS_WANT_READ`.

src/ssh.c
```diff
@@ -3583,66 +3581,50 @@ const char* wolfSSH_GetSessionCommand(const WOLFSSH* ssh)
...
-                ret = sendRet;
-            else
-                ret = recv2;
-        }
-        else {
-            /* Preserve meaningful receive status when send succeeded. */
-            if (sendRet != WS_SUCCESS)
+            /* The send reaches the caller only when the receive had nothing
+             * of its own to report. */
+            if (sendRet != WS_SUCCESS &&
```
yosuke-wolfssl: Fixed by removing the arbitration rather than narrowing it. `rxErr` is gone: `wolfSSH_worker()` now records and returns any send failure other than `WS_WANT_WRITE` unconditionally, so nothing can suppress a hard error.

The root cause was that `rxErr` was `ssh->error` sampled before the send, and `DoReceive()` writes that field only when the receive fails. On a successful receive it held a leftover from an earlier call, which then stood in for "the receive had something of its own to report".

A short write no longer reaches the return at all — it stays in `ssh->error`, which is where a caller asks whether a flush is owed. When the receive and the flush both fail hard on the same call, the flush's error is what reaches the caller; that precedence is now stated in the PR description.

apps/wolfssh/wolfssh.c
```diff
@@ -325,10 +325,6 @@ static int FlushQueuedSend(WOLFSSH* ssh, wolfSSL_Mutex* lock)
             wc_LockMutex(lock);
         }
         ret = wolfSSH_worker(ssh, NULL);
-        if (ret == WS_FATAL_ERROR) {
-            /* the session holds the detail behind a fatal error */
-            ret = wolfSSH_get_error(ssh);
-        }
         if (lock != NULL) {
             wc_UnLockMutex(lock);
         }
```
yosuke-wolfssl, line 330: Fixed. The loop runs while `wolfSSH_get_error()` reports an owed flush within the deadline, leaves on a status outside the receive's own, and returns `WS_WANT_WRITE` when the deadline expires with the packet still queued. It no longer reports `WS_SUCCESS` for an unsent packet.

tests/regress.c
```diff
@@ -4294,7 +4294,6 @@ static void TestShutdownKeepsFlushWantWrite(void)
      * unfinished flush is what the caller has to act on. */
     AssertNotNull(ssh->channelList);
     AssertIntEQ(wolfSSH_shutdown(ssh), WS_WANT_WRITE);
```
yosuke-wolfssl, line 4296: Restored. `TestShutdownKeepsFlushWantWrite` asserts `AssertIntEQ(wolfSSH_get_error(ssh), WS_WANT_WRITE);` again after the first `wolfSSH_shutdown()` call.

PR #1214, opened 2026-08-29
Guard remote forwards on a client, disconnect on a disallowed message

ejohnstown:
Two of the RFC conformance items reported in #1047. Of the other five, items 1, 5 and 6 are already fixed on master (`90692e69`, `df635a3e`, `9e02350a`), item 2 is PR #1148, and item 4 was considered and declined -- `direct-tcpip` is deliberately not direction-checked, and the default no-`fwdCb` path already answers `OPEN_ADMINISTRATIVELY_PROHIBITED`. These two were the remainder.

**Item 3 -- `tcpip-forward` handled without a role guard (RFC 4254 section 7.1).** `DoGlobalRequest()` dispatched `tcpip-forward` and `cancel-tcpip-forward` to `DoGlobalRequestFwd()` whatever side it was on, so a client with a `fwdCb` registered answered `REQUEST_SUCCESS` and acted on a forward the peer asked for. The callback was the only gate. A client now refuses both, answering `REQUEST_FAILURE` when the peer asked for a reply and staying quiet when it did not. The answer is given before the request body is parsed, so no forward state is touched on the way to the refusal.

**Item 7 -- no disconnect for a message the state disallows (RFC 4252 section 6).** `DoPacket()` returned `WS_MSGID_NOT_ALLOWED_E` and `DoReceive()` turned that into a local `WS_FATAL_ERROR`, so the socket simply went away and the peer was left to guess. It now sends `DISCONNECT` with `PROTOCOL_ERROR` where `DoPacket()` rejects the message id, skipped once the session is already over since RFC 4253 section 11.1 forbids sending after a disconnect. Every rejection there is a protocol error, not only the message-id range the RFC names, so the disconnect covers everywhere `IsMessageAllowed()` refuses a received message -- and the client side with it, `DoPacket()` being common to both.

Worth being precise on that second one: the RFC's MUST, respond by disconnecting, was already satisfied by the teardown. What was missing is the "preferably with a proper disconnect message" half.

Regress coverage:

- `TestGlobalRequestFwdOnClientSendsFailure`, `TestGlobalRequestFwdCancelOnClientSendsFailure`: both request names refused on a client, with a `fwdCb` registered throughout -- without the role check that callback is the only gate and would answer success
- `TestGlobalRequestFwdOnClientNoReplyStaysQuiet`: nothing goes back when the peer did not ask for a reply
- `TestGlobalRequestFwdOnServerStillSucceeds`: the same request a client refuses is still honoured on a server, so the role check costs the server nothing
- `TestServerHighMsgIdBeforeAuthDisconnects`: the case the RFC names, keyed but short of user auth
- `TestChannelOpenRejectedBeforeKex` and `TestServerServiceRequestRejectedDuringKeying` both asserted silence on the reject path; they now assert the disconnect

`regress.test`, `unit.test` and `api.test` pass.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -11617,6 +11617,22 @@ static int DoGlobalRequest(WOLFSSH* ssh,
     }
 
     if (ret == WS_SUCCESS) {
+#ifdef WOLFSSH_FWD
+        /* RFC 4254 section 7.1: a remote forward is the client's to ask for,
+         * so a client that receives the request answers a failure rather than
+         * registering a forward on the peer's say-so. Answered here and not
+         * in DoGlobalRequestFwd(), so the request body is never parsed and no
+         * forward state is touched. */
+        if ((globReqId == ID_GLOBREQ_TCPIP_FWD
+                    || globReqId == ID_GLOBREQ_TCPIP_FWD_CANCEL)
+                && ssh->ctx->side == WOLFSSH_ENDPOINT_CLIENT) {
+            WLOG(WS_LOG_WARN, "DGR: rejecting %s received by a client", name);
+            if (wantReply) {
+                ret = SendRequestSuccess(ssh, 0);
+            }
+        }
+        else
+#endif
         switch (globReqId) {
 #ifdef WOLFSSH_FWD
             case ID_GLOBREQ_TCPIP_FWD:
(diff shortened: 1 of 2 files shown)


tests/regress.c
```diff
@@ -2836,6 +2913,96 @@ static void TestGlobalRequestFwdNoCbSendsFailure(void)
...
+
+    inSz = BuildGlobalRequestFwdPacket("0.0.0.0", 2222, 0, 0, in, sizeof(in));
+    InitChannelOpenHarnessClient(&harness, in, inSz);
+    AssertIntEQ(wolfSSH_CTX_SetFwdCb(harness.ctx, AcceptFwdCb, NULL),
+            WS_SUCCESS);
+
+    ret = DoReceive(harness.ssh);
+
+    AssertIntEQ(ret, WS_SUCCESS);
+    AssertIntEQ(harness.io.outSz, 0);
```
ejohnstown, line 3253: Confirmed, and fixed.

You are right that the assertions held either way. With `wantReply=0` the pre-guard path reaches `DoGlobalRequestFwd()`, `AcceptFwdCb` returns `WS_SUCCESS`, and the reply block is skipped, so `WS_SUCCESS` with no output is equally the answer without the role check.

I verified that rather than reasoning about it: with the guard neutralized and only the two sibling tests skipped (those do discriminate, `REQUEST_SUCCESS` vs `REQUEST_FAILURE`), the suite still reported `regress: PASS`.

The test now registers a recording callback through `wolfSSH_SetFwdCbCtx()` and asserts it was never invoked. That is the discriminator, and it also asserts what the guard's comment actually claims -- that the request is turned away before any forward state is touched -- rather than only that no reply went out. With the guard neutralized it now fails with `expected: fwdCbCalled == 0 / result: 1 != 0`. The comment above the test records why silence alone is not enough, so it is less likely to be weakened again.

src/internal.c
```diff
@@ -12255,6 +12271,14 @@ static int DoPacket(WOLFSSH* ssh, byte* bufferConsumed)
...
     if (!IsMessageAllowed(ssh, msg, WS_MSG_RECV)) {
+        /* A message the state does not allow is a protocol error, and RFC
+         * 4252 section 6 asks for a disconnect message so the peer can tell
+         * why it was dropped rather than seeing the socket vanish. Skipped
+         * once the session is already over, since RFC 4253 section 11.1
+         * forbids sending after a disconnect. */
+        if (!ssh->disconnected) {
+            (void)SendDisconnect(ssh, WOLFSSH_DISCONNECT_PROTOCOL_ERROR);
+        }
         return WS_MSGID_NOT_ALLOWED_E;
```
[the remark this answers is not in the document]
ejohnstown: The queued DISCONNECT is not lost. `SendDisconnect()` sets `ssh->disconnectTxd` before it attempts the send, and on `WS_WANT_WRITE` `wolfSSH_SendPacket()` returns with the bytes still sitting in `ssh->outputBuffer`. `FlushQueuedDisconnect()` in `src/ssh.c` keys on exactly that pair -- `disconnectTxd` plus pending output -- and `wolfSSH_shutdown()` calls it ahead of the channel-list check, so a disconnect left queued by a short send goes out on teardown, with or without a channel to tear down. That path exists for this case.

Propagating the return value would also be the wrong fix. `DoReceive()` deliberately does not convert `WS_WANT_WRITE` to `WS_FATAL_ERROR`, so returning it would tell the caller the session is still viable and hide the protocol error. The retry then re-enters `DoPacket()` on the same packet -- the not-allowed path returns before `bufferConsumed` is set -- and `ssh->disconnected` is 1 by then, so `SendDisconnect()` is skipped and it falls through to fatal anyway. One extra round trip that flushes nothing.

The `(void)` form also matches the ten other `SendDisconnect()` call sites; deviating at this one would be the inconsistency.

Net effect at every point is an improvement over what was there before: the peer used to get silence, and now gets a DISCONNECT on a blocking socket and a queued one that teardown flushes.

philljj (changes requested):
needs rebase, merge conflict in tests/regress.c

src/internal.c
```diff
@@ -12246,7 +12313,14 @@ static int DoPacket(WOLFSSH* ssh, byte* bufferConsumed)
         return WS_OVERFLOW_E;
     }
 
-    if (!IsMessageAllowed(ssh, msg, WS_MSG_RECV)) {
+    if (!IsMessageAllowed(ssh, msg, WS_MSG_RECV) && MsgIdKnown(msg)) {
```
ejohnstown: Confirmed, and fixed.

The gate tested implemented-or-not on every refused id. RFC 4252 section 6
names the range instead, so ids 83-89 and 101-255 drew UNIMPLEMENTED before
auth rather than a disconnect. It now disconnects on the range too:

```c
if (!msgAllowed && (MsgIdKnown(msg) || MSGIDLIMIT_POST_USERAUTH(msg))) {
```

Nothing checks the auth state beside it because a refused id of 80 or higher
only happens before auth. A refused id below 80 is now answered off the
dispatch, so it cannot reach a handler either. Id 200 at ACCEPT_KEYED covers
it.

#8825 does not apply here: the frame advance uses the validated packet length,
not payloadIdx.

philljj (approved):

PR #1215, opened 2026-08-31
Allow terminal and SFTP/SCP in embedded client-only builds

peteclarke-del:
Two compile-time guards exclude code that an embedded, client-only target
needs, and neither guard matches what the code behind it actually does.

`WOLFSSH_TERM` is gated on `!NO_FILESYSTEM` in three places, in `src/ssh.c` and
`src/internal.c`. Neither the window dimensions nor the terminal mode string is
read from disk; both are computed. On a target built with `NO_FILESYSTEM`,
`WOLFSSH_TERM` therefore compiles away and a client cannot request a PTY or send
its terminal modes, with no diagnostic to say why.

The SFTP and SCP client entry point in `wolfSSH_stream_read` is gated on
`!NO_WOLFSSH_SERVER`, so a client-only build cannot use SFTP or SCP even though
the guarded code is the client half.

This drops the two guards. A build that defines neither `NO_FILESYSTEM` nor
`NO_WOLFSSH_SERVER` is unaffected.

Found while running wolfSSH as an SSH and SFTP client on a bare-metal Raspberry
Pi with no filesystem and no server, serving an Acorn 8-bit host over its 1MHz
bus. With these two guards dropped the client negotiates a PTY and SFTP
transfers work; without them `WOLFSSH_TERM` is silently inert and SFTP is
unavailable.

I have kept our display defaults out of this change deliberately: the 40-column
and vt100 values that suit the Acorn host are a local choice and are not
proposed here. If a configurable default would be welcome I am happy to follow
up with one.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -20864,7 +20864,7 @@ int SendChannelRequest(WOLFSSH* ssh, byte* name, word32 nameSz)
 }
 
 
-#if defined(WOLFSSH_TERM) && !defined(NO_FILESYSTEM)
+#if defined(WOLFSSH_TERM)
 
 /* out is always a TERMINAL_MODES_MAX_SZ buffer, and one byte is held back
  * for the terminator CreateMode() appends */
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1070,7 +1070,7 @@ int wolfSSH_connect(WOLFSSH* ssh)
             FALL_THROUGH;
 
         case CONNECT_CLIENT_CHANNEL_AGENT_REQUEST_SENT:
-        #if defined(WOLFSSH_TERM) && !defined(NO_FILESYSTEM)
+        #if defined(WOLFSSH_TERM)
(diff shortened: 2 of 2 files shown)


dgarske:
Hi @peteclarke-del , thank you for your contribution! Can you tell us more about your project and if you plan to submit any additional PR's? Since this is a small change we might treat as bug report and fix ourselves. See https://github.com/wolfSSL/wolfssh/blob/master/CONTRIBUTING.md for details. 
Thanks, David Garske, wolfSSL

ejohnstown:
@peteclarke-del Your project sounds pretty cool. I've only just heard of the Acorn; I'm a Commodore 64 guy.

I can only accept your change if you fill out a contributor agreement first. If I don't hear back from you in a few days, I'm going to recreate the change as a bug-fix based on your description text.

--John

ejohnstown:
I am going to close this PR. I'll have another one up treating this as a bugfix/changereq.

PR #1216, opened 2026-08-31
wolfsftp: keep the local file when resuming a get

yosuke-wolfssl:
### Problem
`wolfSSH_SFTP_Get()` opened the local destination with `CREATE_ALWAYS` on every Windows transfer, including a resumed one. The file was truncated to zero before `WriteFile()` resumed at the saved offset, so the already-downloaded prefix came back as a zero-filled gap — a `reget` on Windows silently corrupted the file it was meant to finish.

Separately, neither platform checked the saved offset against the destination's real size. On POSIX `"ab"` appends at the file's current end regardless of `gOfst`, so a destination that was deleted, replaced, or shortened between the interrupt and the `reget` got the remainder spliced in at the wrong position. The corruption is not Windows-only: on POSIX a `reget` onto a missing destination writes the remainder at offset 0, producing a 768-byte file holding `src[256..1024]`.

### Fix (`src/wolfsftp.c`)
- **`STATE_GET_LOOKUP_OFFSET`** stats the destination when the saved offset is nonzero and clears the offset unless the reported size matches it exactly, restarting the transfer from zero. This mirrors `STATE_PUT_STAT_REMOTE`, added for the upload direction in 11659, and is not `#ifdef`-guarded — it corrects the POSIX and Harmony append paths too.
- **`STATE_GET_LOOKUP_OFFSET`** also clears the saved offset when the remote size already in `state->attrib` from `STATE_GET_LSTAT` is less than or equal to it, so a shortened remote file cannot leave the local copy longer than its source. This mirrors `STATE_PUT_LOOKUP_OFFSET`.
- **`STATE_GET_OPEN_LOCAL`** passes `OPEN_EXISTING` when the write offset is nonzero and reserves `CREATE_ALWAYS` for offset zero.
- **`FILE_APPEND_DATA`** is dropped from the desired access: `GENERIC_WRITE` already grants it and the write offset comes from the `OVERLAPPED`. The open now reports `INVALID_HANDLE_VALUE` as `WS_BAD_FILE_E` rather than falling through on a stale `ret`.

Closes f-12547.

### Tests
`test_wolfSSH_SFTP_GetResume()` in `tests/api.c`, six cases over a 1024-byte source with a 256-byte offset:

| Source / destination state | `resume` | Expected result |
|---|---|---|
| holds 256 bytes unlike the source | `1` | staged prefix kept, remainder appended |
| missing | `1` | full re-download |
| shorter than the saved offset | `1` | full re-download |
| longer than the saved offset | `1` | full re-download |
| longer than the source | `0` | truncated to the source |
| source shortened below the saved offset | `1` | full re-download |

Staging a prefix that differs from the source is what separates a real resume from a re-download, and the comparison reads one byte past the expected size so a leftover tail fails.

### Verification
- GCC preflight sweep clean across all 6 configs; `make check` 7 passed / 2 skipped / 0 failed, including `scripts/sftp.test` and `get-put.test`.
- Windows (MSVC, Debug x64): `api-test` and `unit-test` both exit 0.
- Negative controls: without the size check the missing / short / long destination cases each fail independently on Linux; with `CREATE_ALWAYS` restored the resume case fails on Windows.

### Not in this PR
A failed local open jumps to `STATE_GET_CLEANUP` without closing the remote handle. This is pre-existing on every branch — the sibling `if (ret != 0)` path does the same — and routing through `STATE_GET_CLOSE_REMOTE` would overwrite `ret` and surface `WS_CLOSE_FILE_E` instead of `WS_BAD_FILE_E`, so fixing both branches properly belongs in its own change.

### DIFF
--- a/src/wolfsftp.c
+++ b/src/wolfsftp.c
@@ -9775,6 +9775,9 @@ int wolfSSH_SFTP_Get(WOLFSSH* ssh, char* from,
     WS_SFTP_GET_STATE* state = NULL;
     int sz;
     int ret = WS_SUCCESS;
+#ifdef USE_WINDOWS_API
+    DWORD creationDisp;
+#endif
 
     WLOG(WS_LOG_SFTP, "Entering wolfSSH_SFTP_Get()");
     if (ssh == NULL || from == NULL || to == NULL)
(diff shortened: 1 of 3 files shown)


src/wolfsftp.c
```diff
@@ -9853,6 +9856,23 @@ int wolfSSH_SFTP_Get(WOLFSSH* ssh, char* from,
                 /* if resuming then check for saved offset */
                 if (resume) {
                     wolfSSH_SFTP_GetOfst(ssh, from, to, state->gOfst);
+
+                    /* a saved offset is only usable if the local file still
+                     * holds exactly that many bytes */
+                    if (state->gOfst[0] > 0 || state->gOfst[1] > 0) {
+                        WMEMSET(&state->attrib, 0, sizeof(state->attrib));
+                        if (SFTP_GetAttributes(ssh->fs, to, &state->attrib, 1,
```
yosuke-wolfssl, line 9877: Confirmed and fixed in c57a9052.

`STATE_GET_LOOKUP_OFFSET` now reads the remote size that `STATE_GET_LSTAT` left in `state->attrib`, before the destination stat overwrites it, and clears the saved offset when the remote file no longer holds bytes past it:

```c
/* A saved offset is only usable while the remote file
 * still holds bytes past it. */
if ((state->gOfst[0] > 0 || state->gOfst[1] > 0)
        && (state->attrib.flags & WOLFSSH_FILEATRB_SIZE)
        && (state->attrib.sz[1] < state->gOfst[1]
            || (state->attrib.sz[1] == state->gOfst[1]
                && state->attrib.sz[0] <= state->gOfst[0]))) {
    WLOG(WS_LOG_SFTP, "Remote file too short");
    state->gOfst[0] = 0;
    state->gOfst[1] = 0;
}
```

This mirrors the source check `STATE_PUT_LOOKUP_OFFSET` already performs for the upload direction, `<=` included, so `wolfSSH_SFTP_Get()` and `wolfSSH_SFTP_Put()` now each validate both endpoints: the source must still hold bytes past the offset, and the destination must hold exactly that many.

`test_wolfSSH_SFTP_GetResume()` gains a sixth case — destination staged at exactly the saved offset so the destination check passes, source shortened below the offset, expecting a full re-download. It runs last because it shortens the source. Reverting the new gate makes that case fail; with it in place the suite is clean.

ejohnstown (approved):

PR #1219, opened 2026-09-01
ssh: flush any queued output at teardown

ejohnstown:
`wolfSSH_shutdown()` flushes whatever a short send left in the output buffer, not just a queued disconnect. A rejected auth's `USERAUTH_FAILURE` has no channel, and a channel close is retired the moment it is bundled, so neither had anything left to carry the retry.

- The gate still refuses a flush once the peer has disconnected, unless our own disconnect is the thing queued.
- Tests cover the rejected auth with no channel, the close whose channel was already retired, and a flush that short-sends on a live channel.

### DIFF
--- a/src/ssh.c
+++ b/src/ssh.c
@@ -1149,17 +1149,20 @@ static int SendAfterDisconnect(WOLFSSH* ssh)
 }
 
 
-/* A disconnect of ours left queued by a short send still has to reach the
+/* Whatever a short send left in the output buffer still has to reach the
  * peer, and flushing bytes that are already bundled is not the new traffic
- * RFC 4253 section 11.1 forbids. Only our own disconnect qualifies: a
- * disconnect from the peer leaves nothing queued but unrelated traffic,
- * which the session is over for. Call only after a NULL check of ssh. */
-static int FlushQueuedDisconnect(WOLFSSH* ssh)
+ * RFC 4253 section 11.1 forbids. Once the peer has disconnected, though,
+ * only our own queued disconnect still qualifies: anything else in there
+ * belongs to a session that is over. Call only after a NULL check of ssh. */
+static int FlushQueuedOutput(WOLFSSH* ssh)
 {
-    if (!ssh->disconnectTxd || !wolfSSH_OutputPending(ssh))
+    if (!wolfSSH_OutputPending(ssh))
         return 0;
 
-    WLOG(WS_LOG_DEBUG, "Flushing a disconnect left queued by a short send");
+    if (ssh->disconnected && !ssh->disconnectTxd)
+        return 0;
+
+    WLOG(WS_LOG_DEBUG, "Flushing output left queued by a short send");
     return 1;
 }
 
@@ -1168,6 +1171,7 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
 {
     int ret = WS_SUCCESS;
     int flushRet = WS_SUCCESS;
+    int flushed = 0;
     WOLFSSH_CHANNEL* channel = NULL;
 
     WLOG(WS_LOG_DEBUG, "Entering wolfSSH_shutdown()");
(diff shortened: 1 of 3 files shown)


ejohnstown:
Fixed: a flush that completes on a live session now clears the `WS_WANT_WRITE` the short send latched, as the disconnected path already did with `WS_DISCONNECT`. Regression test added.

The cited path does not reproduce, though. On the CHANNEL_CLOSE whose channel `DoChannelClose()` retired, `ssh->error` after the flush is `WS_CHANNEL_CLOSED`; on the rejected auth it is `WS_USER_AUTH_E`. `DoReceive()` overwrites the latched want-write before shutdown runs in both. The reachable path is a short send with no channel involved -- `wolfSSH_SendIgnore()`, `wolfSSH_global_request()` -- where `channelList == NULL` skips both the teardown sends and the `wolfSSH_worker()` call that would otherwise reset the error. That is what the test uses.

ejohnstown:
Fixed: the flush's `WS_WANT_WRITE` now only outranks the teardown result while the output buffer still holds bytes, so a teardown send that carried the leftovers out settles it. Regression test added, with only the data send and the flush short.

The gate is narrower than the suggested `wolfSSH_OutputPending()` check on its own: a flush that fails outright still reports. `WS_CBIO_ERR_GENERAL` frees the output buffer, so a bare pending check would turn that failure on the disconnect path into `WS_CHANNEL_CLOSED`.

src/ssh.c
```diff
@@ -1242,8 +1257,11 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
     }
 
     /* An unfinished flush outranks the channel status: the caller has to
-     * come back for the rest of the disconnect. */
-    if (flushRet != WS_SUCCESS)
+     * come back for the rest of the disconnect. Not so once a teardown send
+     * has carried the leftovers out with it: that write is settled, and the
+     * teardown result stands. A flush that failed outright still reports. */
+    if (flushRet != WS_SUCCESS &&
+            (flushRet != WS_WANT_WRITE || wolfSSH_OutputPending(ssh)))
```
ejohnstown: Fixed: a still-owed flush no longer outranks a teardown send that failed. A reset leaves the output buffer intact, so `wolfSSH_OutputPending()` stays true and the flush still reads as owed while the send holds the real reason the session is over. The `WS_WANT_WRITE` override now applies only when the teardown itself came back clean, so a `WS_SOCKET_ERROR_E` from `SendChannelEof()`/`SendChannelClose()` is what the caller sees. A flush that failed outright still reports regardless, so a reset on the retry is not masked in turn by the `WS_WANT_READ` from the close wait. Regression test added, with the data send and the flush short and the reset arriving under the teardown sends.

Two corrections to the report. The caller does terminate: on the next call the flush hits the reset directly, `flushRet` is no longer `WS_WANT_WRITE`, and the override reports `WS_SOCKET_ERROR_E`. The cost was one wasted round trip, not a loop.

And `ssh->error` staying `WS_WANT_WRITE` is not from this change. Nothing on the send path has ever latched `WS_SOCKET_ERROR_E` into `ssh->error` -- `SendPacketFlush()` sets it only for want-write, and `src/internal.c:5093,5108` are both receive-path -- so the same staleness follows a short `wolfSSH_stream_send()` and a reset on master today. Worth fixing at the source rather than here, since it would change `wolfSSH_get_error()` for every send site in the library.

src/ssh.c
```diff
@@ -1242,8 +1257,16 @@ int wolfSSH_shutdown(WOLFSSH* ssh)
...
     /* An unfinished flush outranks the channel status: the caller has to
-     * come back for the rest of the disconnect. */
-    if (flushRet != WS_SUCCESS)
+     * come back for the rest of the disconnect. Not so once a teardown send
+     * has carried the leftovers out with it: that write is settled, and the
+     * teardown result stands. Nor does it outrank a teardown send that
+     * failed: a reset leaves the buffer intact, so the flush still reads as
+     * owed while the send holds the real reason. A flush that failed
+     * outright reports whatever else happened. */
+    if (flushRet != WS_SUCCESS &&
```
ejohnstown, line 1269: Fixed: `wolfSSH_shutdown()` now skips the wait for the peer's close while the flush left output queued. The peer cannot answer a close it has not finished receiving, so there is nothing to read until the buffer drains, and the retry reports the flush's `WS_WANT_WRITE` instead of the worker's `WS_WANT_READ`. Regression test added: the close short-sends, the retry's own flush short-sends again, and the call returns `WS_WANT_WRITE` with the bytes still pending; a third call drains them and lands on the close wait.

Two corrections to the report. This state was reachable before the PR: with `closeTxd` set from a short-sent close, master's retry also skips the teardown and returns `WS_FATAL_ERROR` with `WS_WANT_READ`, and it never attempts the queued bytes at all. The PR narrowed that to the case where the retry's flush is itself short. And propagating `flushRet` whenever output is pending is the gate the previous thread replaced: a reset leaves the buffer intact, so that would mask a `WS_SOCKET_ERROR_E` from the teardown sends with `WS_WANT_WRITE` again.

philljj (approved):

PR #1220, opened 2026-09-01
Match and resolve remote forwards for the whole session

ejohnstown:
A client only matched inbound "forwarded-tcpip" opens once it had registered a forward, and a request that had registered one resolved it again after its own send, by which point a reentrant callback could have changed the answer.

- Match every client session, so one that registered nothing refuses the opens it has nothing to answer for. `wolfSSH_SetFwdRemoteMatch(WOLFSSH_FWD_MATCH_OFF)` is the opt-out for an application keeping its own bind list.
- Resolve what a request names before its send and hold it on the pending, so a cancel sent from a send callback takes the forward the interrupted request is registering rather than nothing.
- Requests in flight are visible to each other and to the open match, so a setup's listener can feed channels before the call returns and a cancel stops matching from where it goes out.

Issue: ZD-22195

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -4048,6 +4048,21 @@ static void FwdPendingVoid(WOLFSSH* ssh, const WOLFSSH_FWD_REMOTE* entry)
 }
 
 
+/* Hand every request in flight naming one forward over to another. A request
+ * resolves its forward before its send, so when an answer that send pumped in
+ * folds two registrations into one, the survivor is what it named. */
+static void FwdPendingRebind(WOLFSSH* ssh, const WOLFSSH_FWD_REMOTE* from,
+        WOLFSSH_FWD_REMOTE* to)
+{
+    WOLFSSH_FWD_PENDING* pend;
+
+    for (pend = ssh->fwdPendingHead; pend != NULL; pend = pend->next) {
+        if (pend->found == from)
+            pend->found = to;
+    }
+}
+
+
 /* A forward a request in flight is registering, or NULL. It is not on the
  * session's list until that request commits, but it is what the peer is being
  * asked for, so a request a callback sends meanwhile names the same one. */
@@ -4249,10 +4264,12 @@ static void FwdRemoteSettle(WOLFSSH* ssh, WOLFSSH_FWD_REMOTE* entry,
             WLOG(WS_LOG_INFO, "Remote forward reply named a port already "
                     "registered");
             /* Requests still queued on the stale entry asked about this bind,
-             * so they answer for the entry that stands at it now. Unlinking
-             * without this leaves them naming nothing, and a cancel among
-             * them would settle no forward. */
+             * so they answer for the entry that stands at it now, and so does
+             * one still in its send window. Unlinking without this leaves
+             * them naming nothing, and a cancel among them would settle no
+             * forward. */
             FwdReplyRebind(ssh, dup, entry);
+            FwdPendingRebind(ssh, dup, entry);
             FwdRemoteUnlink(ssh, ssh->ctx->heap, dup);
         }
     }
--- a/src/internal.c
+++ b/src/internal.c
@@ -4001,6 +4001,89 @@ static void FwdReplyRebind(WOLFSSH* ssh, const WOLFSSH_FWD_REMOTE* from,
 }
 
 
+/* Put a request on the session's list of requests in flight, where anything
+ * its own send reaches can see it. */
+static void FwdPendingPush(WOLFSSH* ssh, WOLFSSH_FWD_PENDING* pend)
+{
+    pend->next = ssh->fwdPendingHead;
+    ssh->fwdPendingHead = pend;
+}
+
+
+/* Take it back off, its send being over. Requests nest, so this is not always
+ * the head: a callback's request commits inside the one that ran it. */
+static void FwdPendingPop(WOLFSSH* ssh, WOLFSSH_FWD_PENDING* pend)
+{
+    WOLFSSH_FWD_PENDING* cur;
+
+    if (ssh->fwdPendingHead == pend) {
+        ssh->fwdPendingHead = pend->next;
+    }
+    else {
+        for (cur = ssh->fwdPendingHead; cur != NULL; cur = cur->next) {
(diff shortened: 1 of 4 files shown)


src/internal.c
```diff
@@ -4501,29 +4607,39 @@ static int FwdRemoteMatch(WOLFSSH* ssh, const char* addr, word32 port)
...
+    for (pend = ssh->fwdPendingHead; pend != NULL; pend = pend->next) {
+        if (pend->entry == NULL ||
+                !FwdRemoteAddrMatch(ssh, pend->entry, addr, port))
+            continue;
+
+        if (FwdPendingHasCancel(ssh, pend->entry))
+            continue;
+
+        return 1;
     }
```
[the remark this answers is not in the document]
ejohnstown: Confirmed. The pending scan now applies the same newest-request rule as the list scan, so a want-reply cancel that committed inside the setup's send stops matching. Added a regression test for that window; it accepts the open without the change.

src/internal.c
```diff
@@ -4501,29 +4607,45 @@ static int FwdRemoteMatch(WOLFSSH* ssh, const char* addr, word32 port)
...
+            continue;
+
+        /* A cancel a callback sent from this setup's send went out behind it
+         * and committed, so the newest request governs here too. */
+        newest = FwdReplyNewest(ssh, pend->entry);
+        if ((newest != NULL && newest->isCancel) ||
+                FwdPendingHasCancel(ssh, pend->entry))
+            continue;
+
+        return 1;
```
ejohnstown, line 4691: Confirmed. `FwdRemoteReply()` dequeues the slot before parking the verdict, so a mid-send REQUEST_FAILURE is invisible to every reply-queue scan and the pending match admitted the open; the commit then unlinked the forward, leaving the channel behind.

The pending scan now skips a request whose parked answer refused it. Not the suggested `newest == NULL` mirror -- a `wantReply=0` setup mid-send has no slot by construction, so that form would refuse opens it legitimately covers.

Regression test drives the refusal and the open from the IO send callback; it accepts the open without the change.

src/internal.c
```diff
@@ -4392,16 +4490,10 @@ void FwdPendingCommit(WOLFSSH* ssh, WOLFSSH_FWD_PENDING* pend)
...
+    FwdPendingPop(ssh, pend);
 
-    if (pend->entry != NULL && target != NULL) {
-        /* A callback the send ran registered this bind first, so the entry
-         * built for it is one too many. */
-        WFREE(pend->entry->bindAddr, heap, DYNTYPE_STRING);
-        WFREE(pend->entry, heap, DYNTYPE_FWD);
-        pend->entry = NULL;
-    }
+    target = pend->entry != NULL ? pend->entry : pend->found;
```
ejohnstown, line 4514: Doesn't reproduce, and the premise doesn't hold.

The removed dedupe never covered port 0: it was `FwdRemoteFind(ssh, pend->bindAddr, pend->bindPort)` with `bindPort == 0`, and `FwdRemoteFind()` skips `portPending` entries and needs an exact port match. No settled entry ever has `bindPort == 0` -- `FwdRemoteSettle()` unlinks on a reply naming port 0.

The fold is already in `FwdRemoteSettle()`, where `portPending` clears. That is the only place the port becomes known, it scans the whole list, and `FwdPendingCommit()` links `pend->entry` before calling it. The suggested fix is inert for the same reason as the original: at commit a port-0 entry still has `bindPort == 0`.

Probed both mid-send orderings, reply-then-setup and setup-then-reply. Each leaves one entry, and one cancel takes it to zero.

ejohnstown, line 4514: Confirmed, and it is a different case from the port-0 one on this line. Here `pend->entry` has an explicit port and the entry it duplicates is the port-pending one: an answer the send pumps in settles that entry onto the port this request is registering, and `FwdRemoteSettle()`'s fold scans `fwdRemoteList`, which the committing entry has not joined yet. `FwdRemoteMatch()` then keeps accepting opens for the bind after one cancel, as reported.

The commit now looks the bind up before linking and folds onto the registration standing there. Not quite the suggested form: freeing `pend->entry` needs the replies queued against it rebound first, since a request a callback sent can have resolved to it through `FwdPendingFind()` and committed a slot naming it. The lookup is skipped while the entry is still `portPending` -- that is the port-0 case from the other thread on this line.

Regression test drives the port-0 answer in from the IO send callback. Without the change the bind takes two registrations and one cancel leaves one still matching.

ejohnstown: This stands for the case as filed: at commit a port-0 entry still has `bindPort == 0`, so a lookup by its bind finds nothing, and the fold in `FwdRemoteSettle()` is what covers it.

The newer finding on this line is the mirror of it -- an explicit-port entry duplicating a port-pending one that an answer settled onto that port mid-send -- and that one does reproduce. Fixed on that thread, with the lookup skipped while the entry is `portPending` so it stays out of the port-0 case.

src/internal.c
```diff
@@ -4392,15 +4491,26 @@ void FwdPendingCommit(WOLFSSH* ssh, WOLFSSH_FWD_PENDING* pend)
...
+    /* Its send is over, so nothing a later request sends can name it. */
+    FwdPendingPop(ssh, pend);
 
-    if (pend->entry != NULL && target != NULL) {
-        /* A callback the send ran registered this bind first, so the entry
-         * built for it is one too many. */
-        WFREE(pend->entry->bindAddr, heap, DYNTYPE_STRING);
-        WFREE(pend->entry, heap, DYNTYPE_FWD);
-        pend->entry = NULL;
+    target = pend->entry != NULL ? pend->entry : pend->found;
```
ejohnstown, line 4514: Confirmed. `FwdRemoteUnlink()` voids `found` on every request still in its send window, so a cancel whose send pumped in the answer that folded its target committed naming nothing. The fold now rebinds requests in flight to the surviving entry alongside the queued replies. Only `found` is rebound: a pending's own entry is never on the list until it commits, so the fold cannot reach it. Regression test drives the port-0 answer in from the cancel's send; without the change the cancel names no forward and the open is accepted.

philljj (approved):

PR #1230, opened 2026-09-03
agent: let the application open the agent channel

ejohnstown:
The one server-side site that opens auth-agent@openssh.com sits inside
wolfSSH_accept(), so an application driving its own channels cannot reach
it: the session records the request and no channel follows.

- Add wolfSSH_AGENT_ChannelOpen(), the same open lifted out of accept(),
  which still calls it.
- It reports WS_BAD_ARGUMENT until the peer asks and is idempotent
  afterward, so an application can poll it.
- Publish the agent on a queued open too, so a retry after WS_WANT_WRITE
  does not open a second channel and leak the first.

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -1749,6 +1749,13 @@ int wolfSSH_AGENT_ChannelOpen(WOLFSSH* ssh)
          * checks below would report a channel a client never opened. */
         ret = WS_BAD_ARGUMENT;
     }
+    else if (SendAfterDisconnect(ssh)) {
+        /* The session is over, so neither a new open nor the flush of one
+         * queued before the disconnect may go out. RFC 4253 section 11.1.
+         * WS_DISCONNECT is in ssh->error, where the rest of the API puts
+         * it. */
+        ret = WS_FATAL_ERROR;
+    }
     else if (!ssh->useAgent) {
         /* Nothing asked for agent forwarding on this session. */
         ret = WS_BAD_ARGUMENT;
--- a/src/agent.c
+++ b/src/agent.c
@@ -1731,6 +1731,84 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
 }
 
 
+int wolfSSH_AGENT_ChannelOpen(WOLFSSH* ssh)
+{
+    WOLFSSH_AGENT_CTX* newAgent = NULL;
+    WOLFSSH_CHANNEL* newChannel = NULL;
+    int ret = WS_SUCCESS;
+    /* wolfSSH_accept() clears only want-read/want-write/auth-pending, so a
+     * WS_BAD_ARGUMENT latched by a poll kills the handshake. */
+    int recordError = 0;
+
+    WLOG_ENTER();
+
+    if (ssh == NULL)
+        ret = WS_SSH_NULL_E;
+    else if (ssh->ctx->side != WOLFSSH_ENDPOINT_SERVER) {
+        /* Server side only. wolfSSH_connect() sets ssh->agent too, so the
+         * checks below would report a channel a client never opened. */
+        ret = WS_BAD_ARGUMENT;
+    }
+    else if (!ssh->useAgent) {
+        /* Nothing asked for agent forwarding on this session. */
+        ret = WS_BAD_ARGUMENT;
+    }
+    else if (ssh->agent == NULL) {
+        /* Nothing else sets ssh->agent, so a NULL one means "not opened
+         * yet". Idempotent, so a poll cannot open a second channel. */
+        WLOG(WS_LOG_AGENT, "Starting agent channel");
+
+        newAgent = wolfSSH_AGENT_new(ssh->ctx->heap);
+        if (newAgent == NULL)
+            ret = WS_MEMORY_E;
+
+        if (ret == WS_SUCCESS) {
+            newChannel = ChannelNew(ssh, ID_CHANTYPE_AUTH_AGENT,
+                    ssh->ctx->windowSz, ssh->ctx->maxPacketSz);
+            if (newChannel == NULL)
+                ret = WS_MEMORY_E;
+        }
+
+        if (ret == WS_SUCCESS) {
+            recordError = 1;
+            ret = SendChannelOpenSession(ssh, newChannel);
+
+            if (ret < WS_SUCCESS
+                    && ret != WS_WANT_WRITE && ret != WS_WANT_READ) {
+                ChannelDelete(newChannel, ssh->ctx->heap);
+            }
+            else {
+                /* Publish on a queued open too, so a retry takes the
+                 * already-open path rather than opening a second. */
+                ChannelAppend(ssh, newChannel);
+                newAgent->channel = newChannel->channel;
+                ssh->agent = newAgent;
+                newAgent = NULL;
+                if (ssh->ctx->agentCb) {
+                    ssh->ctx->agentCb(WOLFSSH_AGENT_LOCAL_SETUP,
+                            ssh->agentCbCtx);
+                }
+            }
+        }
+
+        if (newAgent != NULL)
+            wolfSSH_AGENT_free(newAgent);
+    }
+    else if (wolfSSH_OutputPending(ssh)) {
+        /* Any queued output, not just this open. Flush it rather than
+         * report a success the peer hasn't seen. */
+        recordError = 1;
+        ret = wolfSSH_SendPacket(ssh);
+    }
+
+    if (recordError)
+        ssh->error = ret;
+
+    WLOG_LEAVE(ret);
+    return ret;
+}
+
+
 int wolfSSH_AGENT_worker(WOLFSSH* ssh)
 {
     int ret = WS_SUCCESS;
--- a/wolfssh/agent.h
+++ b/wolfssh/agent.h
@@ -187,11 +187,12 @@ WOLFSSH_API int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled);
  * this instead. Opens one channel, then flushes what of the open is queued.
  * Returns WS_SUCCESS, WS_BAD_ARGUMENT before the peer asks or on a client
  * session, WS_WANT_READ or WS_WANT_WRITE while output is still queued,
+ * WS_FATAL_ERROR with WS_DISCONNECT in ssh->error once the session is over,
  * WS_SSH_NULL_E, WS_MEMORY_E, or whatever the send reports. WS_SUCCESS says
  * the open went out, not that the peer took it; a refusal reaches the
  * channel-open-fail callback.
- * Only the send records in ssh->error, so a poll ahead of the peer's request
- * leaves the session fit for wolfSSH_accept(). */
+ * Only that and the send record in ssh->error, so a poll ahead of the peer's
+ * request leaves the session fit for wolfSSH_accept(). */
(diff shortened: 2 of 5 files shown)


src/agent.c
```diff
@@ -1731,6 +1731,71 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+            }
+        }
+
+        if (newAgent != NULL)
+            wolfSSH_AGENT_free(newAgent);
+    }
+
+    if (ssh != NULL)
+        ssh->error = ret;
+
```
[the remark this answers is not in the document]
ejohnstown: Fixed. When the agent is already published and `wolfSSH_OutputPending()` is
true, the call now flushes through `wolfSSH_SendPacket()` and returns that
result, so a poll keeps seeing `WS_WANT_WRITE` -- and `ssh->error` keeps
holding it -- until the open is actually on the wire. Routed through
`wolfSSH_SendPacket()` rather than the raw flush so the deferred
`HighwaterCheck()` still runs on the completing send.

Covered in `tests/regress.c`: the harness forces one would-block, so the
first call returns `WS_WANT_WRITE` with nothing sent, and the second must
finish the send, leave `channelListSz` at 1, and put a `CHANNEL_OPEN` on the
wire. Removing the flush branch fails that test on the last two assertions.

wolfssh/agent.h
```diff
@@ -181,6 +181,14 @@ WOLFSSH_API int wolfSSH_CTX_set_agent_cb(WOLFSSH_CTX* ctx,
 WOLFSSH_API int wolfSSH_set_agent_cb_ctx(WOLFSSH* ssh, void* ctx);
 WOLFSSH_API int wolfSSH_CTX_AGENT_enable(WOLFSSH_CTX* ctx, byte isEnabled);
 WOLFSSH_API int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled);
+/* Server side. Opens the auth-agent@openssh.com channel back to the client
+ * once an auth-agent-req@openssh.com request has set the session up for it.
+ * wolfSSH_accept() does this itself on the default path; an application that
+ * drives its own channels returns from accept() before that point and calls
+ * this instead. Idempotent, so it is safe to poll while waiting for the
+ * peer's request. Returns WS_SUCCESS, or WS_BAD_ARGUMENT when the session
+ * never asked for agent forwarding. */
```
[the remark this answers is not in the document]
ejohnstown: Fixed. The comment now names the whole surface: `WS_SUCCESS`,
`WS_BAD_ARGUMENT` before the peer asks, `WS_WANT_READ`/`WS_WANT_WRITE` with
the open queued, `WS_SSH_NULL_E`, and `WS_MEMORY_E`. It also states that the
call opens one channel and then flushes whatever of the open is still
queued, so the retry contract sits where the caller reads it.

src/agent.c
```diff
@@ -1731,6 +1731,76 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+            wolfSSH_AGENT_free(newAgent);
+    }
+    else if (wolfSSH_OutputPending(ssh)) {
+        /* The open is framed but not on the wire. Flush it rather than
+         * report a success the peer hasn't seen. */
+        ret = wolfSSH_SendPacket(ssh);
+    }
+
+    if (ssh != NULL)
+        ssh->error = ret;
```
ejohnstown, line 1812: Fixed, and confirmed first: a poll on the harness at
`ACCEPT_SERVER_USERAUTH_SENT` returned `WS_BAD_ARGUMENT`, left
`ssh->error` at -1002, and the next `wolfSSH_accept()` answered -1030.

`ssh->error` is now recorded only by the branches that reach the
transport -- the open and the flush of a queued open. The
not-requested-yet return and the already-open-with-nothing-pending
return leave it alone, so `wolfSSH_accept()` still finds a clean
session. A queued open keeps latching `WS_WANT_WRITE`, so a caller
polling through the retry still sees it.

`TestAgentChannelOpenWithoutRequest` in `tests/regress.c` now asserts
`ssh->error` is untouched after the refused poll and that the following
`wolfSSH_accept()` returns the retryable `WS_WANT_READ` rather than
`WS_INVALID_STATE_E`. Restoring the unconditional stamp fails it.

One thing beyond the report: the header comment listed a closed set of
return codes that the send paths do not honor, so it now says the send
of the open may report something else.

src/agent.c
```diff
@@ -1731,6 +1731,83 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+     * before the peer asks must leave ssh->error alone; wolfSSH_accept()
+     * clears only want-read/want-write/auth-pending, so a latched
+     * WS_BAD_ARGUMENT there kills the handshake with WS_INVALID_STATE_E. */
+    int recordError = 0;
+
+    WLOG_ENTER();
+
+    if (ssh == NULL)
+        ret = WS_SSH_NULL_E;
+    else if (!ssh->useAgent) {
```
ejohnstown, line 1759: Split verdict. The disconnect half is real and fixed here; the rekey half is
real but not this PR's, and the recommended fix is wrong in part.

**Disconnect: confirmed, then fixed.** Verified first on the `regress.c`
mem-IO harness -- with `ssh->disconnected` set, the call returned `WS_SUCCESS`
and put 48 bytes of `SSH_MSG_CHANNEL_OPEN` on the wire. `wolfSSH_accept()`
gates this for the open it drives, so lifting the open out into a public entry
point is what opened the door.

`wolfSSH_AGENT_ChannelOpen()` now answers `WS_DISCONNECT` on a session that is
over: no channel created, nothing sent, and no flush of an open queued before
the disconnect either -- those bytes belong to a session that is over, the
same rule `wolfSSH_shutdown()` applies to everything but its own queued
disconnect. `SendAfterDisconnect()` is promoted to `WOLFSSH_LOCAL` rather than
reimplemented, so `agent.c` uses the same helper as every other public sender.
`TestAgentChannelOpenAfterDisconnect` and
`TestAgentChannelOpenQueuedThenDisconnect` cover both paths; neutering the
guard fails the first on `0 != -1098`.

**Rekey: not introduced here.** "unlike every other public sending API in
`src/ssh.c`" does not hold on this axis. On the same harness with
`WOLFSSH_PEER_IS_KEYING` set, `wolfSSH_ChannelFwdNewLocal()` also puts message
ID 90 on the wire and returns a channel. `SendChannelOpen()` carries no keying
check and neither does anything below it -- `PreparePacket()`,
`BundlePacket()` and `wolfSSH_SendPacket()` are unconditional. So all three
channel-open APIs are equally exposed, and have been since long before this
branch. `SendChannelData()` and `SendChannelExtendedData()` do have the check;
the open path never got it.

The guard belongs in `SendChannelOpen()`, where one check covers all three
callers and `agent.c` never has to know about keying. That changes the return
of two shipped APIs to `WS_REKEYING` where they previously reported success,
so it is its own commit rather than a rider on this PR. Tracked separately.

One correction to the recommended fix: do not gate the flush on `isKeying`.
Bytes already bundled were framed under the pre-rekey keys and sequence
number, so refusing to push them strands the peer's stream --
`FlushQueuedOutput()` in `src/ssh.c` makes the same point about RFC 4253
section 11.1. "before creating the channel or flushing" is right for the
disconnect case and wrong for the rekey one.

src/agent.c
```diff
@@ -1731,6 +1731,89 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+        /* The session is over, so neither a new open nor the flush of one
+         * queued before the disconnect may go out. RFC 4253 section 11.1.
+         * SendAfterDisconnect() has already recorded WS_DISCONNECT. */
+        ret = WS_DISCONNECT;
+    }
+    else if (!ssh->useAgent) {
+        /* Nothing asked for agent forwarding on this session. */
+        ret = WS_BAD_ARGUMENT;
+    }
+    else if (ssh->agent == NULL) {
```
ejohnstown, line 1763: Confirmed, and the close half with it -- but the recommended fix is wrong in
part, so this one is filed rather than fixed here.

**Verified.** `ssh->agent` is set in exactly one place and cleared in exactly
one, `wolfSSH_free()`. `DoChannelOpenFail()` and `DoChannelClose()` both call
`ChannelRemove()` and leave it standing, so a poll after either answers
`WS_SUCCESS` for a channel that is gone, and `ssh->agent->channel` still names
a removed id that `wolfSSH_AGENT_Relay()` and `wolfSSH_AGENT_worker()` would
work against.

**Why not the suggested fix.** Keying the already-open test on an
`ID_CHANTYPE_AUTH_AGENT` channel in `ssh->channelList` opens two new holes:

- the next poll then reaches `ssh->agent = newAgent` with the old context
  still set, leaking one `WOLFSSH_AGENT_CTX` per retry. The
  `wolfSSH_AGENT_free(ssh->agent)` that used to guard that assignment inside
  `wolfSSH_accept()` was dropped when the NULL test made it unreachable.
- it cannot tell a refusal from a graceful close. `DoChannelClose()` is the
  normal end of the agent channel, so a poll after a clean close would open a
  second one -- re-forwarding the agent rather than retrying a failed open.
  Only the `DoChannelOpenFail()` case should become retryable.

Clearing `ssh->agent` where the agent's own open fails is the real fix, and it
needs `DoChannelOpenFail()` to know the removed channel was the agent's -- it
has the channel and the callback but no channel-type test today -- plus a
decision on whether the `WOLFSSH_AGENT_LOCAL_SETUP` the open fired gets a
matching teardown. That is more than this PR carries, so it is tracked
separately.

**What this PR does change.** The header now states the half a caller can act
on: `WS_SUCCESS` says the open went out, not that the peer took it, and a
refusal arrives at the channel-open-fail callback.

src/agent.c
```diff
@@ -1731,6 +1731,91 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+        }
+
+        if (newAgent != NULL)
+            wolfSSH_AGENT_free(newAgent);
+    }
+    else if (wolfSSH_OutputPending(ssh)) {
+        /* Any queued output, not just this open. Flush it rather than
+         * report a success the peer hasn't seen. */
+        recordError = 1;
+        ret = wolfSSH_SendPacket(ssh);
```
ejohnstown, line 1808: Confirmed the state, but a retry cannot work, so this is not fixed here.

`BundlePacket()` increments `ssh->seq` when it frames the packet, before
anything reaches the socket, and the discard path never gives it back. On
the `regress.c` mem-IO harness `seq` goes from 0 to 1 for the open that is
lost and stays at 1. A rebuilt open would go out under sequence 2 while the
peer still expects 1. Post-userauth traffic is always under a cipher and the
peer's MAC covers its expected sequence number, so the next packet this side
sends fails verification and the peer disconnects. Tearing down the agent
and reopening would fire a second `WOLFSSH_AGENT_LOCAL_SETUP` and then die
the same way.

The caller was already told: the poll that ran the failed flush returned
`WS_SOCKET_ERROR_E` and left it in `ssh->error`, where the later
`WS_SUCCESS` does not clear it, and `wolfSSH_accept()` refuses the session
with `WS_INVALID_STATE_E`.

Every sender that queues on `WS_WANT_WRITE` is left in this shape by a later
general error, forwards included, and has been since before this branch.
Whether a general send error should mark the session over rather than
discard and carry on is a send-path question, tracked separately.

philljj (approved):

PR #1231, opened 2026-09-03
SCP: let the application start the transfer

ejohnstown:
An application that binds an "scp ..." command to a channel itself has no
way to run the transfer; wolfSSH_accept() did it through a WS_SCP_INIT
re-entry only that state machine can drive.

- Add wolfSSH_SCP_accept(), a wrapper over DoScpRequest() reporting
  WS_SCP_COMPLETE for any non-negative result, as accept() does.
- Report a receive-side want as itself so the caller can retry, and clear
  a stale want on entry the way the other re-entrant entry points do.

### DIFF
--- a/src/wolfscp.c
+++ b/src/wolfscp.c
@@ -883,6 +883,40 @@ int DoScpSource(WOLFSSH* ssh)
     return ret;
 }
 
+/* Contract is in wolfssh/wolfscp.h. */
+int wolfSSH_SCP_accept(WOLFSSH* ssh)
+{
+    int ret;
+
+    if (ssh == NULL)
+        return WS_BAD_ARGUMENT;
+
+    /* Clear a want left by the previous call so the retry starts clean,
+     * the way the other re-entrant entry points do. */
+    if (ssh->error == WS_WANT_READ || ssh->error == WS_WANT_WRITE)
+        ssh->error = WS_SUCCESS;
+
+    ret = DoScpRequest(ssh);
+
+    if (ret >= WS_SUCCESS) {
+        /* The tail of DoScpRequest() passes a read count through, so treat
+         * anything non-negative as done the way wolfSSH_accept() does. */
+        ret = WS_SCP_COMPLETE;
+    }
+    else {
+        /* A non-blocking want on a read path surfaces as a generic error
+         * with the want recorded in ssh->error (see GetInputData), so
+         * report it as the want the caller is told to retry on. */
+        int err = wolfSSH_get_error(ssh);
+
+        if (err == WS_WANT_READ || err == WS_WANT_WRITE)
+            ret = err;
+    }
+
+    return ret;
+}
+
+
 int DoScpRequest(WOLFSSH* ssh)
 {
     int ret = WS_SUCCESS;
--- a/src/wolfscp.c
+++ b/src/wolfscp.c
@@ -903,14 +903,11 @@ int wolfSSH_SCP_accept(WOLFSSH* ssh)
          * anything non-negative as done the way wolfSSH_accept() does. */
         ret = WS_SCP_COMPLETE;
     }
-    else {
-        /* A non-blocking want on a read path surfaces as a generic error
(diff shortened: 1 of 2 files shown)


src/wolfscp.c
```diff
@@ -883,6 +883,40 @@ int DoScpSource(WOLFSSH* ssh)
     return ret;
 }
 
+/* Contract is in wolfssh/wolfscp.h. */
+int wolfSSH_SCP_accept(WOLFSSH* ssh)
+{
+    int ret;
+
```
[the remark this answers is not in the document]
ejohnstown: Defer.

src/wolfscp.c
```diff
@@ -883,6 +883,40 @@ int DoScpSource(WOLFSSH* ssh)
...
+         * anything non-negative as done the way wolfSSH_accept() does. */
+        ret = WS_SCP_COMPLETE;
+    }
+    else {
+        /* A non-blocking want on a read path surfaces as a generic error
+         * with the want recorded in ssh->error (see GetInputData), so
+         * report it as the want the caller is told to retry on. */
+        int err = wolfSSH_get_error(ssh);
+
+        if (err == WS_WANT_READ || err == WS_WANT_WRITE)
```
ejohnstown: Confirmed. A short send that SendChannelData() accepts leaves WS_WANT_WRITE in ssh->error with nothing clearing it, and the abort path in SendScpConfirmation() returns WS_FATAL_ERROR right after that send, so the wrapper reported the abort as a want and a retry re-sent the confirmation. Reproduced on a memory transport: one blocked write, two abort confirmations on the wire.

Fixed differently from the suggestion. ScpStreamRead() already passes a write want through by value; it is only the read want that GetInputData() hides behind WS_FATAL_ERROR. The wrapper now substitutes ssh->error only when the result is WS_FATAL_ERROR and the want is WS_WANT_READ, so a stale write want can no longer replace a terminal result.

ejohnstown: Confirmed, with one correction to the path. wolfSSH_stream_read() clears ssh->error on entry, so the stale want reaching WS_SCP_BAD_MSG_E comes from the window adjust that read itself queues, not from the earlier SendPacketFlush(). Reproduced: a blocked adjust after a malformed message came back as WS_WANT_WRITE instead of WS_SCP_BAD_MSG_E.

The suggested fix, substituting only on WS_FATAL_ERROR, covers this case but not the abort case in the other thread, because SendScpConfirmation() returns WS_FATAL_ERROR for an abort. The wrapper now substitutes only when the result is WS_FATAL_ERROR and the want is WS_WANT_READ. Write wants already come back by value from the SCP state machine, so nothing is lost on that side.

wolfssh/wolfscp.h
```diff
@@ -158,6 +158,17 @@ WOLFSSH_API int   wolfSSH_SCP_to(WOLFSSH* ssh, const char* src,
...
+ * command is already bound. This is the same work wolfSSH_accept() does
+ * through its WS_SCP_INIT re-entry, exposed so an application can start the
+ * transfer itself; use one or the other, not both. Call it once
+ * wolfSSH_accept() has returned and the exec channel-request callback has
+ * reported an SCP command, not from inside that callback.
+ *
+ * Returns WS_SCP_COMPLETE when the transfer is done. On a non-blocking
+ * socket it returns WS_WANT_READ or WS_WANT_WRITE with the transfer part
+ * done; call it again on the same session until it completes. */
+WOLFSSH_API int   wolfSSH_SCP_accept(WOLFSSH* ssh);
```
philljj, line 171: there should be an example or unit test calling this

philljj (changes requested):
looks fine, just suggest adding a regression test or example

wolfssh/wolfscp.h
philljj: will be added in https://github.com/wolfSSL/wolfssh/pull/1234

philljj (approved):

PR #1233, opened 2026-09-03
ssh: add opt-in application-driven channels

ejohnstown:
A server that wants to own its channels had no way to get them: accept() ran
the session state machine to the end, and a shell, exec or subsystem request
with no callback registered was granted regardless.

- Add wolfSSH_CTX_SetAppChannels() and wolfSSH_SetAppChannels(), off by
  default. On, accept() returns once the user is authenticated and a
  session request with no callback behind it is refused.
- Keep the stop state out of the pending-send advance, and stop early
  only while the session is short of that state.
- Teach wolfSSH_SFTP_accept() that the mode parks accept() short of an
  established session, and serve the built-in server only on a channel
  the application granted sftp on.
- Match the sftp subsystem name whole, by the parsed wire length as well
  as the bytes, in that gate and in accept()'s divert (F-11665).
- regress.c drives both modes, unit.c checks DoChannelRequest() refuses
  an uncallbacked request once the pivot is on.


src/ssh.c
```diff
@@ -4931,6 +4913,32 @@ int wolfSSH_CTX_SetChannelReqSubsysCb(WOLFSSH_CTX* ctx,
...
+
+
+int wolfSSH_SetAppChannels(WOLFSSH* ssh, byte enable)
+{
+    int ret = WS_SSH_NULL_E;
+
+    if (ssh != NULL) {
+        ssh->appChannels = (enable != 0);
+        ret = WS_SUCCESS;
+    }
```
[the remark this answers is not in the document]
ejohnstown: Confirmed, and the header was the part that was wrong. wolfSSH_accept() already guards the late enable -- stopState only takes the user-auth value while acceptState is at or below ACCEPT_SERVER_USERAUTH_SENT -- but DoChannelRequest() reads the flag when the request arrives, with no such guard. Driving the same shell request twice over an established session shows it: CHANNEL_SUCCESS before the late enable, CHANNEL_FAILURE (WS_CHANOPEN_FAILED at the client) after.

I did not take the suggested gate, for two reasons. The threshold does not line up with the one accept() uses: an enable at ACCEPT_SERVER_CHANNEL_ACCEPT_SENT passes a >= ACCEPT_CLIENT_SESSION_ESTABLISHED check, mutates the flag, and still cannot move where accept() stops, which is the half-applied state the gate is meant to prevent. And a setter that returns WS_SUCCESS while silently doing nothing is a worse contract than one that works -- refusing a request no callback can service is the right answer whenever the application has taken over, whenever the flag went on.

So the sentence now says the flag reaches the requests that follow and that what it cannot do is move where accept() returns. TestAppChannelsLateEnableReturns() grew a second half that sends a shell request through the transport in both modes to pin it; gating the setter as suggested makes that assertion fail.

wolfssh/agent.h
```diff
@@ -181,6 +181,14 @@ WOLFSSH_API int wolfSSH_CTX_set_agent_cb(WOLFSSH_CTX* ctx,
 WOLFSSH_API int wolfSSH_set_agent_cb_ctx(WOLFSSH* ssh, void* ctx);
 WOLFSSH_API int wolfSSH_CTX_AGENT_enable(WOLFSSH_CTX* ctx, byte isEnabled);
 WOLFSSH_API int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled);
+/* Server side. Opens the auth-agent@openssh.com channel back to the client
+ * once an auth-agent-req@openssh.com request has set the session up for it.
+ * wolfSSH_accept() does this itself on the default path; an application that
+ * drives its own channels returns from accept() before that point and calls
+ * this instead. Idempotent, so it is safe to poll while waiting for the
+ * peer's request. Returns WS_SUCCESS, or WS_BAD_ARGUMENT when the session
+ * never asked for agent forwarding. */
```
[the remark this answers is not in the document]
ejohnstown: Agreed, and the retryable codes were the part that mattered. The comment tells a caller it is safe to poll, then lists a return set that makes WS_WANT_WRITE look fatal -- and a queued open is exactly what the polling path produces. wolfSSH_AGENT_ChannelOpen(NULL) returning WS_SSH_NULL_E confirmed the NULL case too.

The comment now names WS_SSH_NULL_E, WS_MEMORY_E and both want codes alongside the two that were there, and says the want codes are a retry that finds the already-published agent rather than opening a second channel.

src/agent.c
```diff
@@ -1731,6 +1731,71 @@ int wolfSSH_AGENT_enable(WOLFSSH* ssh, byte isEnabled)
...
+                }
+            }
+        }
+
+        if (newAgent != NULL)
+            wolfSSH_AGENT_free(newAgent);
+    }
+
+    if (ssh != NULL)
+        ssh->error = ret;
```
ejohnstown, line 1816: Agreed, and it is fixed -- though not on this branch. The agent commits
went to master with #1230, so `src/agent.c` is no longer in this PR's diff.

`wolfSSH_AGENT_ChannelOpen()` now writes `ssh->error` only around the two
paths that actually send: the initial `SendChannelOpenSession()`, and the
flush of an open that was already published. The `WS_BAD_ARGUMENT` returns
a poll produces -- before the peer has asked, and on a client session --
leave `ssh->error` alone, so polling ahead of the request no longer leaves
the session unfit for `wolfSSH_accept()`. The disconnect return needs no
recording of its own, since `SendAfterDisconnect()` already puts
`WS_DISCONNECT` there. `agent.h` now says which returns record and which
do not.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,12 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
     if (ssh->error == WS_WANT_READ || ssh->error == WS_WANT_WRITE)
         ssh->error = WS_SUCCESS;
 
-    /* check accept is done, if not call wolfSSH accept */
-    if (ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED) {
+    /* check accept is done, if not call wolfSSH accept. In
+     * application-driven mode accept() parks at ACCEPT_SERVER_USERAUTH_SENT
+     * and never advances, so that state counts as done here. */
+    if (ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED
+            && !(ssh->appChannels
```
ejohnstown: Confirmed, and the first fix was not enough on its own.

Gating on the session type and command still served a refused request.
`DoChannelRequest()` sets `channel->sessionType` and `channel->command`
before it decides, and leaves both set on a refusal, so a `subsystem sftp`
answered CHANNEL_FAILURE still satisfied the gate. Both refusals reached
it: no subsystem callback registered, where app channels refuse the request
themselves, and a callback that returns non-zero. Each one returned
`WS_SFTP_COMPLETE` and put an SFTP VERSION on the channel.

The gate now asks for the grant rather than inferring it. A channel records
whether a shell, exec or subsystem request was answered CHANNEL_SUCCESS,
and `wolfSSH_SFTP_accept()` requires that alongside the type and the
command. Both refusals now return `WS_INVALID_STATE_E` with nothing sent,
and regress.c covers them either way.

The same shape survives in `wolfSSH_accept()`'s own divert, which reads only
the type and command. That one predates this branch and is addressed by the
F-8852 work in #1235, which puts the type and command back when a request is
refused.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,27 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
...
+         * callback: serve only a session channel it granted sftp on. The
+         * request having named sftp is not enough, so this asks for the
+         * grant as well -- unlike wolfSSH_accept()'s divert, which reads
+         * only the type and command. */
+        const WOLFSSH_CHANNEL* channel = ssh->channelList;
+
+        if (channel == NULL || !channel->sessionGranted
+                || channel->sessionType != WOLFSSH_SESSION_SUBSYSTEM
+                || channel->command == NULL
+                || WSTRNCMP(channel->command, "sftp", 4) != 0) {
```
ejohnstown: Confirmed and fixed. Reproduced it directly: a `subsystem sftpx` granted by the callback returned `WS_SFTP_COMPLETE` from `wolfSSH_SFTP_accept()` and put a `CHANNEL_DATA` VERSION on the wire. The gate now compares the whole name with `WSTRCMP`, and there is a regress case for a granted prefix name.

One scope note: the same four-byte compare in the `wolfSSH_accept()` divert in `src/ssh.c` is pre-existing on master and is not this gate, so it is left alone here.

tests/regress.c
```diff
@@ -3530,6 +3727,204 @@ static void TestChannelReqSubsysCallbackRuns(void)
...
+    channel = SeedUnconfirmedChannel(&harness);
+    AssertIntEQ(ChannelUpdatePeer(channel, 5, 1024, 1024), WS_SUCCESS);
+    channel->openConfirmed = 1;
+
+    inSz = BuildSessionRequestPacket(channel->channel, "subsystem", "sftp",
+            in, sizeof(in));
+    RepointHarnessInput(&harness, in, inSz);
+    AssertIntEQ(DoReceive(harness.ssh), WS_SUCCESS);
+    /* Either way the peer is told the subsystem was refused. */
+    AssertIntEQ(ParseMsgId(harness.io.out, harness.io.outSz),
```
ejohnstown, line 4298: Confirmed and fixed. Checked the claim by deleting the callback registration from the `rejectVia != 0` case: the test still passed, so it was asserting nothing that separated it from `rejectVia == 0`. It now asserts the call count each case expects, 1 with a callback registered and 0 without.

Worth noting for anyone tracing this: a mutation that skips the callback in app-channels mode is caught, but by `SeedAppChannelsSession()` in a different test, which is why the gap looked covered.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,28 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
...
+                || channel->sessionType != WOLFSSH_SESSION_SUBSYSTEM
+                || channel->command == NULL
+                || WSTRCMP(channel->command, "sftp") != 0) {
+            WLOG(WS_LOG_SFTP, "No sftp subsystem granted on the session");
+            return WS_INVALID_STATE_E;
+        }
+    }
     /* check accept is done, if not call wolfSSH accept */
-    if (ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED) {
+    else if (ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED) {
```
ejohnstown, line 1412: Confirmed and fixed. The chain is `wolfSSH_accept()` returning `WS_SUCCESS` with `channelList == NULL`, then `wolfSSH_stream_peek()` returning `WS_BAD_ARGUMENT`, which `wolfSSH_SFTP_buffer_read()` maps to `WS_FATAL_ERROR`.

Fixed differently from the suggestion: the gate no longer has an `acceptState` bound, so in this mode `wolfSSH_SFTP_accept()` never reaches the branch that calls `wolfSSH_accept()`. Re-checking after `accept()` succeeds would still run the whole handshake and set `ssh->channelName` before refusing, which is the part that breaks the documented contract.

`regress.c` now calls it at `ACCEPT_BEGIN` with the mode on and asserts `WS_INVALID_STATE_E`, nothing sent, the state unmoved and `channelName` unset.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,28 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
     if (ssh->error == WS_WANT_READ || ssh->error == WS_WANT_WRITE)
         ssh->error = WS_SUCCESS;
 
+    if (ssh->appChannels
```
ejohnstown: The observation is right, and measurable: with `ssh->error` at `WS_WANT_READ`, a refused call returns `WS_INVALID_STATE_E` and leaves `ssh->error` at `WS_SUCCESS`.

Addressed in the header rather than the code. `ssh.h` now says the call returns `WS_INVALID_STATE_E` without recording an error, and says outright that a pending want-read or want-write is still cleared.

The clear at the top of `wolfSSH_SFTP_accept()` is that function's entry behaviour on every path. Exempting one early return would make it non-uniform in order to preserve a status the caller cannot act on: the refusal is not retryable, so there is nothing for a want-read to resume.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,28 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
...
+            && ssh->acceptState >= ACCEPT_SERVER_USERAUTH_SENT
+            && ssh->acceptState < ACCEPT_CLIENT_SESSION_ESTABLISHED) {
```
ejohnstown: Confirmed, and reproduced: with the mode on, an established session whose channel was granted a shell answered an SFTP INIT with a VERSION and returned `WS_SFTP_COMPLETE`.

Fixed as suggested -- the gate is now `if (ssh->appChannels)` with no `acceptState` bound in either direction. Two tests: an established session granted a shell is refused, and one granted sftp is still served, so the wider gate cannot quietly over-tighten.

Worth noting the underlying looseness is older than this PR. Master's `wolfSSH_SFTP_accept()` has no grant check at all, so an established session serves SFTP whatever the channel was granted. What this PR added is the `ssh.h` promise that it serves only a granted sftp, which is what made the gap a contract violation.

tests/regress.c
```diff
@@ -1486,6 +1486,203 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo,
...
+    (void)channel;
+    (void)ctx;
+    appChannelsShellReqCount++;
+    return 0;
+}
+
+/* Drive an application-driven server: wolfSSH_accept() is expected to return
+ * at userauth, so the channel open and the shell request are answered by
+ * wolfSSH_worker() calls the application makes itself. */
+static void RunAppChannelsHandshake(KexReplyHarness* harness,
```
ejohnstown, line 1503: Correct. Instrumented the block: it is never entered by `regress`, `unit`, `api` or `testsuite`, so the guard had no coverage at all.

It is load-bearing. Replacing the `stopState` test with a constant makes a re-entered `accept()` step past the stop, and since `ACCEPT_CLIENT_SESSION_ESTABLISHED` has no case in the switch, the loop then spins on it.

The suggested repro would not have caught that. With auth-success output pending the state is `ACCEPT_CLIENT_USERAUTH_DONE`, where advancing to the stop is the wanted behaviour and the guard never fires -- that case returns `WS_SUCCESS` parked at `ACCEPT_SERVER_USERAUTH_SENT` with and without the guard. The test added instead re-enters `accept()` already parked at the stop with output queued; it passes now and fails with the guard removed.

tests/regress.c
```diff
@@ -1486,6 +1486,208 @@ static void AssertHandshakeRejectsMutatedReply(const char* keyAlgo,
...
+    AssertIntEQ(wolfSSH_worker(harness.client, NULL), WS_SUCCESS);
+
+    AssertIntEQ(wolfSSH_SetAppChannels(harness.server, 1), WS_SUCCESS);
+    AssertIntEQ(wolfSSH_accept(harness.server), WS_SUCCESS);
+    AssertIntEQ(harness.server->acceptState,
+            ACCEPT_CLIENT_SESSION_ESTABLISHED);
+
+    /* Same request, same session, mode now on: refused instead. */
+    idx = 0;
+    AssertIntEQ(wolfSSH_TestDoChannelRequest(harness.server, payShell,
```
ejohnstown, line 1683: I don't think this one holds. The premise is that wolfSSH enforces one program per channel, so the second request would be refused whether or not late enable works. It doesn't: `DoChannelRequest()`'s shell branch sets `sessionType` and calls the callback with no check for a prior request on that channel.

Measured it rather than reading it -- ran the test with the `wolfSSH_SetAppChannels()` call deleted and nothing else changed. The second shell request is **granted**: `wolfSSH_worker()` returns 0 and `wolfSSH_get_error()` is 0, not `WS_CHANOPEN_FAILED`. So the rejection the test asserts depends only on `appChannels`, which is what it is there to pin.

The RFC 4254 6.5 point is fair on its own terms -- wolfSSH does not enforce the one-program rule, and that is arguably worth a look. It is a separate gap from this test, and its absence is what makes the test valid rather than ineffective.

src/internal.c
```diff
@@ -13287,11 +13300,18 @@ static int DoChannelRequest(WOLFSSH* ssh,
         *idx = len;
     }
 
+    /* Record the answer, not the ask: sessionType and command are set before
+     * the reject decision and stay set on a refusal, so they cannot say
+     * whether the session was granted. Set even without a wantReply, which
+     * changes only whether the peer is told. */
+    if (sessionReq && channel != NULL)
+        channel->sessionGranted = (ret == WS_SUCCESS && !rej);
```
ejohnstown: Confirmed, and fixed. Also introduced here, more clearly than the report says: master's only post-callback dereferences of `channel` are the `WLOG(..., channel->command)` lines in the exec and subsystem branches, and `WLOG` is `WC_DO_NOTHING` without `DEBUG_WOLFSSH`. So a release build of master has no post-callback dereference at all, and the `sessionGranted` write is the first that is unconditional, always compiled and a write.

Fixed as suggested: the channel is looked up again by id after the callback, and the grant is recorded only if it is still there. Nothing else in the function needed changing, since the reply already goes by id. The `channel != NULL` guard that was there did not help, the pointer being dangling rather than null.

`regress.c` now drives a shell callback that calls `wolfSSH_ChannelFree()` on its own channel. The request ends in `WS_INVALID_CHANID` with the channel gone and no reply sent, rather than writing through the freed pointer. An ASan build reports `heap-use-after-free` at that write with the fix reverted, and is clean with it in.

src/internal.c
```diff
@@ -13139,6 +13144,10 @@ static int DoChannelRequest(WOLFSSH* ssh,
             if (ssh->ctx->channelReqExecCb) {
                 rej = ssh->ctx->channelReqExecCb(channel, ssh->channelReqCtx);
             }
+            else {
+                rej = ssh->appChannels;
+            }
+            sessionReq = 1;
             ssh->clientState = CLIENT_DONE;
 
             WLOG(WS_LOG_DEBUG, "  command = %s", channel->command);
```
ejohnstown: Fixed. Cleaned up both locations, and a third.

src/wolfsftp.c
```diff
@@ -1383,8 +1383,31 @@ int wolfSSH_SFTP_accept(WOLFSSH* ssh)
...
+         * request having named sftp is not enough, so this asks for the
+         * grant as well -- unlike wolfSSH_accept()'s divert, which reads
+         * only the type and command. The name matches whole: sftpx is
+         * some other subsystem. */
+        const WOLFSSH_CHANNEL* channel = ssh->channelList;
+
+        if (channel == NULL || !channel->sessionGranted
+                || channel->sessionType != WOLFSSH_SESSION_SUBSYSTEM
+                || channel->command == NULL
+                || WSTRCMP(channel->command, "sftp") != 0) {
```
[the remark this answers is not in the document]
ejohnstown: Confirmed and fixed. Reproduced it with the subsystem name on the wire as the
nine bytes `sftp\0evil`: the callback was called, saw `sftp`, granted it, and
`wolfSSH_SFTP_accept()` returned `WS_SFTP_COMPLETE` -- the built-in server took
the session.

Fixed at the parse rather than the compare. `DoChannelRequest()` was passing
NULL for `GetStringAlloc()`'s length out-param, so the wire length was gone
before anything could check it; it now keeps it in `channel->commandSz`, and
the name checks require the length as well as the bytes. There is a regress
case that sends the NUL name and expects `WS_INVALID_STATE_E`; dropping the
length check alone turns it back into `WS_SFTP_COMPLETE`.

One correction to the report: the callback could not have caught this either.
The length was already discarded when the callback ran, so it was handed the
same `sftp` C string the gate saw and had nothing to refuse on. Nothing was
approved in error, and a peer that can send `sftp\0evil` can equally send
`sftp`.

Scope note, since an earlier thread here said the opposite: the four-byte
compare in `wolfSSH_accept()`'s divert is fixed in the same commit rather than
left alone, now that there is a length to check. That closes the prefix half on
the built-in path as well, which is F-11665.

src/internal.c
```diff
@@ -13120,39 +13121,62 @@ static int DoChannelRequest(WOLFSSH* ssh,
...
+            ret = GetStringAlloc(ssh->ctx->heap,
+                    &channel->command, &channel->commandSz,
                     buf, len, &begin);
+            if (ret == WS_SUCCESS)
+                WLOG(WS_LOG_DEBUG, "  subsystem = %s", channel->command);
+            else
+                WLOG(WS_LOG_DEBUG, "  subsystem = %s", "<bad value>");
             channel->sessionType = WOLFSSH_SESSION_SUBSYSTEM;
             if (ssh->ctx->channelReqSubsysCb) {
                 rej = ssh->ctx->channelReqSubsysCb(channel, ssh->channelReqCtx);
```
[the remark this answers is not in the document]
ejohnstown: Added the functions `wolfSSH_ChannelGetSessionCommandSz()` and `wolfSSH_GetSessionCommandSz()`. Changed `wolfSSH_GetSessionCommand()` call `wolfSSH_ChannelGetSessionCommand()` to parallel the size versions of the functions.

tests/unit.c
```diff
@@ -16848,6 +16910,85 @@ static int test_SshResourceFree_zeroesSecrets(void)
...
+    }
+
+    for (i = 0; i < commandSz; i++) {
+        if (commandBytes[i] != 0) {
+            result = -716;
+            goto out;
+        }
+    }
+
+out:
```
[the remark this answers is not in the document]
ejohnstown: Reversed the defer -- this is fixed.

`out:` now deletes a non-NULL `channel` ahead of `wolfSSH_free(ssh)`. The
order matters: the channel holds a back-pointer to `ssh`, and the heap
argument reads through `ssh->ctx`.

Took `ssh->ctx->heap` rather than the suggested bare `NULL`. The tests build
their contexts with a NULL heap, so the two are the same pointer, but every
other `ChannelDelete()` call in `unit.c` passes the context heap.

Measured rather than reasoned. Forcing the `-713` path and running the suite
under `leaks --atExit` reports `2 leaks for 147776 total leaked bytes`, root
stack `<malloc in ChannelNew>` -- the channel and its 144K input buffer. With
the delete in place, 0 leaks. The success path was already clean: it NULLs
`channel` right after the measured `ChannelDelete()`.

src/ssh.c
```diff
@@ -801,7 +822,9 @@ int wolfSSH_accept(WOLFSSH* ssh)
                     const char* cmd = wolfSSH_GetSessionCommand(ssh);
                     if (cmd != NULL &&
                         WOLFSSH_SESSION_SUBSYSTEM == wolfSSH_GetSessionType(ssh)
-                        && (WSTRNCMP(cmd, "sftp", 4) == 0)) {
+                        && ssh->channelList->commandSz ==
```
padelsbach: Nit: Should we use the getter `wolfSSH_GetSessionCommandSz` for consistency here?

src/ssh.c
```diff
@@ -801,7 +822,9 @@ int wolfSSH_accept(WOLFSSH* ssh)
...
+                                (word32)WSTRLEN("sftp")
+                        && (WSTRCMP(cmd, "sftp") == 0)) {
```
padelsbach, line 834: Also, claude points out that this might allow a malicious client to ignore the `CHANNEL_FAILURE` reply here and carry on with `SFTP_INIT`, which would be answered. Is that real?

src/internal.c
```diff
@@ -13120,39 +13123,68 @@ static int DoChannelRequest(WOLFSSH* ssh,
...
             }
+            else {
+                rej = ssh->appChannels;
+            }
+            sessionReq = 1;
             ssh->clientState = CLIENT_DONE;
         }
         else if (ChannelRequestIs(type, typeSz, "exec")) {
-            ret = GetStringAlloc(ssh->ctx->heap, &channel->command, NULL,
+            ret = GetStringAlloc(ssh->ctx->heap,
```
padelsbach, line 13160: AI finds that `GetStringAlloc` does not zero the command before freeing, and `ChannelDelete` only clears the last one received. So there may still be residual info in memory. Seems possible to me, but I may be missing something

src/ssh.c
ejohnstown: Sure. Changed.

ejohnstown: Switched to `wolfSSH_GetSessionCommandSz()`. The `cmd != NULL` test ahead of it already rules out a NULL channel list, so this is consistency rather than a fix, but it does read better next to the `wolfSSH_GetSessionCommand()` call a line above.

src/ssh.c
ejohnstown: Real, but the path that reaches it is the legacy one, not app-channels.

With `appChannels` set, `wolfSSH_SFTP_accept()` already asks for `sessionGranted`, so a refused subsystem is turned away there. Without it, a subsystem callback that rejects `sftp` still leaves `sessionType` and `command` set -- both are recorded before the reject decision. CHANNEL_FAILURE goes out, `clientState` still reaches `CLIENT_DONE`, and this divert reads only the type and the command, so it sets `ACCEPT_INIT_SFTP` and calls `wolfSSH_SFTP_accept()`. There `appChannels` is 0, and `ACCEPT_INIT_SFTP` sorts above `ACCEPT_CLIENT_SESSION_ESTABLISHED`, so the pre-accept branch is skipped as well and the server drops straight into `SFTP_BEGIN`. The client never has to ignore anything -- the server serves the session it just refused.

The SCP divert just above had the same hole: a rejected `scp ...` exec still returned `WS_SCP_INIT`.

Both diverts now require `channel->sessionGranted`. Two tests cover it, each running a refused request and a granted control.

src/internal.c
ejohnstown: Confirmed. `ChannelDelete()` does wipe the command now, but `GetStringAlloc()` releases the previous buffer with a plain `WFREE`, so a second exec or subsystem request on the same channel leaves the first command line sitting in the freed block. Only the last one is ever wiped.

Fixing it inside `GetStringAlloc()` doesn't work: `sSz` is in/out and most of its callers pass NULL, so it has no length for the buffer it is about to release. A small `ScrubChannelCommand()` helper wipes `channel->command` at both parse sites instead. The free stays with `GetStringAlloc()`, so a parse that fails behind it leaves no dangling pointer.

The test feeds two exec requests with the retain-on-free allocator installed around the second, then reads the first command's freed bytes back. Without the scrub they still read `sh -c 'login --password hunter2'`.

padelsbach (approved):

PR #1243, opened 2026-09-08
wolfsshd: complete the Windows user profile fallback

yosuke-wolfssl:
### Problem

`_GetHomeDirectory()`'s fallback never set `PROFILEINFO.dwSize`, which `LoadUserProfileW()` rejects with `ERROR_INVALID_PARAMETER`. That fallback runs for any user with no profile on the host — the network logon wolfsshd performs does not create one — so a first login by such a user was refused: public-key auth could not locate `authorized_keys`, and SFTP and shell sessions could not resolve a working directory. Both Windows CI workflows worked around this by hand-writing a `ProfileList` registry entry.

Two further defects sat on the same path. The home directory came from `ExpandEnvironmentStringsW(L"%USERPROFILE%")`, which expands against the **service account's** environment, so fixing `dwSize` alone would have handed `systemprofile` back as every user's home. And the loaded registry hive was never unloaded.

### Fix (`apps/wolfsshd/auth.c`)

- **`dwSize`** is set, and the profile is loaded whenever `WOLFSSHD_AUTH.profile` is `NULL` — on every session, not only when the profile is missing. The hive is machine-wide, so a session holding no handle had it unloaded underneath it when another session for the same user ended.
- **`_GetProfileDirectory()`** reads the path with `GetUserProfileDirectoryW()` on the auth token, replacing both `SHGetKnownFolderPath()` and the `%USERPROFILE%` expansion.
- **`wolfSSHD_AuthCloseToken()`** unloads the profile before closing the token.
- **`RevertToSelf()`** moves ahead of the token close in `SHELL_Subsystem()`'s cleanup — `UnloadUserProfile()` needs the service account's `SE_RESTORE_NAME`/`SE_BACKUP_NAME`, which the impersonated user's token lacks.

| Caller | Helper | Builds a profile |
|---|---|---|
| `CheckPublicKeyWIN()`, before authenticating | `_GetProfileDirectory()` | no |
| `SFTP_Subsystem()`, `SHELL_Subsystem()` | `_GetHomeDirectory()` | yes |

The split matters: `userAuthCb` runs before signature verification, so a signature-less public-key probe reaches `CheckPublicKeyWIN()`. Keeping the load out of that path stops an unauthenticated client from forcing profile creation for any account it names.

**Behaviour worth noting:** public-key auth is still impossible for a user who has never logged on, since there is no home directory to hold `authorized_keys`. That matches master, and is deliberate — profile creation is deferred to an authenticated session.

Closes f-13326, and covers the service-account home directory and the unreleased hive, neither of which was filed separately.

### Why loading on every session is safe

Microsoft's `LoadUserProfile` documentation states the obligation directly: when a service impersonates a user the system does not load that user's profile, so the service should, and *"it is your responsibility to load the user's registry hive into the HKEY_USERS registry key with the LoadUserProfile function before you call CreateProcessAsUser"* — otherwise access to `HKEY_CURRENT_USER` "may not produce results consistent with a normal interactive logon". wolfsshd calls `CreateProcessAsUserW()`, so a load is owed on every session; loading only when the profile was absent was the anomaly.

Loading a profile that is already mounted does not reinitialise it — the second caller simply gets its own handle, and Windows keeps the hive mounted until the last handle is released. Neither the `LoadUserProfile` nor the `UnloadUserProfile` page documents that reference counting, so it was confirmed on Windows rather than assumed: with two overlapping sessions for one user, the unfixed code unmounts the hive the second session is still running against when the first ends, and this change keeps it mounted until the second session ends. The concurrent-session test below is that experiment, and it runs on every push.

For a user whose home directory exists but is not a registered valid profile, Windows builds a fresh profile beside it — the same thing it does on an interactive logon. That is what the workflow rework accounts for.

### Tests (`.github/workflows/`)

A `no_profile` leg in `windows-sftp.yml` creates users with `net user` alone and asserts they start with no `ProfileList` entry or home directory. It then checks that an exec session builds the profile, runs as `testuser` and releases the hive; that with two overlapping sessions for a second profile-less user, ending the first leaves the hive loaded for the second and the second releases it; and that an SFTP session lands in `testuser`'s own home rather than `systemprofile` and releases the hive at connection teardown.

Both Windows workflows now log `testuser` on once so Windows builds a real profile, in place of creating the home directory and `ProfileList` entry by hand. The fabricated profile had no `NTUSER.DAT`, so once wolfsshd actually loaded it Windows built a second profile beside it.

### Verification

- All 10 Windows jobs pass: 3 `windows-sftp` legs and 7 `windows-cert-store-test` matrix cells.
- Negative control: the concurrent-session test fails on the unfixed code with `ending session A unmounted the hive under session B`.
- MSVC Debug x64 builds clean, no warnings on the changed files; `api-test` and `unit-test` pass.
- GCC preflight sweep: lint clean, 6/6 configs.

### Not in this PR

- `CreateProcessAsUserW()` passes a `NULL` environment, so a spawned shell inherits the service's environment rather than the user's.
- A failed `UnloadUserProfile()` leaves the hive mounted, and the branch is reachable without fault injection: `SHELL_Subsystem()` never waits for or terminates the process it starts, so a client disconnect mid-command reaches `cleanup:` with the child still live. Master orphans the same child; unloading the hive under it is new. The bounded wait plus `TerminateProcess()` belongs with the orphaned-process bug, and that case is untested on this path.


apps/wolfsshd/auth.c
```diff
@@ -2042,27 +2044,34 @@ static int _GetHomeDirectory(WOLFSSHD_AUTH* auth, const char* usr, WCHAR* out, i
...
-                "[SSHD] Error %lu loading user %s",
-                (unsigned long)GetLastError(), usr);
-            ret = WS_FATAL_ERROR;
+        /* failed with get known folder path, try with loading the user profile.
+         * A second load would replace the handle needed to unload the first. */
+        if (auth->profile == NULL) {
+            pInfo.dwSize = sizeof(pInfo);
+            pInfo.dwFlags = PI_NOUI;
+            pInfo.lpUserName = usrW;
+            if (LoadUserProfileW(wolfSSHD_GetAuthToken(auth), &pInfo) != TRUE) {
```
yosuke-wolfssl: Fixed

apps/wolfsshd/auth.c
```diff
@@ -2083,11 +2092,20 @@ HANDLE wolfSSHD_GetAuthToken(const WOLFSSHD_AUTH* auth)
...
 }
 
-/* Close and clear the impersonation token. Safe to call more than once. */
+/* Unload the profile and close the token. Safe to call more than once. */
 void wolfSSHD_AuthCloseToken(WOLFSSHD_AUTH* auth)
 {
     if (auth != NULL && auth->token != NULL &&
             auth->token != INVALID_HANDLE_VALUE) {
+        if (auth->profile != NULL) {
+            if (UnloadUserProfile(auth->token, auth->profile) != TRUE) {
```
yosuke-wolfssl, line 2102: Fixed

apps/wolfsshd/auth.c
```diff
@@ -2083,11 +2093,21 @@ HANDLE wolfSSHD_GetAuthToken(const WOLFSSHD_AUTH* auth)
...
+                wolfSSH_Log(WS_LOG_ERROR,
+                    "[SSHD] Error %lu unloading user profile",
+                    (unsigned long)GetLastError());
+                RegCloseKey((HKEY)auth->profile);
```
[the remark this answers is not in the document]
yosuke-wolfssl: Accurate on the mechanism, and we are leaving the code as it stands. Reasoning for the record:

`RegCloseKey()` does not release the profile reference — only `UnloadUserProfile()` does — so after a failed unload the hive remains mounted whatever we do with the handle. Closing it at least avoids blocking any other unloader, which is why it is there; it was added at the request of the earlier automated review on this PR.

Retrying is not reachable where suggested. `wolfSSHD_AuthCloseToken()` is the last point at which the token is valid: it closes the token on the following lines, and its callers either free the auth context or replace the token with a new logon's. `UnloadUserProfile()` needs a valid token and handle together, so there is no later moment to retry from.

The known trigger is removed by this PR. The unload now runs as the service account, since `RevertToSelf()` moved ahead of the token close, and that ordering was what made it fail with `ERROR_ACCESS_DENIED`. Testing on Windows did not reproduce a failed unload otherwise, including an abrupt client disconnect during a long-running command.

If one does occur, the cause is a session child process still holding the hive, and the fix belongs in session teardown — waiting on the child in `SHELL_Subsystem()` — rather than in the unload call. Tracked separately.

ejohnstown (approved):

PR #1245, opened 2026-09-08
scp: return directory entries from the Zephyr entry walk

yosuke-wolfssl:
### Problem
The `WOLFSSH_ZEPHYR` branch of `FindNextDirEntry()` ended in `while (1)`, discarding every entry `fs_readdir()` returned. It could only reach `return WS_FATAL_ERROR` or `return WS_NEXT_ERROR` — the shared `return WS_SUCCESS` was unreachable on Zephyr.

That function is called only from the `WOLFSSH_SCP_RECURSIVE_REQUEST` arm of `wsScpSendCallback()`, and the caller reads `WS_NEXT_ERROR` as end-of-directory. A peer running `scp -r user@host:dir .` against a Zephyr server got the directory header and an immediate exit, so **every directory transferred empty**. Single-file SCP was unaffected. Functional only — no overflow or leak, and the loop always terminated, since `fs_readdir()` signals end-of-directory with `name[0] = 0`. 

### Fix (`src/wolfscp.c`)
Skip `.` and `..` and fall through to `return WS_SUCCESS`, matching the POSIX and Windows branches:

```c
    } while (((WSTRLEN(ctx->entry.name) == 1) &&
              (WSTRNCMP(ctx->entry.name, ".", 1) == 0)) ||
             ((WSTRLEN(ctx->entry.name) == 2) &&
              (WSTRNCMP(ctx->entry.name, "..", 2) == 0)));
```

The skip is not cosmetic everywhere:

| Zephyr FS | Reports `.` / `..`? | Effect |
|---|---|---|
| FAT (ELM FatFs) | No — `dir_read()` filters leading `.` | Skip is a no-op |
| littlefs | Yes — `lfs_dir_read_()` synthesizes both | Without it, `.` is a directory and `ScpPushDir()` re-descends the same directory unboundedly |

Closes f-13315.

### Tests (`tests/api.c`)
`test_wolfSSH_SCP_SendRecursiveEntry()` stages a one-file directory and drives three `WOLFSSH_SCP_RECURSIVE_REQUEST` calls through `wsScpSendCallback()`, asserting enter-dir, then the entry name/size/bytes, then final exit. Driving the callback directly is the only route in — the example client cannot issue `scp -r -f`. It needs no threads, sockets, `mkdtemp` or symlinks, so unlike every other SCP test it is **not** gated `!defined(WOLFSSH_ZEPHYR)` and runs in the Zephyr CI.

### Verification
- Host `make check`: 8 passed, 3 skipped, 0 failed.
- `gcc-13 -Werror` sweep: 6/6 configs clean.
- Zephyr qemu_x86 (4.4.99, SDK 1.0.1), FAT RAM disk: `Zephyr wolfSSH tests passed`.
- Negative controls: with `while (1)` restored, the Zephyr run fails at the new assertion (`-1047` instead of 64 bytes); with the POSIX dot-skip disabled, it fails with `-1045`, having descended into `.`.

### Not in this PR
Nested-subdirectory descent (the non-final `WS_SCP_EXIT_DIR`) and the empty-directory case stay uncovered — both pre-existing gaps, flagged in review and deferred.


yosuke-wolfssl:
CI issues would be gone once PR #1242 is merged

ejohnstown (approved):

PR #1251, opened 2026-09-15
examples, ide: classify the worker's status where ssh->error is its own

yosuke-wolfssl:
### Problem

`wolfSSH_worker()` reports a blocked read or write two ways: as its return, or
as `WS_FATAL_ERROR` with the code left in `ssh->error`. `ssh_worker()`
dispatched on the return alone, so the second shape arrived as a bare failure
with its meaning stranded in `ssh->error`. That one gap has two consequences:

- **Live sessions dropped.** The code compensated by re-reading
  `wolfSSH_get_error()` down in the terminal arm, ~200 lines later. The
  `!eofAnswered && echoOnly` drain runs in between and latches its own code, so
  a `WS_WINDOW_FULL` from the drain read as fatal and tore down a healthy
  session.
- **Both ends deadlock.** #1234's `wantWrite` retry is keyed on
  `rc == WS_WANT_WRITE`, which the second shape never produces. `wantWrite`
  stayed 0, so `select()` waited on readability alone while the peer waited on
  output the server was still holding.

A third case carries no want-status at all: an owed write riding a
`WS_CHAN_RXD`, `WS_EOF` or `WS_CHANNEL_CLOSED` return. In `wolfSSH_worker()`'s
flush block, `ret == WS_CHAN_RXD` with `sendRet == WS_WANT_WRITE` fires none of
the three branches, so the event return is handed back with output still
queued. `wolfssh/ssh.h` says to ask `wolfSSH_OutputPending()` rather than read
a status, which is what the POSIX `wolfsshd` loop already does.

### Fix (`examples/echoserver/echoserver.c`)

Carry the worker's meaning forward in `rc`, at the call site, while
`ssh->error` still belongs to the worker:

```c
rc = cnt_r;
if (rc == WS_FATAL_ERROR) {
    int err = wolfSSH_get_error(ssh);

    if (err == WS_WANT_READ || err == WS_WANT_WRITE)
        rc = err;
}
```

- **#1234's `wantWrite = 1; continue;` arm is kept.** The remap feeds it the
  second shape, arming the write set where nothing did before, so this extends
  #1234 rather than competing with it.
- `wantWrite` is additionally set from `wolfSSH_OutputPending()` before
  `writeFds` is built, which is the only form that covers the event-return
  case.
- The terminal arm stops re-reading `wolfSSH_get_error()` and tests `rc`. **The
  two echoservers deliberately differ here:** `rc != WS_WANT_READ` in this
  file, whose `WS_WANT_WRITE` arm consumes that case, and
  `rc != WS_WANT_READ && rc != WS_WANT_WRITE` in the ESP-IDF copy, which has no
  write set and whose empty `WS_WANT_WRITE` arm is removed. Please do not
  "align" them.
- The chain is guarded by `if (cnt_r < 0)`, so a `WS_SUCCESS` pass never
  reaches the terminal arm.

Includes `tests: run app-driven sftp in a block build` (John Safranek), which
un-skips the case that exercises this.

`scripts/sshclient.test` and `scripts/fwd-bulk.test` name the client's missing
non-blocking mode in their skip messages, matching `scripts/scp.test`. Their
old comments said the other echoserver scripts skip a block build for the same
reason, which stopped being true when `sftp.test` un-skipped above. Message and
comment only; no logic changed.

### Verification

- Preflight: lint plus six gcc-13 `-Werror` configs, all clean.
- 166 `unit` + `regress` + `testsuite` pass; `sshclient`, `scp`, `get-put`,
  `sftp` and `fwd` pass run serially.
- Contention (`--enable-sftp -DWOLFSSH_TEST_BLOCK -DWOLFSSH_NO_FPKI`), with the
  app-driven case now running rather than skipped: **21s / 39s / 88s** at
  `WOLFSSH_BLOCK_PROB` 30 / 50 / 70.
- Mutations at prob 70:

| Change | Result |
|---|---|
| Remove the `WS_FATAL_ERROR` remap | fails, `-1009`, 77s |
| Remove #1234's `wantWrite` arm | hangs, killed at 200s |
| Remove the `wolfSSH_OutputPending()` lines | still passes, 88s |

The third row is deliberate disclosure: **no test distinguishes the
`wolfSSH_OutputPending()` arming.** `wolfSSH_worker()` attempts a flush on
every call, so an owed write is retried for as long as the loop keeps calling
the worker; reaching the hole needs the peer to fall silent immediately after
sending channel data while the socket refuses writes. That line is contract
conformance, not a measured fix.

The echo/shell path also stays uncovered: `scripts/sshclient.test` cannot run
under `WOLFSSH_TEST_BLOCK` because `apps/wolfssh` has no non-blocking mode, and
neither does `portfwd` for `scripts/fwd-bulk.test`. That is what their skip
messages now say.


yosuke-wolfssl:
Thanks — noting what this PR does against the Fenrir findings raised on #1217, since the automated scan here only covers newly changed lines.

**Closed by this PR:** `e8072b80658f60e8` — "Nonblocking echoserver stalls on queued output". The suggested fix there was to monitor `sshFd` in `writefds` on `WS_WANT_WRITE` and re-run the worker when it becomes writable. #1234 added that write set; this PR supplies the half it was missing, since `wolfSSH_worker()` also reports a blocked send as `WS_FATAL_ERROR` with the code in `ssh->error`, which never produced `rc == WS_WANT_WRITE` and so never armed the set. `wantWrite` is now also driven from `wolfSSH_OutputPending()`, which covers an owed write riding a `WS_CHAN_RXD`, `WS_EOF` or `WS_CHANNEL_CLOSED` return.

Measured with the app-driven SFTP case un-skipped: passes at `WOLFSSH_BLOCK_PROB` 30/50/70; removing the remap fails `-1009`; removing the write-set arm hangs.

**Deliberately not closed here:** `49a64d03eeb64a47` — the ESP-IDF fork of the same stall. That file gets this PR's status classification but no write set, because porting one is a real port rather than a copy (different state enum names, no agent-open hook, no subsystem dispatch) and nothing in the tree builds or runs it — no `include.am`, no CI job. It would ship uncompiled. Tracked separately; its thread should not be taken as resolved by this PR.

**Still live elsewhere:** `29cd996eb8e57784` — "Client rejects worker's documented transient return", in `examples/client/client.c`. Same defect class as this PR fixes, different file: the whitelist after `wolfSSH_worker()` in the shutdown path accepts `WS_SUCCESS`, `WS_SOCKET_ERROR_E`, `WS_CHANNEL_CLOSED` and `WS_EOF`, so a `WS_FATAL_ERROR` carrying `WS_WANT_READ`/`WS_WANT_WRITE` reaches `err_sys()` and aborts. Unchanged on master. Out of scope here, flagging so it isn't assumed fixed by association.

ejohnstown (approved):

PR #1258, opened 2026-09-16
Minor internal and wolfsshd fixes

stenslae:
Reworked the wide-char to multibyte conversion so the exact buffer size needed is used. Removed dead ECC agent code, as only RSA supports agent signing. Added an intialized to 0 to avoid use-of-unitialized value warnings.


ejohnstown (approved):

apps/wolfsshd/wolfsshd.c
```diff
@@ -4091,19 +4091,29 @@ static void wolfSSHD_ServiceCb(DWORD CtrlCode)
...
-    
-    retSz = (int)wcslen(in) * 2;
-    ret   = (char*)WMALLOC(retSz + 1, heap, DYNTYPE_SSHD);
-    if (ret != NULL) {
-        size_t numConv = 0;
-        if (wcstombs_s(&numConv, ret, retSz, in, retSz) != 0) {
-            WFREE(ret, heap, DYNTYPE_SSHD);
-            ret = NULL;
+/* Set *err to WS_MEMORY_E on alloc failure, or WS_FATAL_ERROR on conversion failure. */
+static char* _convertHelper(WCHAR* in, void* heap, int* err) {
```
ejohnstown: Convention: The opening brace for a function should be on its own line.

apps/wolfsshd/wolfsshd.c
```diff
@@ -4091,19 +4091,29 @@ static void wolfSSHD_ServiceCb(DWORD CtrlCode)
...
+    char* ret = NULL;
+    size_t needed = 0;
+
+    /* Query exact size needed for multi-byte ACPs. Avoids zero-size for empty args. */
```
ejohnstown: Convention: Lines are supposed to wrap before column 80. Just kick the second sentence to the next line.

src/internal.c
```diff
@@ -20936,44 +20936,20 @@ static int PrepareUserAuthRequestEccCert(WOLFSSH* ssh, word32* payloadSz,
...
-                else {
-                    ret = wc_EccPrivateKeyDecode(
-                            authData->sf.publicKey.privateKey,
-                            &idx, &keySig->ks.ecc.key,
-                            authData->sf.publicKey.privateKeySz);
-                }
+        /* No WOLFSSH_AGENT branch: only RSA certs support agent signing. */
+        if (authData->sf.publicKey.privateKey == NULL ||
+                authData->sf.publicKey.privateKeySz == 0) {
+            /* Avoid misleading ASN error for cert-store-only clients without in-memory key. */
```
ejohnstown: Convention: Lines are supposed to wrap before column 80. Just split it somewhere.

src/internal.c
```diff
@@ -20936,44 +20936,20 @@ static int PrepareUserAuthRequestEccCert(WOLFSSH* ssh, word32* payloadSz,
...
-                         "cert-store slot");
-                    ret = WS_BAD_ARGUMENT;
-                }
-                else {
-                    ret = wc_EccPrivateKeyDecode(
-                            authData->sf.publicKey.privateKey,
-                            &idx, &keySig->ks.ecc.key,
-                            authData->sf.publicKey.privateKeySz);
-                }
+        /* No WOLFSSH_AGENT branch: only RSA certs support agent signing. */
```
ejohnstown: Reword. Among the certificate key types, only RSA implements agent signing. Plain ECDSA keys do use the agent, in the non-certificate ECC path.

src/internal.c
```diff
@@ -21138,6 +21090,7 @@ static int BuildUserAuthRequestEccCert(WOLFSSH* ssh,
         }
         else
 #endif /* WOLFSSH_WINDOWS_CERT_STORE */
+        /* No WOLFSSH_AGENT branch: only RSA certs support agent signing. */
```
ejohnstown: Reword. Among the certificate key types, only RSA implements agent signing. Plain ECDSA keys do use the agent, in the non-certificate ECC path.

apps/wolfsshd/wolfsshd.c
```diff
@@ -4091,19 +4091,29 @@ static void wolfSSHD_ServiceCb(DWORD CtrlCode)
...
+    if (wcstombs_s(&needed, NULL, 0, in, 0) == 0 && needed > 0) {
+        ret = (char*)WMALLOC(needed, heap, DYNTYPE_SSHD);
+        if (ret == NULL) {
+            *err = WS_MEMORY_E;
+        }
+        else {
+            size_t numConv = 0;
+            if (wcstombs_s(&numConv, ret, needed, in, needed - 1) != 0) {
```
ejohnstown: `needed` is already big enough. You don't need to subtract 1 here. The sizing query returns the size including the terminating null and that's what you allocate, so pass `needed` for the count too.

ejohnstown (changes requested):

apps/wolfsshd/wolfsshd.c
stenslae: Changed to follow code convention

apps/wolfsshd/wolfsshd.c
stenslae: Resized line

apps/wolfsshd/wolfsshd.c
stenslae: Dropped -1

src/internal.c
stenslae: Reworded

src/internal.c
stenslae: Reworded

ejohnstown (approved):

PR #1269, opened 2026-09-21
Fix wolfsshd's cleared umask and the SFTP attribute encoder

ejohnstown:
Four unrelated Fenrir items, one commit each. The wolfsshd umask is the one with an operational effect: every session inherited a cleared mask, so files created from a shell or through SCP came out world-writable.

- sftp: SFTP_SetAttributes() emits a zero extension count, because it writes no extension records. A caller's atr->extCount promised records that never followed, and the peer's parser read past the block looking for them (F-2479).
- wolfsshd: Subsystem, ChallengeResponseAuthentication, UsePAM, X11Forwarding, PrintMotd, AcceptEnv and UseDNS log a warning instead of parsing and then being dropped. An unknown keyword is fatal, so silence here read as support for a setting that is not enforced (F-2868).
- wolfsshd: the daemon holds a 022 umask after daemonizing, so sessions no longer create 0666 files and 0777 directories. WOLFSSHD_DEFAULT_UMASK sets it at build time (F-3671).
- tests: the canned KEX, host key, cipher and MAC lists are checked for contents, not just for being non-null, so an inverted guard cannot put SHA-1 or AES-CBC back into the default proposal unnoticed (F-13961).

Both new tests were confirmed to fail with their fix reverted. make check passes on macOS and Linux at each commit, and the wolfsshd suite is green.


tests/regress.c
```diff
@@ -11879,6 +11879,80 @@ static void TestOct2DecRejectsInvalidNonLeadingDigit(void)
...
+        in.atime = 0x5A5A5A5A;
+        in.mtime = 0x6B6B6B6B;
+        /* a caller that set the flag may well have set a count too */
+        in.extCount = 3;
+
+        encSz = wolfSSH_TestSftpSetAttributes(buf, (word32)sizeof(buf), &in);
+        AssertTrue(encSz > 0);
+
+        /* the encoder writes exactly the size it advertises */
+        AssertIntEQ(buf[encSz], 0xEE);
```
philljj, line 11931: Does `encSz` need bounds checking, like `encSz <= sizeof(buf) - 1`?

It wasn't obvious to me tracking the flow of calls that `buf[encSz]` would be guaranteed safe.

philljj (changes requested):
Looks good,  just one nit

tests/regress.c
ejohnstown: Good catch. The test wrapper returns an error if the encoded size would be larger than the buffer, so `encSz <= sizeof(buf)` was guaranteed. But `buf[encSz]` needs `encSz < sizeof(buf)`, and nothing in the test checked that. It now asserts it before reading the canary.

While I was there I also tightened the test: every flag combination now comes with its expected encoded size, and the test checks the encoder returns exactly that. Before, it only checked that the size was positive and then compared the encoder and decoder against each other, so a sizing mistake shared by both would have gone unnoticed. The largest case is 36 bytes, well within the 64-byte buffer.

philljj (approved):

PR #1270, opened 2026-09-21
Add wolfSSH_ChannelIdPeek()

ejohnstown:
Adds a peek counterpart to wolfSSH_ChannelIdRead(): it reports what is buffered on the channel named by channelId without consuming it or crediting the window.

- wolfSSH_ChannelIdPeek() carries wolfSSH_stream_peek()'s contract, but names its channel and peeks during a rekey
- a NULL buf asks only for the count, still capped at bufSz; a drained channel reports WS_EOF, or WS_DISCONNECT once the session is gone
- test_ChannelIdPeek() covers the count, the unconsumed bytes, the rekey case, and both drained reports
- the SFTP test gives the echoserver thread and the client their own argv, since the server thread holds ser.argv for as long as it runs


philljj (approved):
looks good to me

